PR #5, opened 2019-09-25
FS-11587: Fixed hardcoded uuid length in msrp_find_uuid function

sergey-safarov:


mjerris:
can you provide reference in spec to what uuid should be please

sergey-safarov:
Are you want reference to uuid length?
My use case is kazoo installations.
Kazoo defines own ID in dialplan and later MSRP messages cannot be route because length of custom uuid and default is different.

mjerris:
looking to confirm in msrp spec if you can have different length uuid or not

sergey-safarov:
please check [RFC4975](https://tools.ietf.org/html/rfc4975#section-4)
Here
```
To-Path: msrp://biloxi.example.com:12763/kjhd37s2s20w2a;tcp
```
UUID here "kjhd37s2s20w2a" shorted then 36 char sting.

mjerris:
ok so that looks good.  Please add a unit test for this.

sergey-safarov:
I cannot create test unit.

mjerris:
please re-open when you are able to add a unit test for this.

mjerris:
for reference, some of our current unit tests:

https://github.com/signalwire/freeswitch/tree/master/tests/unit

https://github.com/signalwire/freeswitch/tree/master/src/mod/applications/mod_http_cache/test

PR #6, opened 2019-09-25
[Build-System] Relocate modules deps in freeswitch.spec

sergey-safarov:
FS core not use this deps


sergey-safarov:
@s3rj1k 
this because `net-snmp-devel` need only for `mod_snmp` and this deps already present in `event-snmp` package.
https://github.com/signalwire/freeswitch/blob/master/freeswitch.spec#L1114

s3rj1k:
@sergey-safarov I see, missed that, looks good, thanks.

andywolk:
Test build is OK. This is good to go.

PR #11, opened 2019-09-25
[New Feature]: Fallback Array Support

luanyata:
**File**: jquery.jsonrcpclient.js

**Feature**: Possibility of passing a string array in fallbackUrl property on Verto installation

**Motivation**: Need to use multiple wss servers as fallback

### DIFF
--- a/html5/verto/js/src/jquery.jsonrpcclient.js
+++ b/html5/verto/js/src/jquery.jsonrpcclient.js
@@ -77,6 +77,7 @@
         }, options);
 
         self.ws_cnt = 0;
+        self.ws_fallback_cnt = 0;
 
         // Declare an instance version of the onmessage callback to wrap 'this'.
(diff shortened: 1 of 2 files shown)


html5/verto/js/src/jquery.jsonrpcclient.js
```diff
@@ -322,8 +323,9 @@
                         self.options.onWSClose(self);
                     }
 
-                    if (self.ws_cnt > 10 && self.options.wsFallbackURL) {
-                      self.options.socketUrl = self.options.wsFallbackURL;
+                    if (self.ws_cnt > ((self.ws_fallback_cnt + 1) * 10) && self.options.wsFallbackURL && self.options.wsFallbackURL.length > 0) {
```
edolix: `self.options.wsFallbackURL` could be a string. I'd suggest to force it to be an array or - at least - check the type.

luanyata:
dealt with in PR #35

PR #13, opened 2019-09-25
$.FSRTC.prototype.stopPeer() missing self variable declaration

thehunmonkgroup:
The stopPeer() method in jquery.FSRTC.js is missing a 'var self = this;'
declaration at the top of the function, which leads the wrong object being
assigned to self, and the self.peer.stop() function never being called.

This breaks proper hangup of calls using this function, such as a screenshare
call.

### DIFF
--- a/html5/verto/js/src/jquery.FSRTC.js
+++ b/html5/verto/js/src/jquery.FSRTC.js
@@ -282,6 +282,8 @@
     };
 
     $.FSRTC.prototype.stopPeer = function() {
+        var self = this;
+
         if (self.peer) {
             console.log("stopping peer");
             self.peer.stop();


mjerris:
we are working on a plan to move this out of freeswitch tree into its own repository soon so holding on this until we sort out those details.

thehunmonkgroup:
@mjerris this is a pretty straightfoward bugfix, is there some reason why it can't be committed now?

mjerris:
> @mjerris this is a pretty straightfoward bugfix, is there some reason why it can't be committed now?

just because we are going to completely replace this file with the new sdk when we migrate to new repo.

thehunmonkgroup:
@mjerris I guess I don't understand the rationale -- this is a one-line bug fix that will take 5 seconds to commit, and then it will be a fixed problem now, no matter how long it takes to get the new repo up.

And with that comment, I surrender :)

crienzo:
code moved to https://github.com/freeswitch/verto-client

PR #14, opened 2019-09-25
Allow overriding video resolution test list in FSRTC lib

thehunmonkgroup:
Moves the list to the supportedResolutions property of the $.FSRTC class.

This is a file-level property, so to override it the developer would need
to set $.FSRTC.supportedResolutions in their init code, prior to the
point where an $.FSRTC instance would call the resolution checking code.

This change is useful for decreasing initial setup time in cases where the application doesn't use all the resolutions currently tested.

### DIFF
--- a/html5/verto/js/src/jquery.FSRTC.js
+++ b/html5/verto/js/src/jquery.FSRTC.js
@@ -72,6 +72,10 @@
         return newLine.join(' ');
     }
 
+    function defaultSupportedResolutions() {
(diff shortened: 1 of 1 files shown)


mjerris:
we are working on a plan to move this out of freeswitch tree into its own repository soon so holding on this until we sort out those details.

crienzo:
Code moved to https://github.com/freeswitch/verto-client

PR #35, opened 2019-10-03
Multi fallback

luanyata:


mjerris:
we are in process of moving this to a new repository so this is on hold for now.  That being said, need to remove all the whitespace only changes from this commit before it can be removed.

crienzo:
moved to github.com/freeswitch/verto-client  please re-open the PR there

Issue #22, opened 2019-09-28
global-opened ports 2855+2856 can't be disabled


sh-dvl:
just started with fs 1.10 from git after using a previous version for longer time, but after it has started there're always two ports opened globally: 0.0.0.0:2855 and 0.0.0.0:2856.

after long-time search in documentation without any result, several tries with disabling modules and then deeper look at the code it's clear: there's no possibility to change this behaviour without patching the code. wtf?

because unnecessarily opened ports (and I've really nothing found, that would explain, why this was always necessary in such a way) are always potential risks, this is a bug for me and should be fixed soon.

seven1240:
it's switch_msrp and enabled by default. I'm open to ideas to disable it but the easiest way perhaps to add a `disable-msrp = true` param in `msrp.conf.xml`. Could you make a patch?

crienzo:
Fixed

PR #36, opened 2019-10-03
Disable MSRP option

cdosoftei:
MSRP is enabled by default; this PR implements a configuration option so the default behaviour can be changed when desired.

Also addresses signalwire/freeswitch#22

### DIFF
--- a/conf/vanilla/autoload_configs/msrp.conf.xml
+++ b/conf/vanilla/autoload_configs/msrp.conf.xml
@@ -1,8 +1,8 @@
 <configuration name="msrp.conf" description="MSRP">
   <settings>
     <param name="listen-ip" value="$${local_ip_v4}"/>
-    <param name="listen-port" value="2855"/>
-    <param name="listen-ssl-port" value="2856"/>
+    <!-- <param name="listen-port" value="2855"/> -->
+    <!-- <param name="listen-ssl-port" value="2856"/> -->
     <!-- <param name="message-buffer-size" value="50"/> -->
     <!-- <param name="debug" value="true"/> -->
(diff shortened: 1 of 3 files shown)


mjerris:
I think i'd rather have this where if you don't configure the msock ports we don't listen on them, instead of a different setting.

cdosoftei:
:+1: That would make it consistent with other configuration options (ws and wss ports for Sofia profiles come to mind).

Can definitely make the change -- @seven1240 do you wish to weigh in (given the feedback in #22)?

seven1240:
@cdosoftei I totally agree with @mjerris, the only concern is that will change behaviour for existing users using msrp. we should alert behaviour change on the next release.

cdosoftei:
Please check the [revised PR](https://github.com/signalwire/freeswitch/pull/36/files), it should be in line with the most recent feedback.

To @seven1240's point, the configuration behaviour change should be documented in the release notes (and perhaps in Confluence as well?); the vanilla configuration is not disrupted by this PR as it would yield the same behaviour, i.e. both the plain text and the secured listeners would be enabled.

conf/vanilla/autoload_configs/msrp.conf.xml
```diff
@@ -3,7 +3,6 @@
     <param name="listen-ip" value="$${local_ip_v4}"/>
     <param name="listen-port" value="2855"/>
     <param name="listen-ssl-port" value="2856"/>
-    <!-- <param name="disable-msrp" value="true"/> -->
```
andywolk: That was added by previous commit and now removed? How so?

PR #39, opened 2019-10-07
(mod_callcenter) add queue fastreload

areski:
Hi,

Reference to https://freeswitch.org/jira/browse/FS-9920
patch from @andreabat 

`fastreload` is useful to not reload all the agents when doing changes on queue settings.


areski:
Thanks for commenting @mochouinard 
To explain briefly, at the moment if we have a queue with some agents in Status "Available", I don't want them to be logged out when we do a simple update of the queue.

In other words, let say that you want to only change the `strategy` from a queue where AgentA & AgentB are connected and Available, I don't want AgentA & AgentB to be reset and being `Logged Out`.

If you have an other approach for this, please let me know

mochouinard:
If you remove status="Logged Out" from your agent xml config, it will not reset these value.  It would default to status="Logged Out" if that user didn't exist and it was created.

areski:
This works fine, thanks!

PR #40, opened 2019-10-07
[mod_callcenter] Unreserve callcenter events when config errors loading

areski:
Ref: https://freeswitch.org/jira/browse/FS-11094

When you load mod_callcenter and there is an issue with the configuration, it doesn't allow you to reload it a second time, this is caused before we reserve the `callcenter::info` event before loading the module successfully.

Here some of the error logs when this is occurring:

> 2018-04-05 06:44:46.614437 [ERR] mod_callcenter.c:4140 Couldn't register subclass callcenter::info!
> 2018-04-05 06:44:46.614437 [CRIT] switch_loadable_module.c:1522 Error Loading module /usr/lib/freeswitch/mod/mod_callcenter.so
*Module load routine returned an error*

To easily reproduce you can put an error in your odbc settings, the load mod_callcenter, after it fails you will not be able to load it again.

### DIFF
--- a/src/mod/applications/mod_callcenter/mod_callcenter.c
+++ b/src/mod/applications/mod_callcenter/mod_callcenter.c
@@ -4208,12 +4208,6 @@ SWITCH_MODULE_LOAD_FUNCTION(mod_callcenter_load)
 	switch_json_api_interface_t *json_api_interface;
 	switch_status_t status;
 
-
-	if (switch_event_reserve_subclass(CALLCENTER_EVENT) != SWITCH_STATUS_SUCCESS) {
-		switch_log_printf(SWITCH_CHANNEL_LOG, SWITCH_LOG_ERROR, "Couldn't register subclass %s!\n", CALLCENTER_EVENT);
-		return SWITCH_STATUS_TERM;
-	}
-
 	memset(&globals, 0, sizeof(globals));
 	globals.pool = pool;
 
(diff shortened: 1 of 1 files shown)


mjerris:
This still isn't quite right.  I'm not sure you want to move it at all, but do need to handle unreserving them in the error return cases (one still exists)

areski:
I have changed as recommended, let me know if that's better.
Now I have moved `switch_event_bind_removable` before `load_config` and using `switch_event_free_subclass` to unreserve `CALLCENTER_EVENT`

src/mod/applications/mod_callcenter/mod_callcenter.c
```diff
@@ -4208,28 +4208,32 @@ SWITCH_MODULE_LOAD_FUNCTION(mod_callcenter_load)
...
 	if (switch_event_reserve_subclass(CALLCENTER_EVENT) != SWITCH_STATUS_SUCCESS) {
 		switch_log_printf(SWITCH_CHANNEL_LOG, SWITCH_LOG_ERROR, "Couldn't register subclass %s!\n", CALLCENTER_EVENT);
 		return SWITCH_STATUS_TERM;
 	}
 
+	/* Subscribe to presence request events */
+	if (switch_event_bind_removable(modname, SWITCH_EVENT_PRESENCE_PROBE, SWITCH_EVENT_SUBCLASS_ANY,
+									cc_presence_event_handler, NULL, &globals.node) != SWITCH_STATUS_SUCCESS) {
+		switch_log_printf(SWITCH_CHANNEL_LOG, SWITCH_LOG_ERROR, "Failed to subscribe for presence events!\n");
+		return SWITCH_STATUS_GENERR;
```
cdosoftei: Should we add
```c
switch_event_free_subclass(CALLCENTER_EVENT);
```
to release the reservation if the handler binding is not successful?

PR #43, opened 2019-10-09
allow process X-AUTH-IP if proxy IP listed in trusted acl.

sergey-safarov:
@lazedo could you look also this PR.

### DIFF
--- a/src/mod/endpoints/mod_sofia/sofia.c
+++ b/src/mod/endpoints/mod_sofia/sofia.c
@@ -10333,16 +10333,74 @@ void sofia_handle_sip_i_invite(switch_core_session_t *session, nua_t *nua, sofia
 
 	if (profile->acl_count) {
 		uint32_t x = 0;
-		int ok = 1;
+		int ok = 0;
 		char *last_acl = NULL;
 		const char *token = NULL;
-		int acl_port = sofia_test_pflag(profile, PFLAG_USE_PORT_FOR_ACL_CHECK) ? network_port : 0;
+		int acl_port;
 
+		int network_ip_is_proxy = 0;
+		const char* x_auth_ip = network_ip;
+		int x_auth_port = network_port;
+		/* Check if network_ip is a proxy allowed to send us calls */
+		if (profile->proxy_acl_count) {
+			switch_log_printf(SWITCH_CHANNEL_LOG, SWITCH_LOG_DEBUG, "%d acls to check for proxy\n", profile->proxy_acl_count);
+			for (x = 0; x < profile->proxy_acl_count; x++) {
+				last_acl = profile->proxy_acl[x];
+				switch_log_printf(SWITCH_CHANNEL_LOG, SWITCH_LOG_DEBUG, "checking %s against acl %s\n", network_ip, last_acl);
+				if (switch_check_network_list_ip_port_token(network_ip, network_port, last_acl, &token)) {
+					switch_log_printf(SWITCH_CHANNEL_LOG, SWITCH_LOG_INFO, "%s is a proxy according to the %s acl\n", network_ip, last_acl);
+					network_ip_is_proxy = 1;
+					break;
+				}
+			}
+		}
+
+		/*
+		 * if network_ip is a proxy allowed to send calls, check for auth
+		 * ip header and see if it matches against the inbound acl
+		 */
+		if (network_ip_is_proxy) {
+
+			/*
+			 * if network_ip is a proxy allowed to send calls,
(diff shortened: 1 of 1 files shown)


sergey-safarov:
In cloud installation (like `kubernetes`) `Kamailio` and `FreeSwitch` IPs dynamically assigned from user defined network. And user cannot assign permanent IP for kamailio host..
When I define this setting in `sip_pofile`
```xml
    <param name="apply-proxy-acl" value="localnet.auto"/>
    <param name="apply-inbound-acl" value="localnet.auto"/>
```
Then I not able auth call by X-AUTH-IP header because Kamailio IP listed in `apply-inbound-acl`.

PR useful for cloud installation.

lazedo:
@sergey-safarov i feel this is going the wrong direction, current implementation works as expected. you just need to provide the correct `acls`. it seems you're trying to provide a `localnet.auto` for both `proxy` and `inbound` which looks wrong. in our use case, we reload the `acls` when there's a change in the network.

sergey-safarov:
1) I have tested kazoo feature FS ACL configuration when added new Kamailio host. I tested on 4.3 branch and found that now it not stable. As example ACL not updated when kamailio + ecallmgr + freeswitch container killed and created new at same time.
2) i think freeswitch ACL update tools is done on kazoo side as workarround of lack ability to use same network as proxy ACL and as inbound ACL.
3) usage case - kazoo installed on three host and two host have installed kamailio + freeswitch on same IP, you will not able transfer call between FS host.That not possible.
4) what must do other projects if need implement Kamailio and FreeSwitch in same network and kamailio IP cannot be fixed? Need create daemon outside of FreeSwitch to dynamically update ACL like do it kazoo? Think this is not good way to solve this.

When FS will be able use same network for proxy and inbound ACL, that will dramatically simplify FS ACL management.
P.S. Do not forget that now you must populate not only kamailio ACL. Need populate FS hosts ACL also to make possible call transfer between FS hosts.

lazedo:
1. probably wrongly tested
2. wrong assumption
3. that's why the port acl was added
4. again, port acl is there for that

sergey-safarov:
```
3. that's why the port acl was added
4. again, port acl is there for that
```
ACL by IP, port pair is not solve issue itself. Still required external daemon to manage ACL if Kamailio IP address is changed.
PR solves this and requires external daemon.

sergey-safarov:
Also ACL `IP+PORT` implementation not allow use TCP and TLS transport on kamailio side to deliver call to FreeSwitch.
That because Kamailio use random source port. [More details](https://github.com/kamailio/kamailio/issues/2092). That may be a stuck for calls with `sips` uri schema.

P.S. That about general kamailio usage. Not only to kazoo usage case.

lazedo:
try this @sergey-safarov 
**acl.conf.xml**
```
<configuration name="acl.conf" description="Network Lists">
	<network-lists>
		<list name="authoritative" default="deny">
                    <node type="allow" cidr="10.0.0.0/8" ports="5060,7000" />
                    <node type="allow" cidr="172.16.0.0/12" ports="5060,7000" />
                    <node type="allow" cidr="192.168.0.0/16" ports="5060,7000" />
		</list>
		<list name="freeswitch" default="deny">
                    <node type="allow" cidr="10.0.0.0/8" ports="11000" />
                    <node type="allow" cidr="172.16.0.0/12" ports="11000" />
                    <node type="allow" cidr="192.168.0.0/16" ports="11000" />
		</list>
	</network-lists>
</configuration>
```
**sip_profile.xml**
```
<?xml version="1.0"?>
<include>
    <profile name="sipinterface_1">
        <settings>
            ...
            <param name="apply-inbound-acl" value="freeswitch"/>
            <param name="apply-proxy-acl" value="authoritative"/>
            <param name="use-port-for-acl-check" value="true"/>
            ...
```

sergey-safarov:
Tnaks Luis @lazedo for fine example.
Yes, this will work kazoo cluster. Because kazoo use UDP transport for internode communications.
Yes, in this case not need external ACL management application.

Could we also discuss other kamailio+FS installation.
How about calls via TCP/TLS transport from kamailio to FS - used random source port on kamailio side.
And how about call that use `sips` URI - equipment may decline call if used non secure transport or `Via` headers contain information about  non secure transports.

lazedo:
@sergey-safarov i think you're deviating from the purpose of the PR.  i was just pointing out that what you want can be achieved without changes, this is just my opinion, i don't have nothing in favor or against this pr as long as it doesn't break anything, i haven't done any tests on this.

PR #45, opened 2019-10-09
[mod_event_multicast] Fix memory leak

micmac1:
Taken from FS-11193:

In "event_handler()" function, when encrypting data using openssl 1.1.0,
the context ctx is allocated by "EVP_CIPHER_CTX_new()", then data is
encrypted, and at the end, EVP_CIPHER_CTX_cleanup is called. This
function resets the context (the function itself is deprecated, and has
been renamed to "EVP_CIPHER_CTX_reset) so that it can be used again. The
correct call would be to "EVP_CIPHER_CTX_free()", which frees the
memory. The code for openssl 1.0 is OK, since the "ctx" struct is kept
in stack.  The same thing happens during decryption. "ctx" is allocated,
but never freed.

Signed-off-by: Eneas U de Queiroz <cote2004-github@yahoo.com>
Signed-off-by: Sebastian Kemper <sebastian_ml@gmx.net>


micmac1:
Updated commit message as per Andrey's gusto.

PR #70, opened 2019-10-22
[build] OSX Catalina configure does not support --extra-cflags and fix broken compiler option test for -Wno-extended-offsetof that was used to fix older osx clang issue

mjerris:
Need to confirm that this does not negatively impact debian builds

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -586,7 +586,7 @@ libs/libzrtp/libzrtp.a:
 	cd libs/libzrtp && $(MAKE)
 
 libs/libvpx/Makefile: libs/libvpx/.update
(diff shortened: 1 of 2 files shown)


MarioG-X:
Just tested master and it failed with the same error. Make attached. Oddly, when I do a "git log" the top commit is from Oct 15: 7fdf8584e45438ef729f54254fc0540693adf87a, not today, unless I need to do something other than git log. usr/local was erased so everything is new today.

[07 FreeSWITCH make.txt](https://github.com/signalwire/freeswitch/files/3758047/07.FreeSWITCH.make.txt)

andywolk:
@MarioG-X This fix is not yet merged. Use the `osx-catalina-build` branch: `git clone https://github.com/signalwire/freeswitch.git -b osx-catalina-build`

MarioG-X:
That did the trick. It appears to build fine and started right up!

MarioG-X:
Master tested good.

Issue #68, opened 2019-10-21
[mod_amqp] binding_key parameter works only for commands with weird behaviour


jpVm5jYYRE1VIKL:
Previously noticed inside discussion at  [#59](https://github.com/signalwire/freeswitch/issues/59#issuecomment-543588417)
Look like parameter "binding_key"  works only for commands and does not work for producers and logging nodes. 
Beside this it is not possible to setup queue name for binding because it is not implemented.  For commands node FS just create random titled queue and bind to it using "binding_key" routing key. it is also not possible to set type of binded queue. by default command queue have a parameter "auto delete" which not always good but at least it is possible to live with it in case if it have more predictable queue names.

mjerris:
happy to review pull requests to address this

cdosoftei:
This issue touches on a couple of different aspects of mod_amqp.

(1) `binding_key` only works for commands, i.e. does not work for producers and logging.

That is by design, the commands portion is, primarily, an AMQP consumer whereas the producer and logging portions are AMQP producers. Therefore, the binding key concept would be applicable to commands but not to the others; however, the latter employ the counterpart concept: _routing keys_. They are constructed dynamically, as follows:

[Logging](https://github.com/cdosoftei/freeswitch/blob/5f628c009053a5419617cba4259c7cdb9d6a169a/src/mod/event_handlers/mod_amqp/mod_amqp_logging.c#L93): `<hostname>.<entity_id_or_NULL>.<log_level>.<source_file>`

e.g.
```
Exchange:    TAP.Logging
Routing Key: localhost.c044bae6-b685-49b7-a689-4e64e1e1313b.INFO.switch_rtp_c
Payload:     {
                "file":"switch_rtp.c",
	        "function":"switch_rtp_add_crypto_key",
	        "line":4190,
	        "level":"INFO",
	        "timestamp":"2019-10-22 17:56:24.800192",
	        "timestamp_epoch":1.571767e+09,
	        "content":" Activating audio Secure RTP RECV\n"
            }
```

[Producer](https://github.com/cdosoftei/freeswitch/blob/5f628c009053a5419617cba4259c7cdb9d6a169a/src/mod/event_handlers/mod_amqp/mod_amqp_producer.c#L41-L69): `FreeSWITCH.<hostname>.<event>..[entity_id]`

e.g.
```
Exchange:    TAP.Events
[...]

PR #71, opened 2019-10-22
[mod_amqp] configurable commands queue properties

cdosoftei:
The current code base creates the commands queue in auto-delete mode by default; in addition, other available [queue properties](https://www.rabbitmq.com/amqp-0-9-1-reference.html#class.queue) are also hardcoded.

This PR exposes the said properties via module's configuration file, while retaining backwards compatibility by defaulting the values so they match the previous behaviour.

This PR partly addresses #68


crienzo:
@mjerris this looks OKs to me

pnepus:
> This PR exposes the said properties via module's configuration file, while retaining backwards compatibility by defaulting the values so they match the previous behaviour.

how about Optional Arguments ("x-arguments")? 
is there a way to add queue argument like "x-queue-type=quorum"?

PR #75, opened 2019-10-25
execute_on_audio_change on codec related changes

surendrasignalwire:
`execute_on_audio_change` is used to set a callback on audio parameters changes during sdp renegotiation. Rightnow, `execute_on_audio_change` is getting executed only when remote-rtp-ip and remote-rtp-port getting changed during reinvite.

This fix is provided to have `execute_on_audio_change` for codec related changes also.


surendrasignalwire:
closing here as not relevant.

PR #78, opened 2019-10-25
[mod_sndfile] don't force channels and sample rate with .raw extension

dragos-oancea:


src/mod/formats/mod_sndfile/mod_sndfile.c
```diff
@@ -117,7 +117,9 @@ static switch_status_t sndfile_file_open(switch_file_handle_t *handle, const cha
 		context->sfinfo.format |= map->format;
 	}
 
-	if (!strcmp(ext, "r8") || !strcmp(ext, "raw")) {
+	if (strcmp(ext, "raw")) {
```
cdosoftei: Should this read
```c
if (!strcmp(ext, "raw")) {
```
instead?

The reported segfault is averted either way though!

dragos-oancea: if (!strcmp(ext, "raw")) , thanks. I updated.

PR #87, opened 2019-10-31
[mod_sofia] allow copy only some custom headers prefixes

lazedo:
allows copying only some prefixes instead of all custom headers


mjerris:
need a unit test for this

lazedo:
@mjerris the reason for `sip_copy_custom_headers_prefixes` was to preserve the current use of `sip_copy_custom_headers` which people may already be using it with `true`

lazedo:
@mjerris i reused the variable. 

- if  `true` or `not set` will copy all custom headers
- if `false` will not copy any custom headers
- if  `other value` with only copy the custom headers that start with `other value`, a list using `|` as separator may be provided

this preserves existing behavior

lazedo:
> please add unit test on this to confirm functionality.

there's nothing in freeswitch that allows a test to be implemented that hits that code. copying the code to mod_sofia test folder doesn't guarantee that the code isn't changed.

PR #88, opened 2019-10-31
[core] simplify raw print

lazedo:
raw print is breaking when nested


lazedo:
main intention was to remove the possible double free from raw print. after verified that upstream already has done this, freewitch team feels that its time to update cJSON in `fs core` from either `upstream` or `libks` (great!!).
closing this

PR #97, opened 2019-11-03
mod_xml_radius, mod_rad_auth, mod_radius_cdr: now use system freeradius-client

sergey-safarov:
For CentOS 7 need rebuild freeradius-client and place to RPM repo.

For debian may be need create binary package and place to repo. Or simple wait when updated on dist.

### DIFF
--- a/freeswitch.spec
+++ b/freeswitch.spec
@@ -519,6 +519,8 @@ while calls are in progress.
 Summary:	FreeSWITCH mod_rad_auth
 Group:          System/Libraries
 Requires:       %{name} = %{version}-%{release}
+Requires:       freeradius-client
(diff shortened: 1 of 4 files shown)


s3rj1k:
@sergey-safarov Please confirm, you propose to drop freeradius-client bootstrap build code and use distro specific package instead?

sergey-safarov:
yes,
RPM-based dist already has required freeradius-client except CentOS7.
For CEntOS7 I suggest create the package and please into FreeSwitchrepo. Like it did for sofia-sip RPM package.

sergey-safarov:
Could you look at this PR?
@andywolk

Issue #105, opened 2019-11-06
FS depends on deprecated OpenSSL interfaces


micmac1:
Hello all,

FS fails to build when openssl 1.1(.1d) is compiled with "no-deprecated". The openssl 1.1.1 release notes ([here](https://www.openssl.org/news/cl111.txt)) say that "no-deprecated" will 

> restrict the build to just the latest API, rather than a fixed API version.

In my opinion it's not good to use that, then. I mean who can always keep up with the latest API? But they suggest that

>applications are strongly advised to compile their source files with -DOPENSSL_API_COMPAT=0x10100000L, which hides the declarations of all interfaces deprecated in 0.9.8, 1.0.0 or the 1.1.0 releases.

Sidenote: You can also build the openssl library with "--api=1.1.0", which seems to make more sense than using "no-deprecated".

Anyway, with FS the difference between compiling openssl with "no-deprecated" and compiling FS with "-DOPENSSL_API_COMPAT=0x10100000L" seems small. With the former the build fails at the configure stage:
```
checking for openssl... yes
checking openssl_CFLAGS... -I/home/sk/tmp/openwrt/staging_dir/target-mips_24kc_musl/usr/include
checking openssl_LIBS... -L/home/sk/tmp/openwrt/staging_dir/target-mips_24kc_musl/usr/lib -lssl -lcrypto
  adding "-DHAVE_OPENSSL" to SWITCH_AM_CFLAGS
checking for SSL_CTX_set_tlsext_use_srtp in -lssl... yes
checking for DTLSv1_method in -lssl... no
configure: error: OpenSSL >= 1.0.1e and associaed developement headers required
```

With the latter it fails a bit later:

```
  CC       src/libfreeswitch_la-switch_core_cert.lo
src/switch_core_cert.c: In function 'switch_ssl_ssl_lock_callback':
src/switch_core_cert.c:41:13: error: 'CRYPTO_LOCK' undeclared (first use in this function); did you mean 'CRYPTO_RWLOCK'?
  if (mode & CRYPTO_LOCK) {
             ^~~~~~~~~~~
             CRYPTO_RWLOCK
src/switch_core_cert.c:41:13: note: each undeclared identifier is reported only once for each function it appears in
src/switch_core_cert.c: In function 'switch_ssl_init_ssl_locks':
[...]

PR #114, opened 2019-11-10
Update to current OpenSSL APIs

micmac1:
Hi all,

This is a follow-up to #105. The PR breaks with old OpenSSL releases. It updates the baseline to OpenSSL 1.0.2 and adds support for OpenSSL 1.1.0.

I'm not a programmer, so I assume some things may not be perfect. But maybe this PR can serve as a stepping stone.

I compile-tested this against OpenSSL 1.1.1d three different ways:

1. OpenSSL compiled with "no-deprecated"
2. OpenSSL compiled without "no-deprecated" and "OPENSSL_API_COMPAT=0x10002000L" defined in FreeSWITCH's CFLAGS and CPPFLAGS (in hindsight this may have been gratuitous because the patches check OPENSSL_VERSION_NUMBER)
3. OpenSSL compiled without "no-deprecated" and "OPENSSL_API_COMPAT=0x10100000L" defined in FreeSWITCH's CFLAGS and CPPFLAGS

I run-tested this on a mips box with OpenSSL 1.1.1d. I setup two phones registering to FS. Both were able to use SRTP without issues.

```
2019-11-10 19:43:46.523847 [INFO] switch_rtp.c:4208 Activating audio Secure RTP SEND
2019-11-10 19:43:46.543737 [DEBUG] switch_core_sqldb.c:2799 Secure Type: srtp:sdes:AES_CM_256_HMAC_SHA1_80
2019-11-10 19:43:46.543737 [INFO] switch_rtp.c:4186 Activating audio Secure RTP RECV
2019-11-10 19:43:46.543737 [DEBUG] switch_core_sqldb.c:2799 Secure Type: srtp:sdes:AES_CM_256_HMAC_SHA1_80
```

On both phones Linphone showed the lock on the upper right corner.

Then I configured TLS on the server and had one phone register with TLS. I told Linphone not to verify neither certificate nor cn (you can't setup the CA pem in Android app directly, I needed to get an ADB shell and I didn't feel like playing around with this for an hour).

Linphone was able to register successfully:

```
 send 440 bytes to tls/[192.168.0.120]:49198 at 19:43:38.002055:
------------------------------------------------------------------------
SIP/2.0 200 OK
v:SIP/2.0/TLS 192.168.0.120:49198;alias;branch=z9hG4bK.AJVeuQKUu;rport=49198
From: <sip:301@192.168.0.1>;tag=2CaNlWS4E
t:sip:301@192.168.0.1;tag=BaBt9HFpagFFB
Call-ID: lSoaFTRPL6
CSeq: 21 REGISTER
m:<sip:301@192.168.0.120:49198;transport=tls>;expires=3600
Date:Sun, 10 Nov 2019 18:43:37 GMT
User-Agent:FreeSWITCH
Allow:INVITE,ACK,BYE,CANCEL,OPTIONS,MESSAGE,INFO,UPDATE,REGISTER,NOTIFY
k:timer,path,replaces
l:0
```

I dumped the traffic and had a look. TLSv1.2 was used:

```
Transport Layer Security
    TLSv1.2 Record Layer: Handshake Protocol: New Session Ticket
        Content Type: Handshake (22)
        Version: TLS 1.2 (0x0303)
        Length: 202
```

Android had offered TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256 up top, which OpenSSL accepted.

Kind regards,
Seb

### DIFF
--- a/src/switch_core_cert.c
+++ b/src/switch_core_cert.c
@@ -46,9 +46,9 @@ static inline void switch_ssl_ssl_lock_callback(int mode, int type, char *file, 
 	}
 }
 
-static inline unsigned long switch_ssl_ssl_thread_id(void)
+static inline void switch_ssl_ssl_thread_id(CRYPTO_THREADID *id)
 {
-	return (unsigned long) switch_thread_self();
+	CRYPTO_THREADID_set_numeric(id, (unsigned long)switch_thread_self());
 }
 
 SWITCH_DECLARE(void) switch_ssl_init_ssl_locks(void)
@@ -69,7 +69,7 @@ SWITCH_DECLARE(void) switch_ssl_init_ssl_locks(void)
 			switch_assert(ssl_mutexes[i] != NULL);
 		}
 
-		CRYPTO_set_id_callback(switch_ssl_ssl_thread_id);
+		CRYPTO_THREADID_set_callback(switch_ssl_ssl_thread_id);
 		CRYPTO_set_locking_callback((void (*)(int, int, const char*, int))switch_ssl_ssl_lock_callback);
 	}
 
@@ -82,6 +82,7 @@ SWITCH_DECLARE(void) switch_ssl_destroy_ssl_locks(void)
 
 	if (ssl_count == 1) {
 		CRYPTO_set_locking_callback(NULL);
+		CRYPTO_THREADID_set_callback(NULL);
 		for (i = 0; i < CRYPTO_num_locks(); i++) {
 			if (ssl_mutexes[i]) {
 				switch_mutex_destroy(ssl_mutexes[i]);
--- a/src/switch_core_cert.c
+++ b/src/switch_core_cert.c
@@ -32,6 +32,7 @@
 #include <switch.h>
 #include <switch_ssl.h>
 
+#ifndef HAVE_OPENSSL_1_1_API
 static switch_mutex_t **ssl_mutexes;
 static switch_memory_pool_t *ssl_pool = NULL;
 static int ssl_count = 0;
@@ -97,6 +98,7 @@ SWITCH_DECLARE(void) switch_ssl_destroy_ssl_locks(void)
 		switch_core_destroy_memory_pool(&ssl_pool);
 	}
 }
+#endif /* HAVE_OPENSSL_1_1_API */
 
 static const EVP_MD *get_evp_by_name(const char *name)
 {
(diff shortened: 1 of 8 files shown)


micmac1:
Forgot to mention: I looked how libmariadb handles OpenSSL API compatibility while doing these patches.

And one other thing: `tls_init_once()` in "tport_tls.c" doesn't currently check the return value of `OPENSSL_init_ssl()`. It probably should, but `tls_init_once()` is a void function so it didn't make sense to me. Maybe this should be looked into later.

src/switch_core_cert.c
```diff
@@ -69,7 +70,7 @@ SWITCH_DECLARE(void) switch_ssl_init_ssl_locks(void)
 			switch_assert(ssl_mutexes[i] != NULL);
 		}
 
-		CRYPTO_set_id_callback(switch_ssl_ssl_thread_id);
+		CRYPTO_THREADID_set_callback(switch_ssl_ssl_thread_id);
```
placebokkk: From  open ssl 1.1.0, this two MACROs ard defined as empty. So the function switch_ssl_ssl_thread_id and switch_ssl_ssl_lock_callback will not be used and will trigger the W-unused-function errors in Clang. (Gcc will ignore this warning for static inline, but clang will not). Then Clang build will fail. Could you also fix this problem in your PR?

mjerris: if that is actually unused does this mean me should be doing something differently here ?

micmac1:
Hi Chao Yang, will do tonight. Thanks for the feedback!

src/switch_core_cert.c
micmac1: > From open ssl 1.1.0, this two MACROs ard defined as empty. So the function switch_ssl_ssl_thread_id and switch_ssl_ssl_lock_callback will not be used and will trigger the W-unused-function errors in Clang. (Gcc will ignore this warning for static inline, but clang will not). Then Clang build will fail. Could you also fix this problem in your PR?

Hi again, I looked at it but it was over my head. These functions/macros are used in various places and if they are ifdef'ed that should probably include other functions as well (CRYPTO_set_locking_callback for instance?).

micmac1:
Hello @neheb 

Thanks for feedback. You might be right, I don't know. I believe this is also a bit outdated now. For instance sofia-sip is now an external lib.

Anyhow, I can't pretend to possess the expertise nor the time to pursue this further. I'll close this pull request.

Best regards,
Seb

PR #132, opened 2019-11-21
RFC4867 compliancy

jay98237438:
According to RFC4867, offered mode-set must be returned or rejected. It can't be altered in the response.


dragos-oancea:
Can you show the exact part of the RFC that pertains to your change ? Also, if it's really an issue it should be fixed in amr-wb too.

jay98237438:
RFC4867:
If a mode set was supplied in the offer, the answerer
         SHALL return the mode-set unmodified or reject the payload
         type.

jay98237438:
Pardon my double commits. I have never used github before.

dragos-oancea:
you will have to squash those in one single commit or maybe two commits (one for each module) . Also, afair, AMR-WB has one extra mode.
"for (i = 0; 8 > i; ++i) {" - I think it should be 9 instead of 8 for AMR-WB.

jay98237438:
sending a new pull request since github web dont seem to support file merge

PR #133, opened 2019-11-21
[mod_amr, mod_amrwb] RFC4867 mode-set compliancy

jay98237438:
RFC4867 says:
If a mode set was supplied in the offer, the answerer SHALL return the mode-set unmodified or reject the payload type

FS was replying with its own default mode-set even if it was offered one. This causes interoperability issues.


dragos-oancea:
you'll need to format the commit messages with the module name in them, eg: "[mod_amr] RFC4867 mode-set compliancy "

jay98237438:
fixed

crienzo:
@andywolk dragos says this one is OK

PR #137, opened 2019-11-25
[mod_callcenter] Adding SAF_SUPPORT_NOMEDIA to callcenter_track since we don't need media up for it. Also, a small fix to make it compile on OSX.

italorossi:
Adding SAF_SUPPORT_NOMEDIA to callcenter_track since we don't need media up for it. 

Also, a small fix to make it compile on OSX.


crienzo:
@andywolk this is ok to merge

Issue #140, opened 2019-11-26
[mod_callcenter] agent-del event action is absent


ukolovda:
It will useful to have event with cc_action `agent-del ` (in addition to `agent-add`).

PR #141, opened 2019-11-26
[mod_callcenter] Append firing event with cc_action `agent-del`.

ukolovda:
See https://github.com/signalwire/freeswitch/issues/140

### DIFF
--- a/src/mod/applications/mod_callcenter/mod_callcenter.c
+++ b/src/mod/applications/mod_callcenter/mod_callcenter.c
@@ -945,7 +945,7 @@ cc_status_t cc_agent_del(const char *agent)
 
   if (deleted_row_count > 0) {
 
-  	switch_log_printf(SWITCH_CHANNEL_LOG, SWITCH_LOG_DEBUG, "Deleted Agent %s\n", agent);
+		switch_log_printf(SWITCH_CHANNEL_LOG, SWITCH_LOG_DEBUG, "Deleted Agent %s\n", agent);
 
 		if (switch_event_create_subclass(&event, SWITCH_EVENT_CUSTOM, CALLCENTER_EVENT) == SWITCH_STATUS_SUCCESS) {
 			switch_event_add_header_string(event, SWITCH_STACK_BOTTOM, "CC-Agent", agent);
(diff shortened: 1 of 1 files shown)


mochouinard (dismissed):
I would add check to see if that agent existed before generating an delete event.  Also, maybe triggering delete for every tier would be good(or bad...)  But I guess this alone is fine.

ukolovda:
Append checking for agent existing when deleting.
Also tiers now deleted before agent.
But it have interface change. Before it returns OK regardless agent exists or not. Now it return "-Err Agent not found" if we try delete unexisting agent.

mochouinard:
I guess this could go in as is... But I must say I don't like how deleting agent was coded in the first place.  It missing so much check and it can cause problems pretty quickly !  That guy really was lazy when he made that code! 
Full disclosure... I made that code ;)
What I would really do here today is these check : 

- First check if agent exist and his the agent status is CC_AGENT_STATUS_LOGGED_OUT and that his state is CC_AGENT_STATE_WAITING.   If not, then return error accordingly so you know why it failed.  So you can easily do individual error message got event doesn't exist, Status is not logged out and State is not Waiting.  Because right now, it would delete stuff during an active call or while an agent is ringing and it will cause issues.

- Then, I would delete the agent with these check again for status and state.  I would delete the agent first.  Check for result number of row affected, if it 0, the return error about potential invalid agent status/state.

- If agent was deleted, then delete the tiers associated to that agent.

This might fix some potential issues you might encounter as well if you start to use the agent delete function.

Let me know what you think.

mochouinard:
In case you haven't seen, I've sent you a pull request on your repo for this patch.  Let me know what you think.  This might help you if you actively use delete agent from issues if that agent is logged in or on a call. Another feature that could be added in the future is a way to send an agent pending for deletion... So you it couldn't received any new call, would allow current call to complete and then it would be destroyed correctly...  But I think to make sure the agent is logged off and not on a call our self is sufficient.

Len-PGH:
This PR is older than 6 months and requires some attention to be merged, so we’re closing this issue as inactive. If you still want this submission accepted, please make sure to update your code to the latest version (this includes FreeSWITCH, sofia-sip, libks, and any other dependencies your code touches), double check that you’re following the [submission guidelines](https://github.com/signalwire/libks/blob/master/CODING_GUIDELINES), and make any revisions/updates suggested by the team. 

We’re doing a [PR Review Day](https://mailchi.mp/freeswitch/pr-review-day) event on March 21st. You’re welcome to make the necessary changes or join with any questions. We'll have team members on hand to provide immediate feedback, work through any issues that come up, and to commit changes that pass review during the event. 

We hope to see you there! If you have any questions or feedback about this event or just want some more information, you can drop in to our [Community Slack](https://signalwire.com/developers/slack-community). 

Note: There will be a second Review Day in April if you cannot make this time.

PR #150, opened 2019-12-01
[docs/phrase] Translated all freeswitch phrases to German

nkssk:
Translated all phrases in phrase_en.xml to phrase_de.xml
Removed duplicate or conflicting phrase definitions from phrase_en.xml


dragos-oancea:
can you squash into 1 commit and prefix the commit message with something like "[docs]" ?

Issue #152, opened 2019-12-02
[mod_callcenter] add options for not-registered-delay-time, max rejected, max busy and max_not_registered


wmasilva:
add extra options to agent hangup causes:

not-registered-delay-time - delay time if the agent is not registered

max-rejected - maximum number of calls rejected to auto-pause the agent
max-busy -  maximum number of calls hangup with busy to auto-pause the agent
max-not-registered  -  maximum number of calls hangup with busy to auto-pause the agent

PR #153, opened 2019-12-02
[mod_callcenter] add options for not-registered-delay-time, max r…

wmasilva:
…ejected, max busy and max_not_registered #152

### DIFF
--- a/src/mod/applications/mod_callcenter/mod_callcenter.c
+++ b/src/mod/applications/mod_callcenter/mod_callcenter.c
@@ -118,6 +118,15 @@ static struct cc_status_table AGENT_STATUS_CHART[] = {
 
 };
 
+static switch_xml_config_enum_item_t config_cc_agent_status[] = {
+	{"Unknown", CC_AGENT_STATUS_UNKNOWN},
+	{"Logged Out", CC_AGENT_STATUS_LOGGED_OUT},
+	{"Available", CC_AGENT_STATUS_AVAILABLE},
+	{"Available (On Demand)", CC_AGENT_STATUS_AVAILABLE_ON_DEMAND},
+	{"On Break", CC_AGENT_STATUS_ON_BREAK},
+	{NULL, 0}
+};
+
 typedef enum {
 	CC_AGENT_STATE_UNKNOWN = 0,
 	CC_AGENT_STATE_WAITING = 1,
@@ -225,17 +234,23 @@ static char agents_sql[] =
    Receiving
    In a queue call
  */
-
-"   max_no_answer INTEGER NOT NULL DEFAULT 0,\n"
 "   wrap_up_time INTEGER NOT NULL DEFAULT 0,\n"
-"   reject_delay_time INTEGER NOT NULL DEFAULT 0,\n"
-"   busy_delay_time INTEGER NOT NULL DEFAULT 0,\n"
+"   max_no_answer INTEGER NOT NULL DEFAULT 0,\n"
+"   no_answer_count INTEGER NOT NULL DEFAULT 0,\n"
 "   no_answer_delay_time INTEGER NOT NULL DEFAULT 0,\n"
+"   max_rejected INTEGER NOT NULL DEFAULT 0,\n"
+"   rejected_count INTEGER NOT NULL DEFAULT 0,\n"
+"   reject_delay_time INTEGER NOT NULL DEFAULT 0,\n"
+"   max_busy INTEGER NOT NULL DEFAULT 0,\n"
+"   busy_count INTEGER NOT NULL DEFAULT 0,\n"
+"   busy_delay_time INTEGER NOT NULL DEFAULT 0,\n"
+"   max_not_registered INTEGER NOT NULL DEFAULT 0,\n"
+"   not_registered_count INTEGER NOT NULL DEFAULT 0,\n"
+"   not_registered_delay_time INTEGER NOT NULL DEFAULT 5,\n"
 "   last_bridge_start INTEGER NOT NULL DEFAULT 0,\n"
 "   last_bridge_end INTEGER NOT NULL DEFAULT 0,\n"
 "   last_offered_call INTEGER NOT NULL DEFAULT 0,\n"
 "   last_status_change INTEGER NOT NULL DEFAULT 0,\n"
-"   no_answer_count INTEGER NOT NULL DEFAULT 0,\n"
 "   calls_answered  INTEGER NOT NULL DEFAULT 0,\n"
 "   talk_time  INTEGER NOT NULL DEFAULT 0,\n"
 "   ready_time INTEGER NOT NULL DEFAULT 0,\n"
@@ -258,6 +273,52 @@ static char tiers_sql[] =
  */
 "   level    INTEGER NOT NULL DEFAULT 1,\n"
 "   position INTEGER NOT NULL DEFAULT 1\n" ");\n";
+
+struct agent_info_callback {
+	switch_bool_t found;
+
+	int no_answer_count;
+	int max_no_answer;
+	int no_answer_delay_time;
+
+	int rejected_count;
+	int max_rejected;
+	int reject_delay_time;
+
+	int busy_count;
+	int max_busy;
+	int busy_delay_time;
+
+	int not_registered_count;
+	int max_not_registered;
+	int not_registered_delay_time;
+};
+typedef struct agent_info_callback agent_info_callback_t;
+
+static int agent_info_callback(void *pArg, int argc, char **argv, char **columnNames)
+{
+	agent_info_callback_t *cbt = (agent_info_callback_t *) pArg;
+
+	cbt->found = SWITCH_TRUE;
+
+	cbt->no_answer_count = atoi(argv[0]);
+	cbt->max_no_answer = atoi(argv[1]);
+	cbt->no_answer_delay_time = atoi(argv[2]);
+
+	cbt->rejected_count = atoi(argv[3]);
+	cbt->max_rejected = atoi(argv[4]);
+	cbt->reject_delay_time = atoi(argv[5]);
+
+	cbt->busy_count = atoi(argv[6]);
+	cbt->max_busy = atoi(argv[7]);
+	cbt->busy_delay_time = atoi(argv[8]);
+
+	cbt->not_registered_count = atoi(argv[9]);
+	cbt->max_not_registered = atoi(argv[10]);
+	cbt->not_registered_delay_time = atoi(argv[11]);
+
+	return 1; /* Single Row Result */
+}
 
 static switch_xml_config_int_options_t config_int_0_86400 = { SWITCH_TRUE, 0, SWITCH_TRUE, 86400 };
 
@@ -463,9 +524,12 @@ struct cc_queue {
 	uint32_t max_wait_time;
 	uint32_t max_wait_time_with_no_agent;
 	uint32_t max_wait_time_with_no_agent_time_reached;
-	char *agent_no_answer_status;
 	uint32_t calls_answered;
 	uint32_t calls_abandoned;
+	cc_agent_status_t agent_no_answer_status;
+	cc_agent_status_t agent_rejected_status;
+	cc_agent_status_t agent_busy_status;
+	cc_agent_status_t agent_not_registered_status;
 
 	switch_mutex_t *mutex;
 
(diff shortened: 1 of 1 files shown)


src/mod/applications/mod_callcenter/mod_callcenter.c
```diff
@@ -850,6 +878,16 @@ struct call_helper {
...
+	int rejected_count;
+	int busy_count;
+	int not_registered_count;
+	int max_rejected;
+	int max_busy;
+	int max_not_registered;
+	int not_registered_delay_time;
+	cc_agent_status_t agent_rejected_status;
+	cc_agent_status_t agent_busy_status;
+	cc_agent_status_t agent_not_registered_status;
```
mochouinard: I would just reposition them so they are grouped with their respective similar value (max, status, count together)

src/mod/applications/mod_callcenter/mod_callcenter.c
```diff
@@ -1344,6 +1403,10 @@ static switch_status_t load_agent(const char *agent_name, switch_event_t *params
 		const char *reject_delay_time = switch_xml_attr(x_agent, "reject-delay-time");
 		const char *busy_delay_time = switch_xml_attr(x_agent, "busy-delay-time");
 		const char *no_answer_delay_time = switch_xml_attr(x_agent, "no-answer-delay-time");
+		const char *max_rejected = switch_xml_attr(x_agent, "max-rejected");
+		const char *max_busy = switch_xml_attr(x_agent, "max-busy");
+		const char *max_not_registered = switch_xml_attr(x_agent, "max-not-registered");
+		const char *not_registered_delay_time = switch_xml_attr(x_agent, "not-registered-delay-time");
```
mochouinard: Same here, group them with their respective group.

src/mod/applications/mod_callcenter/mod_callcenter.c
```diff
@@ -1678,6 +1761,68 @@ static switch_status_t playback_array(switch_core_session_t *session, const char
...
+		default:
+			column_count = "no_answer_count";
+			event_cc_action = "agent-max-no-answer";
+			event_cc_header_count = "CC-Agent-No-Answer-Count";
+			event_cc_header_status = "CC-Agent-No-Answer-Status";
+			break;
+	}
+
+	/* Update Agent column count */
+	sql = switch_mprintf("UPDATE agents SET %q = %q + 1 WHERE name = '%q';", column_count, column_count, agent_name);
```
mochouinard:
```suggestion
	sql = switch_mprintf("UPDATE agents SET %s = %s + 1 WHERE name = '%q';", column_count, column_count, agent_name);
```
Very little performance increase ;)  If you rather have %q then it wouldn't bother me

src/mod/applications/mod_callcenter/mod_callcenter.c
```diff
@@ -2019,7 +2164,8 @@ static void *SWITCH_THREAD_FUNC outbound_agent_thread_run(switch_thread_t *threa
 			switch_channel_set_variable_printf(member_channel, "cc_queue_answered_epoch", "%" SWITCH_TIME_T_FMT, local_epoch_time_now(NULL));
 			/* Set UUID of the Agent channel */
 			sql = switch_mprintf("UPDATE agents SET uuid = '%q', last_bridge_start = '%" SWITCH_TIME_T_FMT "', calls_answered = calls_answered + 1, no_answer_count = 0"
-										 " WHERE name = '%q' AND instance_id = '%q'",
+								 " rejected_count = 0, busy_count = 0, not_registered_count = 0"
```
mochouinard:
```suggestion
								 ", rejected_count = 0, busy_count = 0, not_registered_count = 0"
```

src/mod/applications/mod_callcenter/mod_callcenter.c
```diff
@@ -4287,6 +4450,7 @@ SWITCH_MODULE_LOAD_FUNCTION(mod_callcenter_load)
 	switch_console_set_complete("add callcenter_config queue count tiers");
 
 	switch_console_set_complete("add callcenter_break agent");
+	switch_console_set_complete("add callcenter_config agent set not_registered_delay_time");
```
mochouinard: Move it with the other callcenter_config agent

wmasilva:
Sorry for the delay.

i apply all the changes you told me and to better detect what is going to be merge i reset the branch to a single commit like in stash probably with github no need to do it any more.

src/mod/applications/mod_callcenter/mod_callcenter.c
```diff
@@ -1168,7 +1205,27 @@ cc_status_t cc_agent_update(const char *key, const char *value, const char *agen
...
+		sql = switch_mprintf("UPDATE agents SET max_busy = '%d', instance_id = 'single_box' WHERE name = '%q'", atoi(value), agent);
+		cc_execute_sql(NULL, sql, NULL);
+		switch_safe_free(sql);
+		result = CC_STATUS_SUCCESS;
 
+	} else if (!strcasecmp(key, "max_not_registered")) {
+		sql = switch_mprintf("UPDATE agents SET max_not_registered = '%d', instance_id = 'single_box' WHERE name = '%q'", atoi(value), agent);
+		cc_execute_sql(NULL, sql, NULL);
+		switch_safe_free(sql);
+		result = CC_STATUS_SUCCESS;
```
mochouinard: Not required... But all these simple update which have the SQL column have the same name of the parameters could be groupped within the same else if.  I should have done this in the first place when I made it ! 

Also, I haven't fully looked up the code but does these change generate an event of this change ? If it doesn't, then this should be added (Again, not required per say, since it doesn't do it for the other one, but since you are in that code... :p   I could make a pull request inside your repo as well if you want me to do it.

src/mod/applications/mod_callcenter/mod_callcenter.c
```diff
@@ -1678,6 +1759,68 @@ static switch_status_t playback_array(switch_core_session_t *session, const char
 
 	return status;
 }
+static void cc_agent_noanswer_count_status_update(const char *agent_name, const char *member_uuid, const char *member_session_uuid, switch_call_cause_t hangupcause, cc_agent_status_t newstatus, int count, int max)
```
mochouinard: maybe change the noanswer to failed ?

src/mod/applications/mod_callcenter/mod_callcenter.c
```diff
@@ -1678,6 +1759,68 @@ static switch_status_t playback_array(switch_core_session_t *session, const char
...
+		default:
+			column_count = "no_answer_count";
+			event_cc_action = "agent-max-no-answer";
+			event_cc_header_count = "CC-Agent-No-Answer-Count";
+			event_cc_header_status = "CC-Agent-No-Answer-Status";
+			break;
+	}
+
+	/* Update Agent column count */
+	sql = switch_mprintf("UPDATE agents SET %s = %s + 1 WHERE name = '%s';", column_count, column_count, agent_name);
```
mochouinard: name = '%q'
It not needed for the SET %s = %s because we know those will never have invalid stuff in it... But the agent name could have some injection added to it.

src/mod/applications/mod_callcenter/mod_callcenter.c
```diff
@@ -1015,7 +1051,8 @@ cc_status_t cc_agent_update(const char *key, const char *value, const char *agen
 		if (cc_agent_str2status(value) != CC_AGENT_STATUS_UNKNOWN) {
 			/* Reset values on available only */
 			if (cc_agent_str2status(value) == CC_AGENT_STATUS_AVAILABLE) {
-				sql = switch_mprintf("UPDATE agents SET status = '%q', last_status_change = '%" SWITCH_TIME_T_FMT "', talk_time = 0, calls_answered = 0, no_answer_count = 0"
+				sql = switch_mprintf("UPDATE agents SET status = '%q', last_status_change = '%" SWITCH_TIME_T_FMT "', talk_time = 0, calls_answered = 0, no_answer_count = 0,"
```
mochouinard: The coma at the end should be at the beginning of the next line...   Just so we are consistent

src/mod/applications/mod_callcenter/mod_callcenter.c
```diff
@@ -2133,45 +2277,28 @@ static void *SWITCH_THREAD_FUNC outbound_agent_thread_run(switch_thread_t *threa
 			/* Busy: Do Not Disturb, Circuit congestion */
 			case SWITCH_CAUSE_NORMAL_CIRCUIT_CONGESTION:
 			case SWITCH_CAUSE_USER_BUSY:
-				delay_next_agent_call = (h->busy_delay_time > delay_next_agent_call? h->busy_delay_time : delay_next_agent_call);
+				delay_next_agent_call = (h->busy_delay_time > delay_next_agent_call ? h->busy_delay_time : delay_next_agent_call);
+				cc_agent_noanswer_count_status_update(h->agent_name, h->member_uuid, h->member_session_uuid, cause, h->agent_busy_status, 
+										h->busy_count, h->max_busy);
```
mochouinard: I don't like it that we send individually the busy_count and max_busy here...
I'm not sure if we need the busy_count max_busy( and all the other one) any where else.... If not, then I would move in that function a SELECT that check for these values.  This would also make a better behavior, where if you ring during 60 seconds an agent, if you change the agent value during that 60 seconds, the changes won't be applied until the next call.

src/mod/applications/mod_callcenter/mod_callcenter.c
```diff
@@ -3520,7 +3674,8 @@ static int list_result_json_callback(void *pArg, int argc, char **argv, char **c
 "\tcallcenter_config queue count | \n" \
 "\tcallcenter_config queue count agents [queue_name] [status] [state] | \n" \
 "\tcallcenter_config queue count members [queue_name] | \n" \
-"\tcallcenter_config queue count tiers [queue_name]"
+"\tcallcenter_config queue count tiers [queue_name] | \n" \
+"\tcallcenter_config agent set not_registered_delay_time [agent_name] [wait second] | \n"
```
mochouinard:
```suggestion
"\tcallcenter_config agent set not_registered_delay_time [agent_name] [wait second]"
```

mochouinard: Also, instead of putting it at the end, group it with their sibling.  Wasn't there another type you added as well ?

src/mod/applications/mod_callcenter/mod_callcenter.c
```diff
@@ -4356,4 +4518,4 @@ SWITCH_MODULE_SHUTDOWN_FUNCTION(mod_callcenter_shutdown)
  * End:
  * For VIM:
  * vim:set softtabstop=4 shiftwidth=4 tabstop=4 noet
- */
+ */
```
mochouinard: Missing a new line at the end of the file... (It just annoying).

wmasilva:
> Let me know if you want me to make a pull request to your branch with some of the changes I talked about.

sure, go ahead.

src/mod/applications/mod_callcenter/mod_callcenter.c
wmasilva: yes, agree better to "read", changing it.

src/mod/applications/mod_callcenter/mod_callcenter.c
wmasilva: i think i understand what you mean, i will do the change if is not ok,  feel free to change it on my branch. 

no, there is no event generated when changing this parameters, i guess the events should be generated for all settings inside this function (ready_time, wrap_up_time, type, ..) ?
should i do this change now or leave it for another pr?

wmasilva: the idea of replacing the else if, we probably could use a switch/case, to be able to have a default: result = CC_STATUS_INVALID_KEY;

what do you think?

src/mod/applications/mod_callcenter/mod_callcenter.c
wmasilva: ok, finally i understand the use of %q 
:)

src/mod/applications/mod_callcenter/mod_callcenter.c
wmasilva: done

src/mod/applications/mod_callcenter/mod_callcenter.c
wmasilva: here i got lost.. if you could purpose your change, i'm happy to test it locally. 

note: here i had forget to set the not_registered_delay_time and remove the hardcode 5seconds.

src/mod/applications/mod_callcenter/mod_callcenter.c
mochouinard: Yes and no...  To make use of a switch, we would have to generate ENUM for these options (like we did for the agent,  status and states..   In C, you can't do a switch on a string.
I'm not sure if it would bring much values to do so... It could be used to do a loop of all the argument and auto generate the api list.  But I think just using else if is fine.  You just need to capture all of the string we want in the same if condition and then in that if statement change the sql query with the provided parameters that was provided.

src/mod/applications/mod_callcenter/mod_callcenter.c
mochouinard: I changed that in the pull request I've made.  I've removed reference to the different count and max that exist, and loaded them directly in that function from the agent.

areski:
Anything I can do to help on this PR?

briankwest:
@mochouinard can you follow up and let me know if you can approve this or not?

PR #158, opened 2019-12-03
FS-11432 Fix

ovadbar:
FS-11432 
Changed mod_http_cache.c added file_delete file_exist functions 
Change switch_file_core added switch_core_file_exists and switch_core_file_remove 
Changed mod_voicemail to use switch_core_file_exists and switch_core_file_remove


andywolk:
We made some fixes in master. Please rebase to the latest. Tests should be OK.

ovadbar:
I rebased and am still getting a failing unit test for setup_switch_ivr_play_say. This error is also on the master branch.

`setup_switch_ivr_play_say
  switch_ivr_play_say.c(45):
    switch_loadable_module_exists("mod_test") == SWITCH_STATUS_SUCCESS`

andywolk:
If you want to run tests manually, you should enable mod_test in modules.conf when building.

src/mod/applications/mod_http_cache/mod_http_cache.c
```diff
@@ -1176,6 +1178,181 @@ static switch_status_t http_get(url_cache_t *cache, http_profile_t *profile, cac
...
+	if (!profile) {
+		profile = url_cache_http_profile_find_by_fqdn(cache, url->url);
+	}
+
+	if (profile && profile->append_headers_ptr) {
+		headers = profile->append_headers_ptr(profile, headers, "DELETE", 0, "", url->url, 0, NULL);
+	}
+
+	curl_handle = switch_curl_easy_init();
+	switch_log_printf(SWITCH_CHANNEL_SESSION_LOG(session), SWITCH_LOG_DEBUG, "opening %s for URL cache headers\n", del_data.url->filename);
```
mochouinard:
```suggestion
	switch_log_printf(SWITCH_CHANNEL_SESSION_LOG(session), SWITCH_LOG_DEBUG, "Opening %s for URL cache headers\n", del_data.url->filename);
```

src/mod/applications/mod_http_cache/mod_http_cache.c
```diff
@@ -1176,6 +1178,181 @@ static switch_status_t http_get(url_cache_t *cache, http_profile_t *profile, cac
...
+		switch_curl_easy_setopt(curl_handle, CURLOPT_TIMEOUT, cache->download_timeout);
+	}
+	if (!cache->ssl_verifypeer) {
+		switch_curl_easy_setopt(curl_handle, CURLOPT_SSL_VERIFYPEER, 0L);
+	} else {
+		/* this is the file with all the trusted certificate authorities */
+		if (!zstr(cache->ssl_cacert)) {
+			switch_curl_easy_setopt(curl_handle, CURLOPT_CAINFO, cache->ssl_cacert);
+		}
+	}
```
mochouinard: I would move these (and the one in the other function) to a seperate function that actually use the same common setopt instead of duplicating them everytime.  Reduce the risk of future edit of 1 of them without changing them all.

src/mod/applications/mod_http_cache/mod_http_cache.c
```diff
@@ -1300,7 +1477,6 @@ SWITCH_STANDARD_API(http_cache_get)
 
 	return status;
 }
-
```
mochouinard: Surperflous delete empty line

src/mod/applications/mod_http_cache/mod_http_cache.c
```diff
@@ -1773,6 +1949,48 @@ static switch_status_t http_cache_file_open(switch_file_handle_t *handle, const
...
+{
+	http_profile_t *profile;
+	cached_url_t *u = NULL;
+	switch_core_session_t *session = NULL;
+	u = cached_url_create(&gcache, path, NULL);
+	profile = url_cache_http_profile_find_by_fqdn(&gcache, path);
+	switch_log_printf(SWITCH_CHANNEL_LOG, SWITCH_LOG_DEBUG, "Calling http_cache_file_exists on %s\n", u->filename);
+	//if (switch_file_exists(u->filename, gcache.pool) == SWITCH_STATUS_SUCCESS) {
+	//	return SWITCH_STATUS_SUCCESS;
+	//}
```
mochouinard: I think the policy is not to have // comments.

src/mod/applications/mod_voicemail/mod_voicemail.c
```diff
@@ -1148,7 +1148,7 @@ static int sql2str_callback(void *pArg, int argc, char **argv, char **columnName
 static int unlink_callback(void *pArg, int argc, char **argv, char **columnNames)
 {
 	if (argv[0]) {
-		if (unlink(argv[0]) != 0) {
+		if (switch_core_file_remove(argv[0],NULL) != SWITCH_STATUS_SUCCESS) {
```
mochouinard:
```suggestion
		if (switch_core_file_remove(argv[0], NULL) != SWITCH_STATUS_SUCCESS) {
```

src/mod/applications/mod_voicemail/mod_voicemail.c
```diff
@@ -1227,16 +1227,16 @@ static switch_status_t create_file(switch_core_session_t *session, vm_profile_t
 		args.buf = input;
 		args.buflen = sizeof(input);
 
-		unlink(file_path);
+		switch_core_file_remove(file_path,NULL);
```
mochouinard:
```suggestion
		switch_core_file_remove(file_path, NULL);
```

src/mod/applications/mod_voicemail/mod_voicemail.c
```diff
@@ -1227,16 +1227,16 @@ static switch_status_t create_file(switch_core_session_t *session, vm_profile_t
...
 
 		switch_ivr_record_file(session, &fh, file_path, &args, profile->max_record_len);
 
-		if (switch_file_exists(file_path, switch_core_session_get_pool(session)) == SWITCH_STATUS_SUCCESS) {
+		if (switch_core_file_exists(file_path,NULL) == SWITCH_STATUS_SUCCESS) {
```
mochouinard:
```suggestion
		if (switch_core_file_exists(file_path, NULL) == SWITCH_STATUS_SUCCESS) {
```

src/mod/applications/mod_voicemail/mod_voicemail.c
```diff
@@ -1227,16 +1227,16 @@ static switch_status_t create_file(switch_core_session_t *session, vm_profile_t
...
 			got_file = 1;
 		}
 
 		if (limit && (*message_len = fh.samples_out / (fh.samplerate ? fh.samplerate : 8000)) < profile->min_record_len) {
-			if (unlink(file_path) != 0) {
+			if (switch_core_file_remove(file_path,NULL) != SWITCH_STATUS_SUCCESS) {
```
mochouinard:
```suggestion
			if (switch_core_file_remove(file_path, NULL) != SWITCH_STATUS_SUCCESS) {
```

src/mod/applications/mod_voicemail/mod_voicemail.c
```diff
@@ -1557,7 +1557,7 @@ static char *vm_merge_file(switch_core_session_t *session, vm_profile_t *profile
 
 	}
 
-	if (unlink(announce) != 0) {
+	if (switch_core_file_remove(announce,NULL) != SWITCH_STATUS_SUCCESS) {
```
mochouinard:
```suggestion
	if (switch_core_file_remove(announce, NULL) != SWITCH_STATUS_SUCCESS) {
```

src/mod/applications/mod_voicemail/mod_voicemail.c
```diff
@@ -1902,7 +1902,7 @@ static switch_status_t listen_file(switch_core_session_t *session, vm_profile_t
   end:
 
 	if (forward_file_path) {
-		if (unlink(forward_file_path) != 0) {
+		if (switch_core_file_remove(forward_file_path,NULL) != SWITCH_STATUS_SUCCESS) {
```
mochouinard:
```suggestion
		if (switch_core_file_remove(forward_file_path, NULL) != SWITCH_STATUS_SUCCESS) {
```

src/mod/applications/mod_voicemail/mod_voicemail.c
```diff
@@ -2108,6 +2108,8 @@ static void voicemail_check_main(switch_core_session_t *session, vm_profile_t *p
 				listen_callback_t cbt;
 				char sql[512];
 				int cur_message, total_messages;
+				//switch_bool_t keep_vm;
+				//keep_vm = SWITCH_FALSE;
```
mochouinard: Left over Comments

src/mod/applications/mod_voicemail/mod_voicemail.c
```diff
@@ -2276,7 +2278,7 @@ static void voicemail_check_main(switch_core_session_t *session, vm_profile_t *p
 						switch_event_t *params;
 						file_path = switch_mprintf("%s%sgreeting_%d.%s", dir_path, SWITCH_PATH_SEPARATOR, num, profile->file_ext);
 						tmp_file_path = switch_mprintf("%s%sgreeting_%d_TMP.%s", dir_path, SWITCH_PATH_SEPARATOR, num, profile->file_ext);
-						unlink(tmp_file_path);
+						switch_core_file_remove(tmp_file_path,NULL);
```
mochouinard:
```suggestion
						switch_core_file_remove(tmp_file_path, NULL);
```

src/mod/applications/mod_voicemail/mod_voicemail.c
```diff
@@ -2356,7 +2358,7 @@ static void voicemail_check_main(switch_core_session_t *session, vm_profile_t *p
 					switch_event_t *params;
 					file_path = switch_mprintf("%s%srecorded_name.%s", dir_path, SWITCH_PATH_SEPARATOR, profile->file_ext);
 					tmp_file_path = switch_mprintf("%s%srecorded_name_TMP.%s", dir_path, SWITCH_PATH_SEPARATOR, profile->file_ext);
-					unlink(tmp_file_path);
+					switch_core_file_remove(tmp_file_path,NULL);
```
mochouinard:
```suggestion
					switch_core_file_remove(tmp_file_path, NULL);
```

src/mod/applications/mod_voicemail/mod_voicemail.c
```diff
@@ -2681,7 +2683,7 @@ static void voicemail_check_main(switch_core_session_t *session, vm_profile_t *p
 	switch_safe_free(file_path);
 
 	if (tmp_file_path) {
-		unlink(tmp_file_path);
+		switch_core_file_remove(tmp_file_path,NULL);
```
mochouinard:
```suggestion
		switch_core_file_remove(tmp_file_path, NULL);
```

src/mod/applications/mod_voicemail/mod_voicemail.c
```diff
@@ -3141,8 +3143,8 @@ static switch_status_t deliver_vm(vm_profile_t *profile,
 
   failed:
 
-	if (!insert_db && file_path && switch_file_exists(file_path, pool) == SWITCH_STATUS_SUCCESS) {
-		if (unlink(file_path) != 0) {
+	if (!insert_db && file_path && switch_core_file_exists(file_path, pool) == SWITCH_STATUS_SUCCESS) {
+		if (switch_core_file_remove(file_path,NULL) != SWITCH_STATUS_SUCCESS) {
```
mochouinard:
```suggestion
		if (switch_core_file_remove(file_path, NULL) != SWITCH_STATUS_SUCCESS) {
```

src/mod/applications/mod_voicemail/mod_voicemail.c
```diff
@@ -4155,7 +4157,7 @@ struct holder {
 static int del_callback(void *pArg, int argc, char **argv, char **columnNames)
 {
 	if (argc > 8) {
-		if (unlink(argv[8]) != 0) {
+		if (switch_core_file_remove(argv[8],NULL) != SWITCH_STATUS_SUCCESS) {
```
mochouinard:
```suggestion
		if (switch_core_file_remove(argv[8], NULL) != SWITCH_STATUS_SUCCESS) {
```

src/mod/applications/mod_voicemail/mod_voicemail.c
```diff
@@ -4587,9 +4589,9 @@ SWITCH_STANDARD_API(voicemail_inject_api_function)
 static int api_del_callback(void *pArg, int argc, char **argv, char **columnNames)
 {
 
-	unlink(argv[3]);
+	switch_core_file_remove(argv[3],NULL);
```
mochouinard:
```suggestion
	switch_core_file_remove(argv[3], NULL);
```

src/mod/applications/mod_voicemail/mod_voicemail.c
```diff
@@ -5057,7 +5059,7 @@ static int message_purge_callback(void *pArg, int argc, char **argv, char **colu
 	char *sql;
 	vm_profile_t *profile = get_profile(profile_name);
 
-	if (unlink(file_path) != 0) {
+	if (switch_core_file_remove(file_path,NULL) != SWITCH_STATUS_SUCCESS) {
```
mochouinard:
```suggestion
	if (switch_core_file_remove(file_path, NULL) != SWITCH_STATUS_SUCCESS) {
```

src/switch_core_file.c
```diff
@@ -62,6 +62,78 @@ static switch_status_t get_file_size(switch_file_handle_t *fh, const char **stri
...
+	switch_status_t status;
+	char *rhs = NULL;
+	if (zstr(file_path)) {
+		return SWITCH_STATUS_FALSE;
+	}
+	if (! pool) {
+		if ((status = switch_core_new_memory_pool(&pool)) != SWITCH_STATUS_SUCCESS) {
+			return status;
+		}
+	}
```
mochouinard: You never free this pool ! If you created it, you must delete it before the end of the function. So I would define new_pool, and pool = new_pool, and at the end if (new_pool) delete new_pool.  You will have to change it to use goto instead of return as well after these lines

src/switch_core_file.c
```diff
@@ -62,6 +62,78 @@ static switch_status_t get_file_size(switch_file_handle_t *fh, const char **stri
...
+	if (zstr(file_path)) {
+		switch_log_printf(SWITCH_CHANNEL_LOG, SWITCH_LOG_ERROR, "Invalid Filename\n");
+		return SWITCH_STATUS_FALSE;
+	}
+	if (! pool) {
+		if (switch_core_new_memory_pool(&pool) != SWITCH_STATUS_SUCCESS) {
+			switch_log_printf(SWITCH_CHANNEL_LOG, SWITCH_LOG_ERROR, "Error creating \n");
+			return SWITCH_STATUS_FALSE;
+		}
+	}
```
mochouinard: Same problem here as previous one

tests/unit/switch_core_file.c
```diff
@@ -83,6 +83,7 @@ FST_CORE_BEGIN("./conf")
 			fst_check(status == SWITCH_STATUS_SUCCESS);
 			fst_check(!switch_test_flag(&fh, SWITCH_FILE_PRE_CLOSED));
 
+
```
mochouinard: Superfluous empty line

src/mod/applications/mod_http_cache/mod_http_cache.c
ovadbar: @Marc O Chouinard Thanks for your quick feedback. I know that I am duplicating code and that it would be a good idea to put them into a common function, but I was not sure if I should use the common function in the existing functions http_put and http_get because I am very hesitant to change existing code that I also didn't write. Additionally I am not sure whether that function should be inside the http_cache module or whether it it should be generic to be used with mod_httapi. 
Lastly I am not sure the preferred way to make a function like that. Would it be better to pass in a reference/pointer and return a switch_status or would it be better to have it return a CURL struct? 
I'll try and make updates shortly

mochouinard: I would do it for everything that is identical, like the CURLOPT_USERAGENT for example.  And all the one that are not common, leave them separate for now unless you have a good curl option knowledge.  It would at lest reduce current duplicate code, make sure functionality remain the same, and provide a framework for the next person who come along to change it who have more knowledge about curl to consolidate further.  For now, this is stuff hard coded for mod_http_cache, so it fine to leave it internal to that module.
I'm not sure what you mean with the result status.  Currently these return nothing, so you wouldn't have to deal with it, or just return Success by default for now.  What I would do is send a pointer of the curl_handle and maybe also the cache struct if it was used in all the functions and any other value that is standards and commonly needed, and the common function would access that pointer directly.

ovadbar: Thanks that hits the nail on the head of what I was asking. I made some updates according to your suggestions.

src/switch_core_file.c
```diff
@@ -62,6 +62,88 @@ static switch_status_t get_file_size(switch_file_handle_t *fh, const char **stri
...
+	switch_core_destroy_memory_pool(&new_pool);
+	return status;
+}
+
+SWITCH_DECLARE(switch_status_t) switch_core_file_remove( const char *file_path, switch_memory_pool_t *pool)
+{
+	char stream_name[128] = "";
+	char *rhs = NULL;
+	char *ext;
+	switch_memory_pool_t *new_pool;
```
mochouinard:
```suggestion
	switch_memory_pool_t *new_pool = NULL;
```

src/switch_core_file.c
```diff
@@ -62,6 +62,88 @@ static switch_status_t get_file_size(switch_file_handle_t *fh, const char **stri
...
 
+SWITCH_DECLARE(switch_status_t) switch_core_file_exists(const char *file_path, switch_memory_pool_t *pool)
+{
+	switch_status_t status;
+	char *rhs = NULL;
+	switch_memory_pool_t *new_pool;
+	if (zstr(file_path)) {
+		return SWITCH_STATUS_FALSE;
+	}
+	if (! pool) {
```
mochouinard:
```suggestion
	if (!pool) {
```

src/switch_core_file.c
```diff
@@ -62,6 +62,88 @@ static switch_status_t get_file_size(switch_file_handle_t *fh, const char **stri
 	return status;
 }
 
+SWITCH_DECLARE(switch_status_t) switch_core_file_exists(const char *file_path, switch_memory_pool_t *pool)
+{
+	switch_status_t status;
+	char *rhs = NULL;
+	switch_memory_pool_t *new_pool;
```
mochouinard:
```suggestion
	switch_memory_pool_t *new_pool = NULL;
```

src/switch_core_file.c
```diff
@@ -62,6 +62,88 @@ static switch_status_t get_file_size(switch_file_handle_t *fh, const char **stri
...
+		else {
+			if (file_interface->file_exists) {
+				status = file_interface->file_exists(file_path, pool);
+			}
+		}
+	}
+	else {
+		status = switch_file_exists(file_path, pool);
+	}
+	switch_core_destroy_memory_pool(&new_pool);
```
mochouinard: You need to check if (new_pool).  destroy_memory_pool does this on it first line : switch_assert(pool != NULL);

src/switch_core_file.c
```diff
@@ -62,6 +62,88 @@ static switch_status_t get_file_size(switch_file_handle_t *fh, const char **stri
...
+			if (file_interface->file_delete) {
+				status = file_interface->file_delete(file_path, pool);
+			}
+		}
+	} else if (switch_file_exists(file_path, pool) == SWITCH_STATUS_SUCCESS) {
+		status = switch_file_remove(file_path,pool);
+	} else {
+		status = SWITCH_STATUS_SUCCESS;
+	}
+	switch_core_destroy_memory_pool(&new_pool);
```
mochouinard: You need to check if (new_pool).  destroy_memory_pool does this on it first line : switch_assert(pool != NULL);

src/switch_core_file.c
```diff
@@ -62,6 +62,88 @@ static switch_status_t get_file_size(switch_file_handle_t *fh, const char **stri
...
+	char stream_name[128] = "";
+	char *rhs = NULL;
+	char *ext;
+	switch_memory_pool_t *new_pool;
+	switch_status_t status = SWITCH_STATUS_FALSE;
+	if (zstr(file_path)) {
+		switch_log_printf(SWITCH_CHANNEL_LOG, SWITCH_LOG_ERROR, "Invalid Filename\n");
+		return SWITCH_STATUS_FALSE;
+	}
+	if (! pool) {
```
mochouinard:
```suggestion
	if (!pool) {
```

src/mod/applications/mod_http_cache/mod_http_cache.c
```diff
@@ -1185,6 +1331,175 @@ static switch_status_t http_get(url_cache_t *cache, http_profile_t *profile, cac
...
+	switch_curl_easy_setopt(curl_handle, CURLOPT_MAXREDIRS, 10);
+	switch_curl_easy_setopt(curl_handle, CURLOPT_FAILONERROR, 1);
+	switch_curl_easy_setopt(curl_handle, CURLOPT_NOSIGNAL, 1);
+	if (headers) {
+		switch_curl_easy_setopt(curl_handle, CURLOPT_HTTPHEADER, headers);
+	}
+	switch_curl_easy_setopt(curl_handle, CURLOPT_URL, curl_data.url->url);
+	switch_curl_easy_setopt(curl_handle, CURLOPT_WRITEFUNCTION, get_file_callback);
+	switch_curl_easy_setopt(curl_handle, CURLOPT_WRITEDATA, (void *) &curl_data);
+	switch_curl_easy_setopt(curl_handle, CURLOPT_HEADERFUNCTION, get_header_callback);
```
mochouinard: These should not be done inside this common function I think.  And it might not be limited to these...
The idea is to look at everywhere these switch_curl_easy_setopt are used in this file (which you created new one), then look at the similar options being set with all of them, and create that http_curl_setopts() function with those in them.

In this commit, you have only used that common function once.

What I would do is do a search for CURLOPT_ in the file, and look where they are use and find the one you can group together because they are similar and put them in that function.  It fine that there are easy_setopt remaining in the functions, there should just be a lot less of them, which will make clearer that those are needed for that particular function.

ovadbar: So I think I had a bad merge/cherry-pick from working in a wrong branch that resulted in http_delete and http_exists being duplicated. I did use the http_curl_setopts in those functions in addition to http_get.
I looked at http_put and it was not similar enough to use http_curl_setopts for the following reasons.
It is making multiple requests inside a while loop. 
The curl handle is a CURL type not a  switch_CURL.
The options are not similar enough.

Regarding CURLOPT_WRITEFUNCTION & CURLOPT_WRITEDATA. I was not sure whether it should be included in the http_exists and http_delete requests because I do not care about the responses, however it looks like if I didn't use those than libcurl would use fwrite and I would need to set CURLOPT_WRITEDATA to a void file pointer but I figure if anyone after me wanted to expand the function to log the response on an error doing it this way should make that easier.

mochouinard: Because when I looked at the code where it was used it was different at some place, I would keep outside of the common function for the moment, and someone could add them as parameters if needed in the future.  I'll say stick to all the parameters that are really assigned exactly the same way for your initial change.  We can move other at a latter date if needed but the frame work for it will be there.

ovadbar: So I updated the code and added an empty_callback so it won't print to standard out for the http_delete and http_exist functions. Also I guess I forgot to reinstall when testing because I had a couple other errors which was causing crashes.

If you would like me to rebase this I can.

mochouinard: You forgot : http_put as well

ovadbar:
I was hesitant to modify http_put because there were a couple differences. I have now made those updates. Additionally I have removed setting the header write function in places where it is not needed.

PR #171, opened 2019-12-08
[mod_voicemail] Consolidating logic to find voicemail storage location.

mochouinard:
I had issues a while ago when path resolved correctly in some function, but not in others.  So Instead of trying to fix all the different implementation of the same things, I've abstracted it to it own function.


src/mod/applications/mod_voicemail/mod_voicemail.c
```diff
@@ -829,11 +886,13 @@ static vm_profile_t *get_profile(const char *profile_name)
 
 static void profile_rwunlock(vm_profile_t *profile)
 {
-	switch_thread_rwlock_unlock(profile->rwlock);
-	if (switch_test_flag(profile, PFLAG_DESTROY)) {
-		if (switch_thread_rwlock_trywrlock(profile->rwlock) == SWITCH_STATUS_SUCCESS) {
-			switch_thread_rwlock_unlock(profile->rwlock);
-			free_profile(profile);
+	if (profile) {
```
mjerris: lets simplify changes to:

if (!profile) return;

mochouinard: No problem ! Done !

PR #202, opened 2020-01-03
[mod_opusfile] fix play sample rate (don't use original rate, always use 48k), fix eof.

dragos-oancea:


crienzo:
This fix should include a test w/ the problem file.

davehorton:
thank you, I can confirm it plays correctly and I do get a play done event now.

However, there is a new problem.  The last half second or so of the audio file is clipped.

Here is a [debug logfile](https://gist.github.com/davehorton/2097eb0b45fbd8f6ba43ed4ca4b3be02).

I'm guessing the issue is that when you return EOF to indicate the file has been completely decoded there are still bytes in the switch buffer that have not gotten played out, and those get discarded.

andywolk:
Reverted

PR #206, opened 2020-01-07
[mod_opusfile] fix play sample rate (don't use original rate, always use 48k), fix EOF, add unit-test.

dragos-oancea:


src/mod/formats/mod_opusfile/test/freeswitch.xml.fsxml
```diff
@@ -0,0 +1,21 @@
+<document type="freeswitch/xml">
+
+  <section name="configuration" description="Various Configuration">
+    <configuration name="modules.conf" description="Modules">
+      <modules>
+        <load module="mod_loopback"/>
+        <load module="mod_opusfile"/>
```
crienzo: This file should not be included

PR #214, opened 2020-01-09
Fix memory leak in case of absent session

russagit:


lazedo:
@russagit this and other potential memory leaks should be fixed in #239

PR #219, opened 2020-01-10
Update avformat.c

mangshe0:
fix mod_av not have auido when use avresample.
the define swr_convert is error.
----it cost my long time to find it.


seven1240:
Looks fine to me, can you rebase the pr to the latest master and update the comment to `[mod_av] ...` ?

PR #220, opened 2020-01-10
[mod_av] Fix mod_av not having audio when using avresample

mangshe0:
[mod_av] fix mod_av not have auido when use avresample.


crienzo:
@mjerris - this is the fix up of the previous PR that @seven1240 reviewed.  Looks fine to us.

Issue #201, opened 2020-01-02
[ERR] sofia.c:5061 Invalid ext-rtp-ip (crash)


misuzu:
FreeSwitch 1.10.2 fails to start without internet connection:
```
freeswitch # [   20.792659] freeswitch[619]: 2020-01-02 12:03:08.839465 [NOTICE] switch_loadable_module.c:412 Adding API Function 'console'
freeswitch # [   20.798427] freeswitch[619]: 2020-01-02 12:03:08.844527 [CONSOLE] switch_loadable_module.c:1804 Successfully Loaded [mod_logfile]
freeswitch # [   20.800605] freeswitch[619]: 2020-01-02 12:03:08.895528 [CONSOLE] switch_loadable_module.c:1804 Successfully Loaded [mod_enum]
freeswitch # [   20.808663] freeswitch[619]: 2020-01-02 12:03:08.895601 [NOTICE] switch_loadable_module.c:292 Adding Dialplan 'enum'
freeswitch # [   20.810266] freeswitch[619]: 2020-01-02 12:03:08.895711 [NOTICE] switch_loadable_module.c:350 Adding Application 'enum'
freeswitch # [   20.816485] freeswitch[619]: 2020-01-02 12:03:08.895790 [NOTICE] switch_loadable_module.c:412 Adding API Function 'enum'
freeswitch # [   20.818089] freeswitch[619]: 2020-01-02 12:03:08.895860 [NOTICE] switch_loadable_module.c:412 Adding API Function 'enum_auto'
freeswitch # [   20.819742] freeswitch[619]: 2020-01-02 12:03:08.901481 [CONSOLE] switch_loadable_module.c:1804 Successfully Loaded [mod_cdr_csv]
freeswitch # [   20.826483] freeswitch[619]: 2020-01-02 12:03:08.901538 [NOTICE] switch_loadable_module.c:412 Adding API Function 'cdr_csv'
freeswitch # [   20.828111] freeswitch[619]: 2020-01-02 12:03:08.912054 [CONSOLE] switch_loadable_module.c:1804 Successfully Loaded [mod_event_socket]
freeswitch # [   20.829802] freeswitch[619]: 2020-01-02 12:03:08.912121 [NOTICE] switch_loadable_module.c:350 Adding Application 'socket'
freeswitch # [   20.836390] freeswitch[619]: 2020-01-02 12:03:08.912207 [NOTICE] switch_loadable_module.c:412 Adding API Function 'event_sink'
[...]

PR #221, opened 2020-01-10
[mod_sofia] Add abort-on-empty-external-ip global param disabled by default.

andywolk:


andywolk:
This closes https://github.com/signalwire/freeswitch/issues/201

PR #225, opened 2020-01-13
modify FreeSwitchCore.2017.vcxproj

mangshe0:
vs2017 build switch_curl.c need headfile, modify vcxproj append curl lib headfile.


andywolk:
There is no need for that since curl.props is used:
https://github.com/signalwire/freeswitch/blob/master/w32/Library/FreeSwitchCore.2017.vcxproj#L52

PR #226, opened 2020-01-13
Broken windows

grmt:
make FS build and run on (at least) windows 10
used the latest updated Visual Studio 2017, so you want to update your VS installation
not sure what the agreed vcxproj file format is, I "dos2unix"-ed the ones I changed, so no BOF and only <LF>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,7 +59,6 @@ TAGS
 
 /w32/Library/lastversion
 /w32/Library/tmpVersion.Bat
-!/w32/Console/FreeSwitchConsole.vcproj.user
 !/w32/Setup/inno_setup/vcredist_x64.exe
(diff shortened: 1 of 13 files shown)


andywolk:
Can't reproduce

andywolk:
That's a breaking PR.
Please try installing Windows 10 SDK v10.0.17134 and see if that helps.
We would also like to not stay with VS 2017 but move to Visual Studio 2019, that's on my plate already.

grmt:
Did you try updating VS207 and building and running and **interactively** look at error messages, recently? If you don't run it, and don't compile mod_http_cache you might not have issues.

grmt:
I actually ported everything to  VS2019

PR #253, opened 2020-01-21
[mod_sofia] multi-server call recovery

21stcaveman:
This will enable a freeswitch instance, to recover calls tracked by a different instance, without the need for floating IP setup mentioned in the documentation.


ChuanjiaCai:
hi guys,
i use two freeswitch A and B, two keepalived on FS A and FS B, connect one mysql for HA,
i got one issue,about fsctl recover, my step as blow:
1,Two phones are registered to the float IP,phone a call phone b，when i sent command to main FS A, fsctl crash,
2,the other FS B, send fsctl recover, should recover the calls, (calls channels need transfer to FA B from FA A)
3,but the calls channels cannot transfer to FA B, cannot recover the calls,

can you give me some advice for HA, thanks

PR #260, opened 2020-01-24
[mod_kazoo] Fix performance issue

russagit:
We have found out that after commit https://github.com/signalwire/freeswitch/commit/df1ac5dbe0 (since FS v1.8.6)
the performance of fetch replies drastically decreased and it happened beacause of using
switch_queue_pop_timeout instead of switch_queue_trypop which was used before that commit
When we rolled back to switch_queue_trypop the perfomance was as good as in FS v1.8.5


andywolk:
@lazedo

lazedo:
@russagit can you try #261 ? i think it serves both purposes while keeping ` switch_queue_trypop` as default

russagit:
> @russagit can you try #261 ? i think it serves both purposes while keeping ` switch_queue_trypop` as default

Thanks, #261 works great

russagit:
Closing this in favor of https://github.com/signalwire/freeswitch/pull/261

PR #266, opened 2020-01-26
[mod_conference] fix overflow (video mode == passthrough)

dragos-oancea:
It looks like allocating it on the Heap  makes Asan give more info: "READ of size 16424 at 0x629000050200 thread T50"


src/mod/applications/mod_conference/conference_video.c
```diff
@@ -4834,13 +4834,15 @@ void conference_video_write_frame(conference_obj_t *conference, conference_membe
 {
 	conference_member_t *imember;
 	int want_refresh = 0;
-	unsigned char buf[SWITCH_RTP_MAX_BUF_LEN] = "";
```
mjerris: can we keep this on stack but make it the right size instead.  This gets called a LOT.

PR #268, opened 2020-01-27
[core] add audio channels limit when opening files, configurable through switch.conf.xml (param "max-audio-channels")

dragos-oancea:


src/switch_core_file.c
```diff
@@ -361,6 +361,18 @@ SWITCH_DECLARE(switch_status_t) switch_core_perform_file_open(const char *file,
 		goto fail;
 	}
 
+	if (fh->channels > 2) {
```
mjerris: why warn here?

src/switch_core_file.c
```diff
@@ -361,6 +361,18 @@ SWITCH_DECLARE(switch_status_t) switch_core_perform_file_open(const char *file,
...
+		/* just show a warning for more than 2 channels, no matter if we allow them or not */
+		switch_log_printf(SWITCH_CHANNEL_LOG, SWITCH_LOG_WARNING, "File [%s] has more than 2 channels: [%u]\n", file_path, fh->channels);
+	}
+
+	if (switch_core_max_audio_channels(0) && fh->channels > switch_core_max_audio_channels(0)) {
```
mjerris: can we run this once and put in a var

src/switch_core_file.c
mjerris: this was for logging to see whats actually happening in the wild.  thats fine

PR #274, opened 2020-01-29
[core] scan-build: Value stored to 'file' is never read

dragos-oancea:


src/switch_ivr_originate.c
```diff
@@ -231,15 +231,10 @@ static void *SWITCH_THREAD_FUNC collect_thread_run(switch_thread_t *thread, void
 
 	while (switch_channel_ready(channel)) {
 		switch_size_t len = strlen(collect->key);
-		const char *file = collect->file;
```
mjerris: i think the right solution on this is to pass file instead of collect->file to switch_ivr_read so that we play silence from switch_ivr_read

PR #277, opened 2020-01-30
[core] scan-build: Null pointer passed as an argument to a 'nonnull' parameter - switch_event_base_add_header()

dragos-oancea:


src/switch_event.c
```diff
@@ -1113,7 +1113,7 @@ static switch_status_t switch_event_base_add_header(switch_event_t *event, switc
 
 	redraw:
 		len = 0;
-		for(j = 0; j < header->idx; j++) {
+		for(j = 0; j < header->idx && header->array[j]; j++) {
```
mjerris: is it possible for this to ever be NULL?  maybe we want to just if (!header->array[j]) continue; inside the loop instead

PR #278, opened 2020-01-30
[core] scan-build: Access to field 'microseconds_per_packet' results in a dereference of a null pointer

dragos-oancea:


src/switch_loadable_module.c
```diff
@@ -2677,7 +2677,7 @@ static void switch_loadable_module_sort_codecs(const switch_codec_implementation
 	switch_log_printf(SWITCH_CHANNEL_LOG, SWITCH_LOG_ERROR, "--BEFORE\n");
 #endif
 
-	for (i = 0; i < arraylen; i++) {
+	for (i = 0; i < arraylen && array[i]; i++) {
```
mjerris: we probably want to do if (!array[i]) continue here.

PR #283, opened 2020-01-30
[core] scan-build: Value stored to 'frame' during its initialization is never read - eavesdrop_callback()

dragos-oancea:


src/switch_ivr_async.c
```diff
@@ -2001,7 +2001,7 @@ static switch_bool_t eavesdrop_callback(switch_media_bug_t *bug, void *user_data
 	}
 
 	if (nframe) {
-		switch_frame_t frame = {0};
+		switch_frame_t frame;
 		uint8_t buf[SWITCH_RECOMMENDED_BUFFER_SIZE] = "";
 		
 		frame = *nframe;
```
mjerris: just move the = *nframe up to the decl

PR #290, opened 2020-01-31
allow mariadb URLs for odbc-dsn for mod_db

nneul:


nneul:
@andywolk 

Not really sure why this module in particular has this dsn validity check when none of the others do - probably worth removing that entirely so it'll handle other potentially valid dsn forms. This change at least just gets it to support the mod_mariadb changes.

andywolk:
@nneul I get what you say, but what do you think about this: https://lists.freeswitch.org/pipermail/freeswitch-users/2019-November/132682.html
Pgsql was there before we implemented the database interface. Mariadb is not hardcoded since it comes via that database interface (and does have multiple prefixes). The thing is there may be something else missing in that regex and not just mariadb.
So I'm open for ideas.

nneul:
That's a nice little hack that'll work to avoid needing the merge. I still would recommend removing this entirely -- or (would require more development) - move it to a core function so that it can be applied to _all_ of the dsn using modules and check against validity based on currently loaded modules.

nneul:
In lieu of that, what about just flattening to `^[a-z]+://|`to allow stuff that looks like a url.

andywolk:
> or (would require more development) - move it to a core function so that it can be applied to _all_ of the dsn using modules and check against validity based on currently loaded modules.

Exactly. We can keep merging little things or do it right once.

lomelee:
please add this change in mater version, thanks

PR #310, opened 2020-02-04
mod_loopback: resolve missing loopback_initial_codec when calling from originate cmd with no session

wmasilva:
when executing from the fs_cli:

originate {loopback_initial_codec=L16@16000h}loopback/123 &playback(/home/Track4.wav)

The loopback channel A is created with always use the default codec L16@8000h and don't respect the variable because the variable is not set in the new session that is created. 

This commit fix it.

### DIFF
--- a/src/mod/xml_int/mod_xml_curl/conf/autoload_configs/xml_curl.conf.xml
+++ b/src/mod/xml_int/mod_xml_curl/conf/autoload_configs/xml_curl.conf.xml
@@ -44,6 +44,10 @@
 
       <!-- one or more of these imply you want to pick the exact variables that are transmitted -->
       <!--<param name="enable-post-var" value="Unique-ID"/>-->
+
+      <!-- optional: maximum response size for this binding in bytes. 
+           Defaults to XML_CURL_MAX_BYTES (1MB) if omitted -->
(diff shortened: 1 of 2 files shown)


dragos-oancea (dismissed):

src/mod/endpoints/mod_loopback/mod_loopback.c
```diff
@@ -1170,6 +1170,9 @@ static switch_call_cause_t channel_outgoing_channel(switch_core_session_t *sessi
 				switch_channel_set_variable(channel, "loopback_leg", "A");
 				switch_channel_set_variable(channel, "is_loopback", "1");
 			}
+
+			switch_channel_set_variable(channel, "loopback_initial_codec", switch_event_get_header(var_event, "loopback_initial_codec"));
```
crienzo, line 1174: I think this should old be set if `loopback_initial_codec` is passed into `var_event`.

wmasilva: ok, i just change to check if its set.

wmasilva:
i needed to do a force push to avoid the commits from master... 
Is there a way to re-open this pr or i create a new one? 

Thanks,

PR #312, opened 2020-02-05
[sofia-sip] Don't try to recover WS and WSS.

andywolk:


andywolk:
Moved to https://github.com/freeswitch/sofia-sip/pull/43

PR #342, opened 2020-02-12
[core] scan-build: Value stored to 'last_speed' is never read - switch_ivr_play_file()

dragos-oancea:


src/switch_ivr_play_say.c
```diff
@@ -1817,7 +1817,6 @@ SWITCH_DECLARE(switch_status_t) switch_ivr_play_file(switch_core_session_t *sess
 					switch_size_t r = newlen - wrote;
 					switch_buffer_write(fh->sp_audio_buffer, bp, r * 2);
 				}
-				last_speed = fh->speed;
```
mjerris: seems like removing this basically gets rid of the entire point of last_speed (and also the init of it at the top of the for loop.  It looks to me that the right fix for this might be to not init last speed at the top of the for loop, and instead is to move that init outside the for loop.  This needs testing to confirm that this is intended behavior.

PR #343, opened 2020-02-12
[core] scan-build: Value stored to 'midstate' during its initialization is never read - switch_core_session_hangup_state()

dragos-oancea:


src/switch_core_state_machine.c
```diff
@@ -803,7 +803,7 @@ SWITCH_DECLARE(void) switch_core_session_hangup_state(switch_core_session_t *ses
 	int do_extra_handlers = 1;
 	int silly = 0;
 	int index = 0;
-	switch_channel_state_t state = switch_channel_get_state(session->channel), midstate = state;
+	switch_channel_state_t state = switch_channel_get_state(session->channel), midstate;
```
mjerris: why do we have ,midstate here?

dragos-oancea: it is used in STATE_MACRO

PR #347, opened 2020-02-12
[core] scan-build: Value stored to 'status' is never read - switch_ivr_eavesdrop_session()

dragos-oancea:


dragos-oancea:
> should we be returning STATUS_BREAK if doing require group and no match?

it goes to "end" to  rwunlock session . (we are in a "switch_core_session_locate(uuid)" condition

mjerris:
> > should we be returning STATUS_BREAK if doing require group and no match?
> 
> it goes to "end" to rwunlock session . (we are in a "switch_core_session_locate(uuid)" condition

I mean should we actually be returning BREAK from function in these cases?  I'm not sure... needs to be dug in to see

PR #377, opened 2020-02-16
[mod_hiredis] Fix dependency for Debian Buster.

alexey-khabulyak:
debian buster contains libhiredis0.14.
for now u have to install libhiredis0.13 or 0.10 from old repo to install mod-hiredis on debian 10


andywolk:
Thank you very much for the patch.

PR #380, opened 2020-02-17
[mod_httapi] Introduction of connect-timeout param

saurabh-iitrke:
Added a new param for mod_httapi, which will limit the time taken to establish connection to the server.
Current default timeout (i.e. 300 seconds) is too high.


saurabh-plivo:
> can we either document what 0 means, or simply remove the = from >= ?

Thanks for pointing it out. Documented.

PR #383, opened 2020-02-18
Add Charge SIP Header Support

daniel-lucio:
I have found that some proprietary PBX's (such as Broadsoft) send the Charge header as identification instead of using Diversion or X-Accountcode.


mjerris:
what is the format of this header?  should we be parsing it at all ?

daniel-lucio:
> what is the format of this header? should we be parsing it at all?

Charge: <tel:+199999999999>;noa=clgp-ani-natl-num

I am thinking, once it is captured, anyone could parse using regex right?

PR #391, opened 2020-02-19
[sofia-sip] scan-build: Fix "Result of operation is garbage or undefined" in sres_log_response()

andywolk:


andywolk:
@placebokkk Will be addressed by https://github.com/signalwire/freeswitch/pull/394

PR #432, opened 2020-02-25
[mod_rtmp] scan-build: Value stored to 'status' is never read - RTMP_INVOKE_FUNCTION(rtmp_i_sendevent)

dragos-oancea:


dragos-oancea:
this fixes an use-after-free too.

PR #445, opened 2020-02-26
[sofia-sip] Add urn: scheme support to sofia.

cauca-signalwire:
As an effort to support NENA i3 specs for NG 9-1-1, this adds support
for INVITE URI appearing as urn:service:sos to request for emergency services.


cauca-signalwire:
> date > libs/sofia-sip/.update

For good mesure, I rebased and used `LC_ALL=en_US LANG=en_US.utf8 date > libs/sofia-sip/.update`

PR #446, opened 2020-02-26
Add mod_pulseaudio

cauca-signalwire:
Adds a new mod_pulseaudio module to use on desktops.


cauca-signalwire:
I found a lot of cleanup to do while implementing multiple endpoints.  You might want to hold on the code review if it is not started.

cauca-signalwire:
This will make the module much more cleaner, it still works using pulseaudio/auto_answer however, I just found out about how endpoints (pulseaudio/endpoint/<name>) works and they never worked in pulseaudio, I'll be fixing those since I need them too. Incoming commits soon.

cauca-signalwire:
This is now feature complete for us and ready for review.

A couple questions;
- Does the file header needs be updated?
- Should the functions be localized to not clash with other modules in the future?
- Would it be better to remove pablio.c which was used to make portaudio synchronous? (pa_simple is synchronous)
- Do I need to check for memory leaks before submitting for review?

configure.ac
```diff
@@ -1324,6 +1324,16 @@ PKG_CHECK_MODULES([PORTAUDIO], [portaudio-2.0 >= 19],[
   AC_MSG_RESULT([no]); AM_CONDITIONAL([HAVE_PORTAUDIO],[false])])
 PKG_CHECK_MODULES([LDNS], [libldns-fs >= 1.6.6],[
   AM_CONDITIONAL([HAVE_LDNS],[true])],[
+PKG_CHECK_MODULES([PULSEAUDIO], [libpulse-simple >= 0.1],[
```
mjerris, line 1327: this is stuffed inside the HAVE_LDNS conditional.

crienzo:
@cauca-signalwire we can host this out of tree in github.com/freeswitch/mod_pulseaudio and add a link to it in modules.conf

cauca-signalwire:
@crienzo is it quite an easy process to independently build this as an external module, I don't think I ever tried building out-of-tree modules with FreeSWITCH.
I'd also like to add that mod_pulseaudio is much more robust than mod_portaudio on Linux, the code looks more clean to me and has less quirks to get working because all the output selection and volumes are handled by Pulse itself.  It could bring other softphones or bridges that build on top of FreeSWITCH too.
Just giving some thoughts on this.

nicsalv:
Tried out this mod on FreeSWITCH 1.10.5 on Debian 10.6, everything works as expected.
It would be nice having this mod available on the official repo.

src/mod/endpoints/mod_pulseaudio/pablio.c
```diff
@@ -0,0 +1,146 @@
...
+
+	/* Allocate PABLIO_Stream structure for caller. */
+	aStream = (PABLIO_Stream *) malloc(sizeof(PABLIO_Stream));
+	switch_assert(aStream);
+	memset(aStream, 0, sizeof(PABLIO_Stream));
+
+	/* Open a PulseAudio stream that will be used to communicate with the underlying
+	 * audio drivers. */
+	bzero(&buffer_attr, sizeof(buffer_attr));
+	buffer_attr.prebuf = (uint32_t) -1;
```
nicsalv, line 86:
```suggestion
	buffer_attr.prebuf = (uint32_t) -1;
	buffer_attr.minreq = (uint32_t) -1;
	buffer_attr.maxlength = (uint32_t) -1;
```
According to PulseAudio guidelines about latency control [https://www.freedesktop.org/wiki/Software/PulseAudio/Documentation/Developer/Clients/LatencyControl/](url), these fields should be initialised at `(uint32_t) -1`.

src/mod/endpoints/mod_pulseaudio/pablio.c
```diff
@@ -0,0 +1,146 @@
...
+
+	if (inputParameters) {
+		buffer_attr.fragsize = pa_usec_to_bytes(latency_msec * PA_USEC_PER_MSEC, inputParameters);
+		buffer_attr.maxlength = buffer_attr.fragsize;
```
nicsalv, line 90:
```suggestion
```

src/mod/endpoints/mod_pulseaudio/pablio.c
```diff
@@ -0,0 +1,146 @@
...
+		aStream->istream = pa_simple_new(NULL, appName, PA_STREAM_RECORD, NULL, channelName, inputParameters, NULL, &buffer_attr, &err);
+		if (!aStream->istream) {
+			goto error;
+		}
+	}
+
+	if (outputParameters) {
+		channels = outputParameters->channels;
+		buffer_attr.tlength = pa_usec_to_bytes(latency_msec * PA_USEC_PER_MSEC, outputParameters);
+		buffer_attr.maxlength = buffer_attr.tlength;
```
nicsalv, line 101:
```suggestion
```

src/mod/endpoints/mod_pulseaudio/pablio.c
```diff
@@ -0,0 +1,142 @@
...
+#include <stdlib.h>
+#include <string.h>
+#include <pulse/timeval.h>
+#include <pulse/simple.h>
+#include "pablio.h"
+
+long WriteAudioStream(PABLIO_Stream * aStream, void *data, size_t datalen, switch_timer_t *timer)
+{
+	pa_simple_write(aStream->ostream, data, datalen, NULL);
+	switch_core_timer_next(timer);
```
nicsalv, line 41:
```suggestion
```
Why use a timer? This function gets called inside the `channel_write_frame` callback, that is supposed synchronized.

src/mod/endpoints/mod_pulseaudio/pablio.c
```diff
@@ -0,0 +1,142 @@
...
+
+	return datalen;
+}
+
+long ReadAudioStream(PABLIO_Stream * aStream, void *data, size_t datalen, switch_timer_t *timer)
+{
+	pa_simple_read(aStream->istream, data, datalen, NULL);
+	switch_core_timer_next(timer);
```
nicsalv, line 49:
```suggestion
```
Why use a timer? This function gets called inside the `channel_read_frame` callback, that is supposed synchronized.

src/mod/endpoints/mod_pulseaudio/pablio.c
ticpu: Anything in pablio.h/c is code that was imported from PortAudio, so it might just be remains from the ring buffer that was there.

I didn't take time to understand this part, I'll give it a try without the timer, thanks for the suggestion.

I'm also thinking about removing or restructuring/renaming pablio.h/c to more accurately represent PulseAudio's API.

nicsalv: Thank you for replying.

I tried a call without the timers and everything seems to work. 

Another issue I'm facing regards latency. I'm using this mod on an embedded device with low computational resources. When the `module-echo-cancel` is enabled the latency grows heavily (several seconds).
Using the `module-loopback` together with `module-echo-cancel` does not present this issue: latency still increases but only of several tenths of a second. So I guess the Simple API is not suited for low-latency applications.
Would you have any suggestion about this? 

Thank you

ticpu: When I decided about using libpulse-simple library instead of the fully fledged libpulse here is what I took into consideration.

I took mod_portaudio as a base for building upon since it was more recent than mod_alsa, in mod_portaudio, pablio.c's main job was to build a ring buffer inside freeswitch and to make calls to ReadAudioStream/WriteAudioStream synchronous. This is also why the pablio files are so lightweight now.

The main goal of using libpulse is to be able to attach pulsaudio's library to an existing main loop and make those calls asynchronous.  The other advantage of using libpulse is to have control over audio volumes and other internals to better handle pulseaudio specific events.

The pulse-simple library is a wrapper around libpulse that automatically creates a thread and a mainloop for pulseaudio and handle all the locking and callbacks to make sure that read/write calls are synchronous.

Now, the part I'm not sure is whatever Freeswitch can work with asynchronous read/write for audio channels. From a design perspective it'd make sense to remove the pulseaudio mainloop thread as this would be less overhead, and it seems like quite a big rewrite.

One other detail, I just noticed latency_msec is hard-coded to 40, I was sure it was configurable, you could try making this a configuration and tweaking it for your application, it might help.

nicsalv: I totally agree with you, it'd make sense to remove PulseAudio's main loop, but as you said it seems quite complicated. 

What do you mean by asynchronous read/write with Freeswitch? As far as I know, audio is read/written with the IO routines `channel_read_frame` and `channel_write_frame`. Is this already an asynchronous behavior, isn't it? 

Anyway, after several tests, the threaded main loop of PulseAudio does not seem too heavy by itself, but when the echo cancellation is enabled server-side the overhead becomes too much for low computational resources, as my desktop can make calls without delay while the low-powered device cannot.

I wonder if the overall logic could be lighten in some way.

PR #453, opened 2020-02-27
[mod_verto] scan-build: Null pointer passed as an argument to a 'nonnull' parameter - SWIG_TypeProxyCheck

dragos-oancea:


dragos-oancea:
generated by SWIG

PR #460, opened 2020-02-28
VISYTEL-0003 [core] Change switch_ivr_wait_for_answer to support EARLY MEDIA and call_timeout prior to answer (when not called from a bridge function)

Visytel:
Changes are to allow the FreeSWITCH/Lua function: session:waitForAnswer(session) to actually wait for the call to be answered when the originate command includes (ignore_early_media=false) and the outbound call detects EARLY MEDIA.  The originate call_timeout setting will also work as expected and now the returned cause will be SWITCH_CAUSE_NO_ANSWER when call_timeout has been reached with the outbound call has not been answered.  Also for non EARLY MEDIA calls if the status == SWITCH_STATUS_TIMEOUT then instead of returning the cause SWITCH_CAUSE_DESTINATION_OUT_OF_ORDER the hangup cause will also be SWITCH_CAUSE_NO_ANSWER.  Should switch_ivr_wait_for_answer be called from a bridging function then the function will revert to the previous way it worked.  Lastly extra DEBUG logging has been added to enable clear debugging of these changes in production.


Visytel:
@mjerris @andywolk I thought that I would add what I added when responding to Michael Jerris on freeswitch-dev regarding this PR.

Essentially the current way session:waitForAnswer(session) works (prior to my PR) (when ignore_early_media=false is used inside the originate command) is that it returns immediately when early_media is detected and does not wait for answer (as the function name suggests).  The aim I have is to let originate start my Lua script when early media is detected (or ring ready) then call the session:waitForAnswer(session) function and actually wait for answer (thus allowing for the call_timeout variable to also work as expected).  This can only be achieved with this PR.  

I am more than happy to explain this further once this PR is being reviewed and provide clear examples of the problem if required (however I feel my code changes are well documented inside the PR itself).  Also this change only affects originate API calls and not bridging.

Visytel:
@anthmFS @mjerris @andywolk I have applied the changes suggested by Anthony (via the Slack freeswitch-dev channel.  Now there is a switch_yield(10000); added to avoid a tight loop in two parts of the switch_ivr_wait_for_answer() function.  Handling of session == NULL and using the peer_session has also been included.  Testing has been done successfully at my end.

Visytel:
@anthmFS @mjerris @andywolk Please see the latest commit (37c5e12) as there was one left over change that I have fixed.

Visytel:
@anthmFS @mjerris @andywolk I am closing this PR since I have found another way of achieving this without the need to change the codebase.  Thanks Anthony for your feedback.

Issue #467, opened 2020-03-03
STUN Binding Request verification truncates username & fails


LarryHemenway:
FreeSWITCH Version 1.10.2-release.4~64bit (-release.4 64bit)

When sending in a STUN username of greater than 34 bytes, the STUN Binding request is ignored because it truncates the username. 

The problem is in switch_rtp.c handle_ice() function: 

    static void handle_ice(switch_rtp_t *rtp_session, switch_rtp_ice_t *ice, void *data, switch_size_t len)
    {
        ....
        char username[34] = { 0 };
        unsigned char buf[512] = { 0 };

For example, my STUN Binding request is 

    dTg40bESPYEAj58h:HkOwvylni/1IHG4YWcO+tC1PBu6QjWV+

but it gets truncated to 

    dTg40bESPYEAj58h:HkOwvylni/1IHG4YW 

The function 'icecmp' compares the truncated value stored in username above to the full username stored in ice->user_ice and returns failure. The request is then ignored. 

I looked at a [RFC5389 STUN](https://tools.ietf.org/html/rfc5389) and found the following

>The value of USERNAME is a variable-length value.  It MUST contain a
>UTF-8 [RFC3629] encoded sequence of less than 513 bytes, and MUST
>have been processed using SASLprep [RFC4013].

The easiest solution would be to increase the username to conform to RFC5389 and make it 513 bytes. After doing that, the buf that contains the copied data would also need to increase. I have less analysis on this, but I think it makes sense to bump it up to a typical MTU of 1500. Combined, the updates would be

    static void handle_ice(switch_rtp_t *rtp_session, switch_rtp_ice_t *ice, void *data, switch_size_t len)
    {
        ....
        char username[513] = { 0 };   // conform with RFC5389
        unsigned char buf[1500] = { 0 }; // a typical MTU size

<h2> Other Potential Issue</h2>

Stun trace does not appear to work from the CLI, or at least not in the way I would expect:
sofia loglevel <all|default|tport|iptsec|nea|nta|nth_client|nth_server|nua|soa|sresolv|stun> [0-9]

When I type:

    freeswitch@88d481f5770a> sofia loglevel stun 5

I get the following:
    
    USAGE:
    --------------------------------------------------------------------------------
    sofia global siptrace <on|off>
    sofia        capture  <on|off>
                 watchdog <on|off>
    
    sofia profile <name> [start | stop | restart | rescan] [wait]
[...]

PR #468, opened 2020-03-03
[Core] switch_rtp: update username to allow size specified in STUN RFC 5389

LarryHemenway:
See [Issue 467: STUN Binding Request verification truncates username & fails](https://github.com/signalwire/freeswitch/issues/467)

### DIFF
--- a/src/switch_rtp.c
+++ b/src/switch_rtp.c
@@ -82,10 +82,10 @@ static const switch_payload_t INVALID_PT = 255;
 
 #define rtp_session_name(_rtp_session) _rtp_session->session ? switch_core_session_get_name(_rtp_session->session) : "-"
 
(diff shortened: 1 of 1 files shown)


andywolk:
Looking good, but let's format everything properly:
1) Please don't use `//` comments, use `/* */` instead
2) Commit messages should start from a [component], so this should look like: `[Core] Switch_rtp STUN username size update` (If you don't know how to squash multiple commits into one, just leave as is, we will squash for you when merging)

LarryHemenway:
@andywolk I updated the comments. Sorry - I don't know how to squash commits. Thanks for the input and quick review.

andywolk:
Thank you for the contribution!

PR #488, opened 2020-03-08
[core] minor logging fix for switch_rtp_set_max_missed_packets

Karn:
**Proposed Changes**:
Looks like the log to indicate that the max_missed_packets threshold is changing to something below the current missed packet threshold is incorrectly outputting the transition as `missed_count -> max` instead of `max_missed_count -> max`.


dragos-oancea:
fixed in 0877bd5770b8853dc81132c64b971bdb81b56d8c .

PR #509, opened 2020-03-15
[mod_xml_curl] Allow XML_CURL_MAX_BYTES to be configured at runtime 

talmakion:
[mod_xml_curl] Allow XML_CURL_MAX_BYTES to be configured at runtime per-binding. 

* Added "response-max-bytes" config param for mod_xml_curl
* XML_CURL_MAX_BYTES define still used as default if param omitted
* Added example configuration to module example config.


DusanBaric:
Any update on this pull request? It would be very useful if we could configure XML_CURL_MAX_BYTES with xml configuration at runtime as we started to hit hardcoded limit of 1 MB.

DusanBaric (approved):

crienzo (approved):

PR #514, opened 2020-03-17
[Verto Communicator] Update adapter-js to latest version

StefanYohansson:
Some browsers start to comply on RFC and stop receiving callbacks for some WebRTC functions. The adapter-js now shims those callbacks behavior to work as promise underneath the library.

### DIFF
--- a/html5/verto/js/src/vendor/adapter-latest.js
+++ b/html5/verto/js/src/vendor/adapter-latest.js
@@ -97,6 +97,7 @@ function adapterFactory() {
       chromeShim.shimOnTrack(window);
       chromeShim.shimAddTrackRemoveTrack(window);
       chromeShim.shimGetSendersWithDtmf(window);
+      chromeShim.shimGetStats(window);
       chromeShim.shimSenderReceiverGetStats(window);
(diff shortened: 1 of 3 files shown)


mjerris:
I'd like to see this code stripped out of freeswitch tree into its own repo.  Is this something you could assist with ?

StefanYohansson:
Yes, I can. Give me some time to remember how Verto Communicator was built lol

StefanYohansson:
@mjerris do you prefer from npm install and bundled with `grunt` (the current bundler tool but this way someone still needs to bump the version in order to upgrade) or just a `<script src` on index.html pointing to `https://webrtchacks.github.io/adapter/adapter-latest.js`? The second it will be always the latest version but the downside is that link can experience downtimes.

italorossi:
Is it possible to upgrade the adapter.js and include it in the bundle whenever we npm build it?

StefanYohansson:
Yes, it is what I am doing now. It will fetch from npm and include in the bundle so we don't need to update by hand

crienzo:
This code has moved to https://github.com/freeswitch/verto-client

PR #516, opened 2020-03-18
to destory a created event

nwaycn:
has create an event for SWITCH_EVENT_DETECTED_SPEECH,but not destory


seven1240:
forget it if it successfully queued, it should be covered in other place.

crienzo (approved):

crienzo:
@andywolk this looks OK- edge case, but it could happen.  Spacing in the commit needs to be fix along with the commit message.

crienzo:
Actually, @seven1240 is right- i misread the switch_core_session_queue_event() function.  This doesn't leak.

PR #545, opened 2020-04-04
[Configuration] sip_profiles: Use $${external_sip_ip} for ext-sip-ip, $${external_rtp_ip} for ext-rtp-ip

nwaycn:
<param name="ext-sip-ip" value="$${external_rtp_ip}"/>  in files internal.xml and external.xml,but it maybe is $${external_sip_ip}


mjerris:
not following what you are trying to say.  ext-sip-ip is a profile param and external_sip_ip is a variable.  the configs look accurate to me

PR #547, opened 2020-04-07
add final chat function to prevent sms from further process

windy-wang:
In practice, it is found there is not flag to show the final sms that the caller would do unexpected further process. This change is to inform the called this is the final.


windy-wang:
This is duplicated PR

PR #551, opened 2020-04-07
add jason api for mod_hiredis

windy-wang:
Add jason api support for mod_hiredis so that it can be used with jason format.


crienzo (approved):

andywolk (changes requested):
Please set json_reply in all error cases.

PR #553, opened 2020-04-07
expose rtp_session and payload

windy-wang:
This provide API for custom module  access the rtp_session and payload directly.


dragos-oancea:
please change the commit message to say [core]

PR #562, opened 2020-04-15
[mod_silk] Fix mod_silk always init with maxaveragebitrate=20000

sunlin7:
Hi, I try enable mod_silk in FS and find SILK always has maxaveragebitrate=20000, it cause from the `switch_silk_fmtp_parse()` will seek maxaveragebitrate according the `codec_fmtp.actual_samples_per_second` but it's 0. 
The solution is, init the `actual_samples_per_second before` before calling `switch_silk_fmtp_parse()`. 
Please review the changes. Thanks.


sunlin7:
Hi @mjerris could you please review and approve this change? Thank you.

PR #578, opened 2020-04-24
This pull adds support for Vosk speech recognition server over websocket

nshmyrev:
For more details on Vosk see

https://github.com/alphacep/vosk-server

https://github.com/alphacep/vosk-api


andywolk:
Never mind the failed test message. We are working on that.

crienzo:
@nshmyrev we can host this in github.com/freeswitch/mod_vosk.  Or, if you wish, you can host the module in your org.  This would allow you to be maintainer w/ write access to the module.    In either case, we can add it to `modules.conf` here as long as the module is OK.

nshmyrev:
@crienzo Great idea, I'll work to make it standalone module first.

PR #586, opened 2020-04-27
[mod_sofia] Fix rtp config ipv4/ipv6 index confusion

TheJJ:
The index counter check for maximum configuration entries was swapped by mistake.

### DIFF
--- a/src/mod/endpoints/mod_sofia/sofia.c
+++ b/src/mod/endpoints/mod_sofia/sofia.c
@@ -5110,13 +5110,13 @@ switch_status_t config_sofia(sofia_config_t reload, char *profile_name)
 						}
 
 						if (strchr(ip, ':')) {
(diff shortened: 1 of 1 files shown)


src/mod/endpoints/mod_sofia/sofia.c
```diff
@@ -5110,13 +5110,13 @@ switch_status_t config_sofia(sofia_config_t reload, char *profile_name)
 						}
 
 						if (strchr(ip, ':')) {
-							if (profile->rtpip_index < MAX_RTPIP) {
+							if (profile->rtpip6_index < MAX_RTPIP) {
```
andywolk, line 5113: Should be rtpip_index6, not rtpip6_index

TheJJ: uh-oh you're right :D 
fixed.

PR #591, opened 2020-04-28
[mod_av] changes to solve some video conference recording issues

blackhorse-reddog:
s->url should be created by av_strdup instead of strdup, because ffmpeg's util.c uses av_freep(&s->url); to free it. Currently it can cause a crash.

If codec_id is AV_CODEC_IP_MPEG4 (which is choosen by ffmpeg if the file extension is mp4), timebase denominator should be 48000 instead of 90000: "timebase 1/90000 not supported by MPEG 4 standard, the maximum admitted value for the timebase denominator is 65535"

### DIFF
--- a/src/mod/applications/mod_av/avformat.c
+++ b/src/mod/applications/mod_av/avformat.c
@@ -382,7 +382,7 @@ static int mod_avformat_alloc_output_context2(AVFormatContext **avctx, AVOutputF
 #if (LIBAVCODEC_VERSION_INT < AV_VERSION_INT(58,7,100))
 		av_strlcpy(s->filename, filename, sizeof(s->filename));
 #else
-		s->url = strdup(filename);
+		s->url = av_strdup(filename);
 		switch_assert(s->url);
 #endif
 	}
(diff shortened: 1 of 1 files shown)


seven1240:
av part LGTM. not sure the recorder of `conference_video_set_floor_holder`

blackhorse-reddog:
"not sure the recorder of conference_video_set_floor_holder"
If a member arrives with the MFLAG_JOIN_VID_FLOOR (typically used for screen share), the video floor should be locked by this member. This is usually the case, but sometimes another member acquires it after the new member with the mentioned flag takes it (by conference_video_set_floor_holder(conference, member, SWITCH_TRUE);), but before it is locked (by conference_utils_set_flag(member->conference, CFLAG_VID_FLOOR_LOCK);). This modification - changing these two lines - solved the problem, but I didn’t dig deep (for example, I don’t know where the function call which has obtained the floor status comes from), so I don’t know if there’s a better solution. I'll make a call stack to check it (may be just tomorrow).

PR #614, opened 2020-05-07
mod_callcenter: fix tracking agent external calls 

wmasilva:
we cannot track multiple agents in a single channel because in the hangup hook is used the variable cc_tracked_agent and will use the last value in this variable.

For example:
	<extension name="Track agent">
		<condition field="destination_number" expression="^1234$">
			<action application="callcenter_track" data="1000@default"/>
			<action application="callcenter_track" data="1001@default"/>
			<action application="bridge" data="user/1001@${domain_name}"/>
		</condition>
	</extension>

This will increment external_call_count for agents 1000@default and 1001@default but on hangup it will only decrease agent 1001@default.

### DIFF
--- a/src/mod/applications/mod_callcenter/mod_callcenter.c
+++ b/src/mod/applications/mod_callcenter/mod_callcenter.c
@@ -3336,6 +3336,7 @@ SWITCH_STANDARD_APP(callcenter_track)
 	char agent_status[255];
 	char *agent_name = NULL;
 	char *sql = NULL;
+	const char *tracked_agent = NULL;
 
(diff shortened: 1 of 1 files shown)


crienzo (changes requested):
There are extra commits associated with this PR.  Please resubmit with only the mod_callcenter changes.

wirelessmundi:
i don't know why this happen... i just sync, rebase local branch and the pull request keep register everything.. 
do you know how i can fix it, or need to create a fresh pull request? 

Thanks.

PR #639, opened 2020-05-17
[Core] Fix the null char in truncated value returned by switch_b64_decode

windy-wang:
switch_b64_decode returns truncated value issue:

When `encode` "ABC" and `decode` back with a 4 bytes buffer, the value will be truncated to "AB" and the test case fails.

    chk_eq_str: 'AB' != 'ABC'

This patch fix it, and changed to return length not included the NULL char.

### DIFF
--- a/tests/unit/switch_utils.c
+++ b/tests/unit/switch_utils.c
@@ -76,7 +76,7 @@ FST_TEST_BEGIN(b64)
     switch_size_t size = switch_b64_decode((const char *)b64_str, decoded_str, sizeof(decoded_str));
     switch_log_printf(SWITCH_CHANNEL_LOG, SWITCH_LOG_INFO, "decoded_str: %s\n", decoded_str);
     fst_check_string_equals(decoded_str, str);
-    fst_check(size == 3);
+    fst_check(size == 4);
 }
 FST_TEST_END()
 
--- a/tests/unit/switch_utils.c
+++ b/tests/unit/switch_utils.c
@@ -23,7 +23,7 @@
  *
  * Contributor(s):
  * Seven Du <seven@signalwire.com>
- *
+ * Windy Wang <xiaofengcanyuexp@163.com>
  *
  * switch_utils.c -- tests switch_utils
  *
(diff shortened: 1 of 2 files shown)


seven1240:
@mjerris I agree that return value might change behaviour, but I think return the actual size of the data is good, though we always add an extra NULL in the end. The buffer size is known before calling this function anyway. 

Did a grep looks no where check the return value in the current code base so it could be safe.

mjerris:
> @mjerris I agree that return value might change behaviour, but I think return the actual size of the data is good, though we always add an extra NULL in the end. The buffer size is known before calling this function anyway.
> 
> Did a grep looks no where check the return value in the current code base so it could be safe.

We for sure should include the null in the return value, such as how system string functions that act on buffers do, and we should make sure to always null terminate like those as well.  will you make these changes?

tests/unit/switch_utils.c
```diff
@@ -63,6 +63,24 @@ FST_TEST_BEGIN(benchmark)
...
+    char decoded_str[4];
+    switch_status_t status = switch_b64_encode((unsigned char *)str, strlen(str), b64_str, sizeof(b64_str));
+    switch_log_printf(SWITCH_CHANNEL_LOG, SWITCH_LOG_INFO, "b64_str: %s\n", b64_str);
+    fst_check(status == SWITCH_STATUS_SUCCESS);
+    fst_check_string_equals((const char *)b64_str, "QUJD");
+
+    switch_size_t size = switch_b64_decode((const char *)b64_str, decoded_str, sizeof(decoded_str));
+    switch_log_printf(SWITCH_CHANNEL_LOG, SWITCH_LOG_INFO, "decoded_str: %s\n", decoded_str);
+    fst_check_string_equals(decoded_str, str);
+    fst_check(size == 3);
```
mjerris, line 79: this should be 4 not 3

windy-wang: Hi, Michale

I will do the change now.

Thanks for your comments.

windy-wang: Done, Please check.

seven1240 (dismissed):

PR #646, opened 2020-05-20
fixes default to dtls 1.2 on openssl 1.1.0 and later

vma:
The new openssl api functions `DTLS_server_method()` and `DTLS_client_method()` support both dtls v1.0 and dtls v1.2.

So if we want to limit only to dtls v1.2, we need to set the minimum protocol version as version specific methods (DTLSv1_2_server_method() and DTLSv1_2_client_method()) are deprecated in openssl 1.1.1.


vma:
I have updated the commit message as suggested.

mjerris:
This is going to break sending to older clients that only support 1.0

vma:
Wouldn't `want_DTLSv1_2` be false in this case (through legacyDTLS chan var)?

This patch mirrors the `#else` block just below which chooses the DTLSv1_2_XXX() methods when `want_DTLSv1_2` is set.

crienzo:
@mjerris is this OK now?

ukolovda:
I still have this error.
I have 2 FreeSwitch servers and 2 WebRTC clients and try bridge it.
If I replace 1 WebRTC client to simple SIP (I use MicroSIP), it works fine.

PR #656, opened 2020-05-25
[mod_mosquitto] Cleanup

NormB:
mod_mosquitto: Interface to MQTT brokers using Mosquitto
				Implements a Publish/Subscribe (pub/sub) messaging pattern using the mosquitto API library
				Publishes FreeSWITCH events to one more more MQTT brokers
				Subscribes to topics located on one more more MQTT brokers

MQTT http://mqtt.org/
Mosquitto https://mosquitto.org/


andywolk:
This is a standalone module now.
See https://github.com/freeswitch/mod_mosquitto

PR #698, opened 2020-06-17
Implement SEND_RTCP_MESSAGE

noahmehl:
We utilize: https://github.com/sipcapture/hepipe.js to monitor RTCP for our calls.  As such, it's super helpful to get the sent RTCP data along with the received RTCP data.  The new: `SEND_RTCP_MESSAGE` basically mirrors the existing `RECV_RTCP_MESSAGE`.


noahmehl:
This is a sponsored featured from the [Open Telecom Foundation](https://opentelecom.foundation).

noahmehl:
@andywolk @crienzo

Can this please be reviewed?

src/switch_rtp.c
```diff
@@ -2043,6 +2044,81 @@ static int using_ice(switch_rtp_t *rtp_session)
...
+#if SWITCH_BYTE_ORDER == __BIG_ENDIAN
+				tmpLost = report->lost; // signed 24bit will extended signess to int32_t automatically
+#else
+				tmpLost = ntohl(rtcp_report_block->lost)>>8;
+				tmpLost = tmpLost | ((tmpLost & 0x00800000) ? 0xff000000 : 0x00000000); // ...and signess compensation
+#endif
+				snprintf(value, sizeof(value), "%u", tmpLost);
+				switch_event_add_header_string(event, SWITCH_STACK_BOTTOM, header, value);
+				// snprintf(header, sizeof(header), "Source%u-Loss-Avg", i);
+				// snprintf(value, sizeof(value), "%u", rtp_session->rtcp_frame.reports[i].loss_avg);
```
dragos-oancea, line 2101: commented out code should be removed.

src/switch_rtp.c
```diff
@@ -2192,6 +2268,7 @@ static int check_rtcp_and_ice(switch_rtp_t *rtp_session)
 				rtcp_generate_report_block(rtp_session, rtcp_report_block, nack_dup);
 				rtp_session->rtcp_send_msg.header.count = 1; /* reception report block count */
 				stats->sent_pkt_count = 0;
+				sendRTCPEvent(rtp_session, sr, rtcp_report_block);
```
dragos-oancea, line 2271: call of sendRTCPEvent() should be under cond 'if (switch_channel_var_true(session->channel, "fire_rtcp_events")'

src/switch_rtp.c
```diff
@@ -2043,6 +2044,81 @@ static int using_ice(switch_rtp_t *rtp_session)
 	return 0;
 }
 
+static void sendRTCPEvent(switch_rtp_t *rtp_session ,struct switch_rtcp_sender_report *sr,struct switch_rtcp_report_block *rtcp_report_block) {
```
dragos-oancea, line 2047: bad naming. see the other function names in switch_rtp.c

src/switch_rtp.c
noahmehl: @dragos-oancea this should be complete.

src/switch_rtp.c
noahmehl: @dragos-oancea please review our update.

src/switch_rtp.c
noahmehl: @dragos-oancea please review our update.

noahmehl:
I have also rebased from master.

noahmehl:
Big shout out to @hthetiot for helping us get this wrapped up!

noahmehl:
@dragos-oancea can you please review?

noahmehl:
@crienzo any chance we can look at this?

crienzo:
@dragos-oancea if all your concerns are addressed, this seems fine

src/switch_rtp.c
```diff
@@ -2047,6 +2048,78 @@ static int using_ice(switch_rtp_t *rtp_session)
...
+
+				snprintf(value, sizeof(value), "%u", ntohl(sr->sender_info.ts));
+				switch_event_add_header_string(event, SWITCH_STACK_BOTTOM, "Last-RTP-Timestamp", value);
+
+				snprintf(value, sizeof(value), "%" SWITCH_TIME_T_FMT, switch_time_now());
+				switch_event_add_header_string(event, SWITCH_STACK_BOTTOM, "Capture-Time", value);
+
+				// Add sources info
+				snprintf(header, sizeof(header), "Source-SSRC");
+				snprintf(value, sizeof(value), "%.8x", rtp_session->stats.rtcp.ssrc);
```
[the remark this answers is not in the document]
hthetiot: @noahmehl 
- rtp_session->stats.rtcp.ssrc
+ rtcp_report_block->ssrc

ref: https://github.com/signalwire/freeswitch/blob/master/src/switch_rtp.c#L506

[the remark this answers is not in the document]
hthetiot: thx @rjm2718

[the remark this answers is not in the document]
noahmehl: @rjm2718 Thanks for the heads up! This has been updated.

noahmehl:
> Unit-tests failed: https://public-artifacts.signalwire.cloud/drone/signalwire/freeswitch/572/artifacts.html

Rebased on master and pushed.  Hopefully that will fixed the failing test.

noahmehl:
The tests are failing in openh264, I don't think this is related to this PR.

andywolk:
> The tests are failing in openh264, I don't think this is related to this PR.

Not related and that's already fixed. Let's see how next build runs. Should be green now.

hthetiot:
CI fail cause: No space left on device

> mawk: write failure (No space left on device)
mawk: close failed on file /dev/stdout (No space left on device)
config.status: error: could not create src/mod/event_handlers/mod_event_zmq/Makefile

src/switch_rtp.c
```diff
@@ -2049,6 +2050,78 @@ static int using_ice(switch_rtp_t *rtp_session)
...
+
+				snprintf(value, sizeof(value), "%u", ntohl(sr->sender_info.ts));
+				switch_event_add_header_string(event, SWITCH_STACK_BOTTOM, "Last-RTP-Timestamp", value);
+
+				snprintf(value, sizeof(value), "%" SWITCH_TIME_T_FMT, switch_time_now());
+				switch_event_add_header_string(event, SWITCH_STACK_BOTTOM, "Capture-Time", value);
+
+				// Add sources info
+				snprintf(header, sizeof(header), "Source-SSRC");
+				snprintf(value, sizeof(value), "%.8x", rtcp_report_block->ssrc);
```
rjm2718, line 2092: This still isn't correct, because rtcp_report_block->ssrc is written with network byte ordering, and we want host for this sprintf method.  From rtcp_generate_report_block():
`rtcp_report_block->ssrc = htonl(rtp_session->stats.rtcp.peer_ssrc);`
Sorry I didn't catch that sooner.  So the line this comment refers to should be:
`snprintf(value, sizeof(value), "%.8x", rtp_session->stats.rtcp.peer_ssrc);`

rjm2718 (changes requested):

src/switch_rtp.c
noahmehl: @rjm2718, thank you so much for finding this.  Branch is updated to reflect this, and rebased from master.

noahmehl:
@dragos-oancea @andywolk @crienzo can we merge this?  The checks have all passed, and @rjm2718 has been testing on his end as well.

src/include/switch_types.h
```diff
@@ -241,6 +241,7 @@ SWITCH_BEGIN_EXTERN_C
 #define SWITCH_DEFAULT_VIDEO_SIZE 1200
 #define SWITCH_RTCP_AUDIO_INTERVAL_MSEC "1000"
 #define SWITCH_RTCP_VIDEO_INTERVAL_MSEC "1000"
+#define SWITCH_RTCP_FIRE_EVENTS "rtcp_fire_events"
```
dragos-oancea, line 244: "rtcp_send_fire_events"

src/switch_rtp.c
```diff
@@ -2050,6 +2051,78 @@ static int using_ice(switch_rtp_t *rtp_session)
...
+				snprintf(value, sizeof(value), "%u", ntohl(sr->sender_info.oc));
+				switch_event_add_header_string(event, SWITCH_STACK_BOTTOM, "Octect-Packet-Count", value);
+
+				snprintf(value, sizeof(value), "%u", ntohl(sr->sender_info.ts));
+				switch_event_add_header_string(event, SWITCH_STACK_BOTTOM, "Last-RTP-Timestamp", value);
+
+				snprintf(value, sizeof(value), "%" SWITCH_TIME_T_FMT, switch_time_now());
+				switch_event_add_header_string(event, SWITCH_STACK_BOTTOM, "Capture-Time", value);
+
+				// Add sources info
```
dragos-oancea, line 2091: comments should be C style, eg:  /*  foo */

src/switch_rtp.c
```diff
@@ -2199,6 +2272,9 @@ static int check_rtcp_and_ice(switch_rtp_t *rtp_session)
 				rtcp_generate_report_block(rtp_session, rtcp_report_block, nack_dup);
 				rtp_session->rtcp_send_msg.header.count = 1; /* reception report block count */
 				stats->sent_pkt_count = 0;
+				if (switch_channel_var_true(switch_core_session_get_channel(rtp_session->session), SWITCH_RTCP_FIRE_EVENTS)) {
```
dragos-oancea, line 2275: you'll want to check a flag here, not a channel var. Also, this should be done only for AUDIO .

rjm2718: > you'll want to check a flag here, not a channel var. Also, this should be done only for AUDIO .

Why only audio?  The rtcp sender report is sent either way, so should the esl event too?
Regarding gating the event on a flag, what is the intention?  If we want this gated by flag, what sets or clears that flag?

Your comment from Dec 2th:
> call of sendRTCPEvent() should be under cond 'if (switch_channel_var_true(session->channel, "fire_rtcp_events")'

By comparison, firing of SWITCH_EVENT_RECV_RTCP_MESSAGE in switch_core_media.c happens in any case.

noahmehl:
Rebased on master...

src/switch_rtp.c
dragos-oancea: you're calling switch_core_session_get_channel() from where you shouldn't call it. it's got an assert in it, it will fail the assert and kill FS when there's no channel.  Also, switch_channel_var_true() calls switch_channel_get_variable_dup() which mutex locks and it will lock for every RTCP SR. That's why I suggested using a RTP session flag, you can see the pattern across the whole check_rtcp_and_ice() function.  For video the RTCP SR is much more often than for audio. If you need for video too add them on different flags at least (enable them separately). You'll get probably hundreds or even thousands of these events on a typical 1 hour video call.  To what purpose ?

dragos-oancea:
needs to be squashed with a proper explicit commit message, eg: "[core] add RTCP SR event (the SEND_RTCP_MESSAGE basically mirrors the existing RECV_RTCP_MESSAGE) "

src/switch_rtp.c
noahmehl: @dragos-oancea the point of receiving all of the RTCP events on the bus is to be able to send them to a system (e.g. Homer) to be able to monitor quality of all rtp session of all calls/channels.  This occurs currently for RTCP_Receive, we are just duplicating for SEND.  The other reason is that for many calls, there will be encryption, and therefore the packet capture method is not possible.  Also for many client (e.g. web browsers), it may not be possible to capture the RTCP events client side for monitoring.

dragos-oancea: just RTCP SR will not help very much with monitoring video quality, you'll have to do the same for PLI/FIR/NACK.     Just RTCP SR will give you an idea about packet loss and RTT, and the price to pay on generating and processing these events is too high, imo.

src/switch_rtp.c
dragos-oancea: I didn't mean to tell you put the condition here, but where the RTP session is created.

noahmehl:
@dragos-oancea thank you for the heads up about the squash commit message, we will get that done.

src/switch_rtp.c
rjm2718: @dragos-oancea now triggering event only for audio, or for video if rtp session flag set (which gets set if channel var is set)

src/switch_core_media.c
```diff
@@ -8646,6 +8646,8 @@ SWITCH_DECLARE(switch_status_t) switch_core_media_activate_rtp(switch_core_sessi
 		flags[SWITCH_RTP_FLAG_BYTESWAP] = 0;
 	}
 
+
+
```
dragos-oancea, line 8650: remove the new lines.

src/switch_rtp.c
```diff
@@ -2050,6 +2057,78 @@ static int using_ice(switch_rtp_t *rtp_session)
 	return 0;
 }
 
+static void switch_send_rtcp_event(switch_rtp_t *rtp_session ,struct switch_rtcp_sender_report *sr,struct switch_rtcp_report_block *rtcp_report_block) {
```
dragos-oancea, line 2060: "{" - open the curly bracket after a new line, not on the line with the function name.

src/switch_core_media.c
noahmehl: Done :)

src/switch_rtp.c
noahmehl: Done :)

noahmehl:
@dragos-oancea I have made the styling updates requested.  Also just rebased off of master and squashed the commits.  Please let us know if there's anything else needed here.

Thanks again for all of your help on this!

dragos-oancea (dismissed):

dragos-oancea:
the commit message should say "[core]"  and be more meaningful and succint.

src/switch_rtp.c
```diff
@@ -2199,6 +2279,11 @@ static int check_rtcp_and_ice(switch_rtp_t *rtp_session)
 				rtcp_generate_report_block(rtp_session, rtcp_report_block, nack_dup);
 				rtp_session->rtcp_send_msg.header.count = 1; /* reception report block count */
 				stats->sent_pkt_count = 0;
+				/* send event for audio, or video if flag set */
+				if (rtp_session->flags[SWITCH_RTP_FLAG_VIDEO_FIRE_SEND_RTCP_EVENT] ||
```
dragos-oancea, line 2283: we had an internal discussion about this patch and we need more clarifications about the run conditions here. I thought I made it clear in my previous comments that we don't want this code run by default.  Also the naming of the vars/flags is confusing, eg: "SWITCH_RTP_FLAG_VIDEO_FIRE_SEND_RTCP_EVENT" - is this meant to work only for video ? according to this conditions it fires for audio every time.

dragos-oancea: just make it so that it fires for both audio and video with the same code - and use the flag to disable all the patch.

noahmehl: @dragos-oancea I don't fully understand this.  `RECV_RTCP_MESSAGE` fires by default now, so I don't understand why `SEND_RTCP_MESSAGE` wouldn't?  And yes, the flag is to enable the `SEND_RTCP_MESSAGE` for video, in addition to audio.

rjm2718: My interpretation is that video is the greater concern, with respect to load on the events system.  If audio is also a concern, I propose that we have a config setting, and a channel var to override, that sets the rtp session flag, that enables RTCP events either direction, for each audio or video separately.  so, two flags:
SWITCH_RTP_FLAG_RTCP_EVENTS_VIDEO
SWITCH_RTP_FLAG_RTCP_EVENTS_AUDIO
The flag toggles send/recv events together, for video and audio respectively.

With a config and/or channel var, plus rtcp-audio/video-interval-msec, this gives the user pretty fine control over these events, and hence the load on the events system.  Enable audio, disable video by default.

dragos-oancea: "I don't fully understand this. RECV_RTCP_MESSAGE fires by default now" - this will change soon with upcoming code which already exists but it's not merged yet. The creation and enqueuing of these events will be conditioned by a bool chan var named "fire_rtcp_events". I agree with the two flags, the whole idea is to NOT have this code run by default, even if you rightfully think it's consistent with the existing code.

rjm2718: ok, that makes more sense now.  Can you point to that existing code that's not yet merged?  I'll try to make the code here congruent with "fire_rtcp_events".  Please advise.

dragos-oancea: no, I can't point you to private code. all you need to do is enable your whole patch based on a chan var named "fire_rtcp_events".

rjm2718: that implies one rtp session flag, SWITCH_RTP_FLAG_FIRE_RTCP_EVENTS, that toggles events for both video and audio, for both send & recv.  agree?

noahmehl: @dragos-oancea @rjm2718 it would still be helpful to have a config setting for this as well, as we don't want to *have* to update our applications to send new channel vars, if we're going to do it on every session anyways.

dragos-oancea: @rjm2718  something like this is acceptable, for more fine-tuning:
```
if (switch_channel_var_true(session->channel, "fire_rtcp_events")) {
                       flags[SWITCH_RTP_FLAG_AUDIO_FIRE_SEND_RTCP_EVENT] = 1;
			if (switch_channel_var_true(session->channel, "enable_send_rtcp_message_event_video")) { 
				flags[SWITCH_RTP_FLAG_VIDEO_FIRE_SEND_RTCP_EVENT] = 1;
			}
}
```

dragos-oancea: @noahmehl , config setting ? you can just use the chan vars as "config settings". you can set them in vars.xml and have them on all channels.

noahmehl: @dragos-oancea understood.  Thanks!

src/switch_core_media.c
```diff
@@ -9504,6 +9507,11 @@ SWITCH_DECLARE(switch_status_t) switch_core_media_activate_rtp(switch_core_sessi
 			if (v_engine->tmmbr) {
 				flags[SWITCH_RTP_FLAG_TMMBR]++;
 			}
+
+			if (switch_channel_var_true(session->channel, "fire_rtcp_events") &&
```
dragos-oancea, line 9511: why are you checking the chan var two times ? should be only once.

src/switch_rtp.c
```diff
@@ -2199,6 +2279,10 @@ static int check_rtcp_and_ice(switch_rtp_t *rtp_session)
 				rtcp_generate_report_block(rtp_session, rtcp_report_block, nack_dup);
 				rtp_session->rtcp_send_msg.header.count = 1; /* reception report block count */
 				stats->sent_pkt_count = 0;
+				if ((rtp_session->flags[SWITCH_RTP_FLAG_VIDEO] && rtp_session->flags[SWITCH_RTP_FLAG_VIDEO_FIRE_SEND_RTCP_EVENT]) ||
+				    (!rtp_session->flags[SWITCH_RTP_FLAG_VIDEO] && rtp_session->flags[SWITCH_RTP_FLAG_AUDIO_FIRE_SEND_RTCP_EVENT])) {
```
dragos-oancea, line 2283: evaluate this first: ```!rtp_session->flags[SWITCH_RTP_FLAG_VIDEO] && rtp_session->flags[SWITCH_RTP_FLAG_AUDIO_FIRE_SEND_RTCP_EVENT])``` .  not sure about what happens the text mode (SWITCH_RTP_FLAG_TEXT). did you test it ? maybe we should not send these events in this mode.

src/switch_core_media.c
rjm2718: @dragos-oancea given the complexity of the method it's in, I thought it safest to put the check closely above the switch_rtp_new() call.  the way I did it, it's clear what applies to audio vs video vs text (no flag is set for text).  I'll combine them if that's better -- I think anywhere below switch_assert(session) will work, but let me know.

src/switch_rtp.c
rjm2718: per comment above, condition will be false for text mode

dragos-oancea:
we'll probably merged this branch where I rebased and cherry-picked your commits: https://github.com/signalwire/freeswitch/pull/1370 . I'm also in process of adding an unit-test.

dragos-oancea:
we merged https://github.com/signalwire/freeswitch/pull/1370 . I am closing this one. thanks.

PR #709, opened 2020-06-24
[core] Exact cipher matching by string splitting

### LINKED ISSUE #714: [core] Exact cipher matching by string splitting
This request build on the previous one as it was discovered that it didn't fully resolve the issues with the Cisco phones. I believe that the Cisco devices are also relying on substring matches, meaning the shorter ciphers need to be placed first, This causes issues where FreeSWITCH behaves in the same way and I originally resolved by adding a space into the string.

This pull request splits the string and checks for an exact match in the second field, eliminating the need for the space in the string and meaning that as far as FreeSWITCH is concerned, the cipher order is now less relevant. In my tests, calls in both directions now function correctly using the AEAD_AES_265_GCM codec.

zandeez:
This request build on the previous one as it was discovered that it didn't fully resolve the issues with the Cisco phones. I believe that the Cisco devices are also relying on substring matches, meaning the shorter ciphers need to be placed first, This causes issues where FreeSWITCH behaves in the same way and I originally resolved by adding a space into the string.

This pull request splits the string and checks for an exact match in the second field, eliminating the need for the space in the string and meaning that as far as FreeSWITCH is concerned, the cipher order is now less relevant. In my tests, calls in both directions now function correctly using the AEAD_AES_265_GCM codec.

Resolves #714


zandeez:
Rebased on current master head.

dragos-oancea:
We know sometimes it may be too much to ask, but adding unit-tests to this sort of core patches would speed-up merging considerably.

zandeez:
Rebased on master, Claude generated unit tests (visually look fine) also pulled in semi-related PR #951 including the suggested use of the preexisting function.

PR #716, opened 2020-06-27
mod_kazoo SmartOS Portability Updates

noahmehl:
After testing we have found one small item that needed update for mod_kazoo to work in SmartOS.

### DIFF
--- a/src/mod/event_handlers/mod_kazoo/kazoo_ei_utils.c
+++ b/src/mod/event_handlers/mod_kazoo/kazoo_ei_utils.c
@@ -333,6 +333,12 @@ switch_socket_t *create_socket_with_port(switch_memory_pool_t *pool, switch_port
 		return NULL;
 	}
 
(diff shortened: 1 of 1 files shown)


noahmehl:
@lazedo we're pretty happy with this, but please comment if you'd like any changes.

lazedo:
@noahmehl can you explain why this is an issue ? and only for `#ifdef __sun` ?
besides that, it seems you are making the change in the wrong place, this function is shared by the node and eventstream acceptors and eventstream explicitly sets this.

noahmehl:
@lazedo from what we can tell: create_socket_with_port uses the embedded apr lib to create a socket, and for SunOS derivatives, the NON_BLOCK check is required, otherwise all TCP actions fail here.

We wrapped in a sunos macro just to be safe, as this doesn't seem to be the same behavior with linux, etc.

PR #719, opened 2020-06-30
[mod_opusfile] fix compiling write file support

### LINKED ISSUE #718: mod_opusfile: Makefile.am does not honor configure vars
Makefile.am for mod_opusfile is buggy: Does not "pick up" the configure results, i.e. configure finds libopusenc and enables it. However, the related *.c files are compiled without the proper flags and thus writing/recording opus files will fail with a generic error (and no log.error gets emitted explaining the problem =8-( ).

jelmd:
fix #718 : honor $OPUSFILE_ENCODE_CFLAGS and -DHAVE_OPUSFILE_ENCODE when compiling mod_opusfile.c

### DIFF
--- a/src/mod/formats/mod_opusfile/Makefile.am
+++ b/src/mod/formats/mod_opusfile/Makefile.am
@@ -4,15 +4,16 @@ MODNAME=mod_opusfile
 noinst_LTLIBRARIES = libopusfilemod.la
 
 libopusfilemod_la_SOURCES  = mod_opusfile.c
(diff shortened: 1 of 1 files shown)


dragos-oancea:
can you change the commit message to add square brackets ? [mod_opusfile]

jelmd:
You mean as prefix? Instead of `mod_opusfile:` use `[mod_opusfile]`?

dragos-oancea:
yes...like the other commit messages are.

andywolk:
Please remove the fixes part from the commit message.

PR #726, opened 2020-07-09
[switch_rtp] switch_rtp_add_dtls: fix memory leak.

sunlin7:
Hi,

The `dtls_bio_filter_methods` is used as global static instance in `switch_rtp_add_dtls()`, similar its behavior when OPENSSL_VERSION_NUMBER < 0x10100000L. So it should be allocated just for once. But currently it will be allocated for each calling `switch_rtp_add_dtls()`, that will cause memory leak. 
The change will try fix the memory leak.

Regards
Lin


sunlin7:
Hi @andywolk 
Could you review and approve this change please? Thanks

Regards
Lin

andywolk:
Thank you. Yes that's a bug. Although solution is not quite right.
There is a leak still on shutdown (BIO_meth_free() never called) plus that change introduces a race condition.
See https://github.com/signalwire/freeswitch/pull/727

sunlin7:
Yes, totally, and your solution #727 can resolve this issue. Thanks.

PR #727, opened 2020-07-09
[Core] Fix leak of BIO_METHOD *dtls_bio_filter_methods in switch_rtp_add_dtls()

andywolk:


sunlin7:
@andywolk the declare of `*dtls_bio_filter_methods` should move to the position before `switch_rtp_init()`, or there is compile error that undefined `dtls_bio_filter_methods`.
```
#if OPENSSL_VERSION_NUMBER < 0x10100000L
static BIO_METHOD dtls_bio_filter_methods;
#else
static BIO_METHOD *dtls_bio_filter_methods;
#endif
```

andywolk:
Yes, that's true, I missed that. This is why we have Drone CI integrated with github to check the build. And Drone is in maintenance right now. So this PR will not be merged until Drone is back and says OK.

PR #733, opened 2020-07-16
[mod_sofia] Fix crash when proxy INFO without body

MikhailKalashnikov:
FS crashes when proxies INFO without a body.

In sip_profile parameter `proxy-info=true`

Example of INFO sent via sipp
```
INFO sip:STAS@127.0.0.1:11000;transport=UDP SIP/2.0
      Via: SIP/2.0/[transport] [local_ip]:[local_port];branch=[branch]
      Route: <sip:192.168.0.34:5063;lr>
      From: sipp <sip:sipp@[local_ip]:[local_port]>;tag=[pid]SIPpTag00[call_number]
      To: [service] <sip:[service]@[remote_ip]:[remote_port]>[peer_tag_param]
      Call-ID: [call_id]
      CSeq: 2 INFO
      Contact: sip:sipp@[local_ip]:[local_port]
      Max-Forwards: 70
```


andywolk:
Good catch! Thank you for the PR.

PR #770, opened 2020-08-04
[switch_utils.c] optimization for switch_cut_path() function.

sunlin7:
Hi, 
Here is the optimization for `switch_cut_path()` function, it improve about 50%.
$ gcc -std=c11 test-switch_cut_path.c && ./a.out
func[0] take time 9 s
func[1] take time 15 s
The testing code is
```
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <time.h>
const char *switch_cut_path(const char *in) {
  const char *p, *ret = in;
  const char delims[] = "/\\";
  const char *i;

  if (in) {
    for (i = delims; *i; i++) {
      p = in;
      while ((p = strchr(p, *i)) != 0) {
        ret = ++p;
      }
    }
    return ret;
  } else {
    return NULL;
  }
}

const char *switch_cut_path_new(const char *in) {
#ifdef WIN32
  const char delim = '\\';
#else
  const char delim = '/';
#endif
  const char *p = in, *ret = in;

  if (NULL == in) {
    return NULL;
  }
  while ((p = strchr(p, delim)) != NULL) {
    ret = ++p;
  }
  return ret;
}

int main(int argc, char *argv[])
{
  char seps[] = "/\\";
  char path[] = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ";
  const char *(*funcs[])(const char *) = {switch_cut_path_new, switch_cut_path};
  for (int i = 0; i < sizeof(funcs)/sizeof(funcs[0]); i++) {
    time_t t1 = time(NULL);
    for (int j = 0; j < 1024 * 1024 * 200; j++) {
      int idx = j % sizeof(path);
      char orig = path[idx];
      path[idx] = '/';
      funcs[i](path);
      path[idx] = '\\';
      funcs[i](path);
      path[idx] = orig;
    }
    time_t t2 = time(NULL);
    printf("func[%d] take time %ld s\n", i, t2 - t1);
  }
  return 0;
}
```


sunlin7:
@andywolk Could you help to review this change please? Thanks

mjerris:
This changes behavior of the function when either using mixed / and \ or when using unix delims on windows.   If you want to optimize this function the first step would be to do a set of unit tests including those scenarios.  Is there a compelling reason to optimize this function, if so please provide some detail of the scenario that has this function being called frequently

mjerris:
any optimization or refactoring type changes should start with a complete set of unit tests to confirm we don’t break anything in the process.

sunlin7:
@mjerris I turn on all level log, and the `switch_cut_path()` will be called for each log print, so I try to make the calling faster.
I had update the change list for adding test cases in the `tests/unit/switch_utils.c`. 
Please notice that the original function maybe has problem for something like `a\\b/c` , but new function will give correct result.
Here's an example:
$ `makedir -p a\\b/ &&　touch a\\b/c`
original `switch_cut_path("a\\b/c")` return "b/c"
new  `switch_cut_path("a\\b/c")` return "c"

mjerris:
does the new way will break cutting unix style paths on windows?

sunlin7:
Hi @mjerris The updated changes won't break cutting unix style paths on windows. The changes didn't depend on the macro `WIN32`,  always search for both `/` and `\\`, ack like it's  preview version.

sunlin7:
The unit testing case will verify it's behavior.
https://github.com/signalwire/freeswitch/blob/c1f604178f60d300ce0206a646aa9e377a847b04/tests/unit/switch_utils.c#L65-L68

sunlin7:
@mjerris The new version release days ago, shall we merge this bug fix then it will get wide testing?

sunlin7:
@mjerris Could you review this bug fix please? It will fix the incorrect path spliting on *nix, and keep behavior on Windows.

PR #787, opened 2020-08-12
Added variable sip_contact_host to change the host part in the Contact header in INVITE requests

jfirles:
Some providers like Microsoft Teams require set the host part in the "Contact" header with the required value for auth the INVITE request, with this pull request we can change the contact host part of that header.

This is useful for multitenant scenarios.

### DIFF
--- a/src/mod/endpoints/mod_sofia/sofia_glue.c
+++ b/src/mod/endpoints/mod_sofia/sofia_glue.c
@@ -1270,7 +1270,24 @@ switch_status_t sofia_glue_do_invite(switch_core_session_t *session)
 
 		if (zstr(tech_pvt->invite_contact)) {
 			const char *contact;
-			if ((contact = switch_channel_get_variable(channel, "sip_contact_user"))) {
+			const char *contactHost;
+			if ((contact = switch_channel_get_variable(channel, "sip_contact_user")) && (contactHost = switch_channel_get_variable(channel, "sip_contact_host"))) {
+                                if (sofia_glue_transport_has_tls(tech_pvt->transport)) {
+                                        tech_pvt->invite_contact = switch_core_session_sprintf(session, "sip:%s@%s:%d", contact,
+                                                                                                                                                   contactHost, tech_pvt->profile->tls_sip_port);
+                                } else {
+                                        tech_pvt->invite_contact = switch_core_session_sprintf(session, "sip:%s@%s:%d", contact,
+                                                                                                                                                   contactHost, tech_pvt->profile->extsipport);
+                                }
+                        } else if ((contactHost = switch_channel_get_variable(channel, "sip_contact_host"))) {
+				if (sofia_glue_transport_has_tls(tech_pvt->transport)) {
+                                        tech_pvt->invite_contact = switch_core_session_sprintf(session, "sip:mod_sofia@%s:%d", contactHost,
+                                                                                                                                                   tech_pvt->profile->tls_sip_port);
+                                } else {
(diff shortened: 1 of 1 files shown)


wmasilva:
in current version, deecaae, when using a gateway we can set the contact-host parameter, so the contact is correctly changed.

I do find an issue, the incoming call from MS TEAMS, is authenticated by IP, so it doesn't hit the gateway detection so the replies we send back have the sipip address of our profile. We can set the  sip-ip address to the domain name, but for multi-tenant is not valid. 

We need to set it from the dialplan, the the incoming invite reach the default context. 
(if there is a way of identifying a gateway by ip, please let me know :) ) 

I test this code and it doesn't solve the issue, the 200OK sent have in contact the profile ip. 
i suggest the following fix: #1189 

attach the log and my configuration simulating this scenario. 
[freeswitch-pr787.log.txt](https://github.com/signalwire/freeswitch/files/6436556/freeswitch-pr787.log.txt)
[freeswitch.xml.txt](https://github.com/signalwire/freeswitch/files/6436561/freeswitch.xml.txt)

jfirles:
I'm using it for multitenant scenario.

You must set the sip-ip to de base domain and in dialplan the sip_contact_host var.

Best regards.

wmasilva:
Hi, 
for outgoing calls it works, what i found is for incoming calls, the 200OK will have the value you put in sip-ip. 

In the log, the variable is set correctly: 
2021-05-06 19:16:36.395925 [DEBUG] mod_dptools.c:1685 SET sofia/a/user@192.168.1.83 [sip_contact_host]=[vm.lab.local]

but the 200OK Contact uses value of sip-ip
Contact: <sip:123@192.168.1.83:5060;transport=udp>


for what i could test the "sofia_glue_do_invite" function is not called when receiving a call.

jfirles:
> Hi,
> for outgoing calls it works, what i found is for incoming calls, the 200OK will have the value you put in sip-ip.
> 
> In the log, the variable is set correctly:
> 2021-05-06 19:16:36.395925 [DEBUG] mod_dptools.c:1685 SET sofia/a/user@192.168.1.83 [sip_contact_host]=[vm.lab.local]
> 
> but the 200OK Contact uses value of sip-ip
> Contact: sip:123@192.168.1.83:5060;transport=udp
> 
> for what i could test the "sofia_glue_do_invite" function is not called when receiving a call.

Yes in the 200 OK the contact is the sip-ip value, but this works for Teams integration.
This is an example of 200 OK from my sbc to Teams:

> SIP/2.0 200 OK
> Via: SIP/2.0/TLS 52.114.76.76:5061;branch=z9hG4bK486d67fb;rport=7745
> Record-Route: <sip:sip-du-a-eu.pstnhub.microsoft.com:5061;transport=tls;lr>
> FROM: Teams User<sip:+34XXXX@sip.pstnhub.microsoft.com:5061;user=phone>;tag=9ef297297e464f578c6f5456b136d282
> To: <sip:+34XXXXXXXXX@customer1.sbc.mycompany.com:5061;user=phone>;tag=SUeUFXNj0BpZD
> CALL-ID: 561c7283db84588fac46ec9bb5435997
> CSEQ: 1 INVITE
> Contact: <sip:+34XXXXXXXXX@sbc.mycompany.com:5061;transport=tls>
> User-Agent: Siptize Teams SBC
> Allow: INVITE, ACK, BYE, CANCEL, OPTIONS, MESSAGE, INFO, UPDATE, REGISTER, NOTIFY
> Require: timer
> Supported: timer, path, replaces
> Allow-Events: talk, hold, conference, refer
> Session-Expires: 300;refresher=uac
> Content-Type: application/sdp
> Content-Disposition: session
> Content-Length: 793
> 
> SDP CONTENT

PR #791, opened 2020-08-13
[verto-client] move verto client to github.com/freeswitch/verto-client

crienzo:


andywolk:
@crienzo We should edit links in `scripts\FreeSWITCH-debian-raspbian-installer.sh` in the case someone uses it.

andywolk (approved):

PR #792, opened 2020-08-13
doc:  add missing parameters to switch_core_media_bug_add()

NormB:
Add the following parameter definitions:

\param function user defined module/function/reason identifying this bug
\param target user defined identification of the target of the bug


crienzo:
see #795

PR #793, opened 2020-08-13
[Build-System] Use spandsp3 package

andywolk:


andywolk:
> Yes, I packed spandsp 1.99 from files.freeswitch.org and the SPEC inside the tarball points to spandsp-1.99 not spandsp3
I updated the tarball. Should be fine now:
https://files.freeswitch.org/downloads/libs/spandsp-3.0.0-6ec23e5a7e.tar.gz

PR #795, opened 2020-08-14
[Documentation] Fix parameter definitions in switch_core_media_bug_add()

NormB:


crienzo:
@andywolk this is OK- commit message might need to be changed

PR #809, opened 2020-08-18
#807 set cid_type on tech_pvt

### LINKED ISSUE #807: Can't set cid_type to pid
 - Version master

**Describe the bug**

I'm trying to use cid_type for custom extensions (which is not bridged to any user). 

I defined following param in profile and FS sends `P-Asserted-Identity` on bridged calls.

```
        <param name="caller-id-type" value="pid"/>
```

However this is not working for non-bridged calls e.g. an extension with simple `queue` application.

I've debugged a bit and `tech_pvt->cid_type` is always 0. Probably we're not setting it from `caller-id-type` or `sip_cid_type` therefore it just uses default value which is `rpid`.

https://github.com/signalwire/freeswitch/blob/master/src/mod/endpoints/mod_sofia/mod_sofia.c#L257

zek:
Closes #807 

@mjerris

### DIFF
--- a/src/mod/endpoints/mod_sofia/sofia_glue.c
+++ b/src/mod/endpoints/mod_sofia/sofia_glue.c
@@ -133,6 +133,7 @@ void sofia_glue_attach_private(switch_core_session_t *session, sofia_profile_t *
 		}
 	}
 
+	tech_pvt->cid_type = tech_pvt->profile->cid_type;
 	tech_pvt->session = session;
 	tech_pvt->channel = switch_core_session_get_channel(session);
 


crienzo:
Will RPID be used if the profile is set to `"pid"` and the dialstring has `sip_cid_type=rpid` ?  I think you have to set this like 
```
		if ((val = switch_channel_get_variable(channel, "sip_cid_type"))) {
			cid_type = sofia_cid_name2type(val);
		}
		tech_pvt->cid_type = cid_type;
```
in `sofia_glue_do_invite()`

PR #818, opened 2020-08-26
modify log set

mangshe0:
modify log set


seven1240:
The original code was like this change and tony intentionally changed to less verbose.

PR #820, opened 2020-08-26
fix mod_av record crash bug.

mangshe0:
fix mod_av record crash bug
it will lead  ffmpeg-4.1\libavformat\utils.c(4424):	if (s->url) {
   av_freep(&s->url);	  
}
this time freep will dump.

### DIFF
--- a/src/mod/applications/mod_signalwire/mod_signalwire.c
+++ b/src/mod/applications/mod_signalwire/mod_signalwire.c
@@ -86,8 +86,12 @@ static struct {
 
 	char relay_connector_id[256];
 	
+#if SIGNALWIRE_CLIENT_C_VERSION_MAJOR >= 2
+	swclt_sess_t *signalwire_session;
+#else
 	swclt_sess_t signalwire_session;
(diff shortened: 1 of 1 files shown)


src/mod/applications/mod_av/avformat.c
```diff
@@ -382,7 +382,8 @@ static int mod_avformat_alloc_output_context2(AVFormatContext **avctx, AVOutputF
 #if (LIBAVCODEC_VERSION_INT < AV_VERSION_INT(58,7,100))
 		av_strlcpy(s->filename, filename, sizeof(s->filename));
 #else
-		s->url = strdup(filename);
+		s->url = av_mallocz(strlen(filename));
```
seven1240, line 385: should we do strlen() + 1 ?

also need to put the strcpy after switch_assert

mangshe0: ye, you are right,  i usually c++ string , it`s long time forget it , sorry for that.
i will recommit again.
thank you seven du

PR #65, opened 2019-10-18
FS-9676 [mod_http_cache] Support for AWS Signature version 4

baonq-me:
This PR adds support for AWS Signature version 4 into mod_http_cache. It also drops support for version 2 which is currently running.

I also add an option `backup-folder` into `http_cache.conf.xml` to write uploaded files (using `http_put`) to it in case of failure (for example recording a .wav file to a storage service).

PR #831, opened 2020-08-28
fix core dump on S3-authenticated HTTP GET

vornado22:
FS-9676 (https://github.com/signalwire/freeswitch/pull/65/files/7ad77d8a3ca6954a32d3aaf6bacbdbba189fd46e) was only partially merged; looks like you intended to not merge portions of it due to it containing an unrelated feature request, but part of the necessary code was also missed.

The result is that master is core dumping when attempting to authenticate to S3 on an HTTP GET through mod_http_cache.

This PR incorporates the rest of the original S3 v4 authentication code (without reintroducing the other feature you intended to ignore)


crienzo:
This crashes if query_string is NULL because of unsafe free of full_url.  Closing as we have another PR that is closer to the right fix.

PR #845, opened 2020-09-08
[Core] Create x509v3 certificates

fippo:
following the advice from
  https://bugs.chromium.org/p/chromium/issues/detail?id=1123946#c12
since boringssl in recent versions enforces x509v3 certificates
when using extensions like subjectAltName


mjerris (approved):

fippo:
you folks are too fast -- I only got to pester @anthmFS via slack ;-)

mjerris:
sometimes patches are easy and straightforward... frequently we are not so lucky :D

PR #850, opened 2020-09-11
[core] Enable MSRP over IPv6

cdosoftei:
Enables MSRP listening over IPv6, including dual stack binding.

### DIFF
--- a/src/switch_msrp.c
+++ b/src/switch_msrp.c
@@ -267,7 +267,7 @@ static switch_status_t msock_init(char *ip, switch_port_t port, switch_socket_t 
 #ifdef WIN32
 	/* Enable dual-stack listening on Windows */
 	if (switch_sockaddr_get_family(sa) == AF_INET6) {
-		rv = switch_socket_opt_set(*sock, 16384, 0);
+		rv = switch_socket_opt_set(*sock, SWITCH_SO_IPV6_V6ONLY, 0);
 		if (rv) goto sock_fail;
 	}
 #endif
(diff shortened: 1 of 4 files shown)


src/switch_msrp.c
```diff
@@ -252,15 +252,26 @@ static switch_status_t msock_init(char *ip, switch_port_t port, switch_socket_t
...
 	rv = switch_socket_create(sock, switch_sockaddr_get_family(sa), SOCK_STREAM, SWITCH_PROTO_TCP, pool);
 	if (rv) goto sock_fail;
 
 	rv = switch_socket_opt_set(*sock, SWITCH_SO_REUSEADDR, 1);
 	if (rv) goto sock_fail;
 
+#ifdef WIN32
+	/* Enable dual-stack listening on Windows */
+	if (switch_sockaddr_get_family(sa) == AF_INET6) {
+		rv = switch_socket_opt_set(*sock, 16384, 0);
```
andywolk, line 270: Please add this definition to the `switch_apr.h`
```
#define SWITCH_SO_IPV6_V6ONLY 16384 /* Don't accept IPv4 connections on an IPv6 listening socket. */
```

andywolk (changes requested):

src/switch_msrp.c
cdosoftei: Appreciate the advice! I took the liberty to update the other IPv6 use cases so the code is more readable/consistent.

andywolk (approved):

PR #853, opened 2020-09-15
[mod_dptools] allow inline dialplan to proceed to next dialplan handler

lazedo:
support for verto.conf.xml
<param name="dialplan" value="inline,XML:$${conf_dir}/autoload_configs/verto.dialplan.xml,XML"/>


mjerris (changes requested):
apps do not require arguments.  can you include in commit note example of how this is used

PR #854, opened 2020-09-16
[mod_sofia] add queue monitoring

lazedo:
requires https://github.com/freeswitch/sofia-sip/pull/28


mjerris (changes requested):
please ifdef this based on sofia version or feature flag define you add to sofia so it just doesn’t provide this with older sofia but still builds

lazedo:
@mjerris done

PR #855, opened 2020-09-16
[core] disallow duplicated entries in switch_event array

lazedo:
* fixes originated legs (and others) when using user endpoint

### DIFF
--- a/src/switch_event.c
+++ b/src/switch_event.c
@@ -1052,6 +1052,14 @@ static switch_status_t switch_event_base_add_header(switch_event_t *event, switc
 				}
 				goto end;
 			} else {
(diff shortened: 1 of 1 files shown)


mjerris (commented):
can you provide some more details of what the issue is here?

lazedo:
when you bridge to user/someuser and that user as a dial string to sofia/uri the `originated_legs` has duplicated entries. this happens for other variables (arrays) where items are duplicated

anthmFS:
What about when someone wants to have duplicate entries in arrays. In programming languages, arrays don't enforce uniqueness, its an operation of chosen behavior so it seems overreaching to force it.

PR #869, opened 2020-09-23
[api] normalized API response preamble

cdosoftei:
Minor adjustments so API responses are more consistent: positive results always prefixed with `+OK` whereas failure messages commence with `-ERR`.


crienzo:
Probably can't make this change without a major version increment.

seven1240:
It's a nice change. We can make it in 1.11, or 2.0 ;)

PR #824, opened 2020-08-27
AV timebase improve

seven1240:
replaces #819

PR #873, opened 2020-09-25
FIX parse config param "profile" and "level" in mod_av

ros-tel:
There was a replacement in the loop with default values

### DIFF
--- a/src/mod/applications/mod_av/avcodec.c
+++ b/src/mod/applications/mod_av/avcodec.c
@@ -1964,6 +1964,9 @@ static void parse_profile(avcodec_profile_t *aprofile, switch_xml_t profile)
 
 	ctx = &aprofile->ctx;
 
+	ctx->profile = FF_PROFILE_H264_BASELINE;
+	ctx->level = 31;
+
 	for (param = switch_xml_child(profile, "param"); param; param = param->next) {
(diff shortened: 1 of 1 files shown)


seven1240:
it is covered in https://github.com/signalwire/freeswitch/pull/824/files , can you also try that patch?

ros-tel:
> it is covered in https://github.com/signalwire/freeswitch/pull/824/files , can you also try that patch?

Faced the problem of incorrect application of the config as part of solving another problem (lack of fmtp for h264 [h264-default-fmtp](https://github.com/ros-tel/freeswitch/tree/h264-default-fmtp)) 
Therefore, I made a separate pull request.

I don't even know how to check your patch.
The only softphone that I was running under Linux with h264 Blink support and it does not decode the incoming stream (Linphone VP8> FS> H264 Blink).
What exactly is the problem I cannot figure out.

sergey-safarov:
In my use case, I also faced with not proper config parsing.
I wrote the same patch for this issue.
https://github.com/sergey-safarov/freeswitch/commit/a5f757ba8316e072174fc73000c927ea5a404c97

So, please merge.

dragos-oancea:
please add prefix "[mod_av]" to commit message.

ros-tel:
#1495

PR #875, opened 2020-09-27
[core] Fix ICE candidate priorities

traud:
Seven years ago, commit cb076e6 used xor as pow. The compiler Clang 10
errors about this copy-paste-error from RFC 5245 section 4.1.2.1. Now,
Clang compiles successfully.


mjerris:
This change looks right. what testting have you done on it?

mjerris (approved):

PR #877, opened 2020-09-29
[mod_av] Update error message

traud:
Five years ago, commit b29174e added the scale/conversion library of
FFmpeg. However, that library is not part of libavformat-dev and added
an additional requirement. That was not reflected in the error message,
when FFmpeg libraries are missing.

### DIFF
--- a/src/mod/applications/mod_av/Makefile.am
+++ b/src/mod/applications/mod_av/Makefile.am
@@ -36,5 +36,5 @@ else
 install: error
 all: error
 error:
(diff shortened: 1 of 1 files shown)


traud:
@mjerris `scripts/debian_min_build.sh` might need `libswscale-dev` as well. However, that script was created after the blaming commit. So I am confused why `libswscale-dev` is not in the script already. Is it installed as dependency of another package? I would prefer to mention it explicitly because that dependency could vanish on day.

crienzo (approved):

Issue #885, opened 2020-10-03
Registrations from "!"


pikachu937:
There is:
 stock fs (stock config, current profile is selected by default internal)
 2 user 1001 and 1002
 domain 192.168.0.1
 authorization without registration is disabled in the profile (login and pass must be entered, without this there is no way)
 Actually, he registers from both accounts and calls each other, everything seems to be fine, but there is one BUT!
 For some reason I do not understand, I can register under the user "!"(Although there is no such thing at all) with the password from the first real user from the list, in this case 1001.

     
     
     <param name = "apply-inbound-acl" value = "domains" />
     
     <param name = "auth-all-packets" value = "false" />
     <param name = "auth-calls" value = "true" />
     <param name="inbound-reg-force-matching-username" value="true"/>

     <list name = "domains" default = "deny">
       <node type = "allow" domain = "$$ {domain}" />
     </list>

     <list name = "lan" default = "allow">
       <node type = "allow" cidr = "192.168.42.42/32" />
     </list>

[reg.txt](https://github.com/signalwire/freeswitch/files/5321645/reg.txt)

alexey-khabulyak:
i think the problem is here
https://github.com/signalwire/freeswitch/blob/master/src/switch_xml.c#L444

Freeswitch matches a user with the first node if the user has a leading exclamation mark.

alexey-khabulyak:
I can't understand the meaning of this check.
The only one value with a leading "!" is a pointer. I added some logs and got this:

`2020-10-04 23:49:31.330156 [DEBUG] switch_xml.c:424 attrs name:id, value:!`
`2020-10-04 23:49:31.330156 [DEBUG] switch_xml.c:424 attrs name:number-alias, value:!`
`2020-10-04 23:49:31.330156 [DEBUG] switch_xml.c:424 attrs name:type, value:!pointer`


if vals contains leading "!", we cut it(`const char *sval = *vals + 1`)
In case of "!pointer" we get "pointer".
But in case of "!" we get "", or "1000" in case of "!1000".
Moreover. why do we use "x" as an index?  X increases for every attr. So for attr _id_ x=0, for attr _number-alias_ x=1, for attr _type_ x=2.

then it tries to compare _aname_(1000 for the first one user) and "!"
`strcasecmp("1000", "!")` returns positive integer and goes to `done`
So we get the first node(user 1000 in this case)

alexey-khabulyak:
also.
`1000 != !1000`
[...]

PR #890, opened 2020-10-05
[switch] issue #885. fix typo in strcasecmp in switch_xml.c

alexey-khabulyak:
obviously it's a type.
without an exclamation mark, it returns true for different values.
for example `user_exists id ! ${domain}` returns true
I've described this in issue 885.

### DIFF
--- a/src/switch_xml.c
+++ b/src/switch_xml.c
@@ -443,7 +443,7 @@ SWITCH_DECLARE(switch_xml_t) switch_xml_find_child_multi(switch_xml_t node, cons
 				if (aname) {
 					if (*vals[x] == '!') {
 						const char *sval = vals[x] + 1;
-						if (strcasecmp(aname, sval)) {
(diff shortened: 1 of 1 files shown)


src/switch_xml.c
```diff
@@ -443,7 +443,7 @@ SWITCH_DECLARE(switch_xml_t) switch_xml_find_child_multi(switch_xml_t node, cons
 				if (aname) {
 					if (*vals[x] == '!') {
 						const char *sval = vals[x] + 1;
-						if (strcasecmp(aname, sval)) {
+						if (!strcasecmp(aname, sval)) {
```
crienzo, line 446: vals[x] == '!' is looking for the negation here... this looks correct to me.

crienzo:
This is working as designed.  I'll accept a PR for this if you can come up with a test scenario that demonstrates a bug.

PR #898, opened 2020-10-11
Fix for the uuid_break all API. Command not only terminates currently…

figaro2015:
… played media, but also queued


src/switch_ivr_play_say.c
```diff
@@ -1600,10 +1600,18 @@ SWITCH_DECLARE(switch_status_t) switch_ivr_play_file(switch_core_session_t *sess
 			}
 
 			if ((f = switch_channel_test_flag(channel, CF_BREAK))) {
+				switch_log_printf(SWITCH_CHANNEL_SESSION_LOG(session), SWITCH_LOG_INFO, "CF_BREAK, f = %d\n", f);
```
dragos-oancea, line 1603: why would be want this to be logged with SWITCH_LOG_INFO ? DEBUG is more appropiate. maybe not even log at all.

src/switch_ivr_play_say.c
```diff
@@ -1600,10 +1600,18 @@ SWITCH_DECLARE(switch_status_t) switch_ivr_play_file(switch_core_session_t *sess
...
 				switch_channel_clear_flag(channel, CF_BREAK);
 				if (f == 2) {
 					done = 1;
 				}
+
+				if (switch_channel_test_flag(channel, CF_BREAK_ALL)) {
+					switch_log_printf(SWITCH_CHANNEL_SESSION_LOG(session), SWITCH_LOG_INFO, "CF_BREAK_ALL\n");
```
dragos-oancea, line 1610: same as above.

PR #905, opened 2020-10-13
[core] Allow other (D)TLS Curves/Groups beside P-256

traud:
Five years ago, commit 8e1b2ea enabled ECDHE for the DTLS server by hard-coding P-256. Released before that, OpenSSL 1.0.2, allows auto selection of the curve (P-256, P-384, and P-512). OpenSSL 1.1.x has this enabled on default, which adds groups/curves like X25519 and X448 automatically. This change here allows DTLS clients with a demand for Security Level 4 (192 bit) and Level 5 (256 bit) connecting to the DTLS server.

### DIFF
--- a/src/switch_rtp.c
+++ b/src/switch_rtp.c
@@ -3734,7 +3734,9 @@ SWITCH_DECLARE(switch_status_t) switch_rtp_add_dtls(switch_rtp_t *rtp_session, d
 	DH *dh;
 	switch_status_t status = SWITCH_STATUS_SUCCESS;
 #ifndef OPENSSL_NO_EC
+#if OPENSSL_VERSION_NUMBER < 0x10002000L
 	EC_KEY* ecdh;
+#endif
 #endif
 
 #ifndef HAVE_OPENSSL_DTLS_SRTP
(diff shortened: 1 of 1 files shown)


traud:
Anything blocking this request? Do I have to search/ask possible reviewers myself (via a mailing list, interactive chat, …)? Or does the change need more explanation (in commit message or in code source)?

Or is the change too complicated; another more streamlined approach would be to undo commit 8e1b2ea. That would have exactly the same effect. Only those users with older OpenSSL versions (1.0.2, 1.0.1, 1.0.0, 0.9x) would not get ECDHE then.

crienzo (approved):

mjerris:
the condition:

#if OPENSSL_VERSION_NUMBER < 0x10002000L

seems wrong... we dont even support openssl less than 1.0.2 already.  Please review the conditions you added, the logic seems incorrect.

traud:
Revisited again, and I do not see an issue in the logic. Can you elaborate more why you came to that conclusion?

- OpenSSL 1.1.x does not need anything because it is done automatically.
- OpenSSL 1.0.2 needs `SSL_CTX_set_ecdh_auto(.)` which is not available in older and deprecated in newer versions
- OpenSSL 1.0.x needs `SSL_CTX_set_tmp_ecdh(.)`

It is a bit unclear what to do with OpenSSL 0.9.8 (which has `SSL_CTX_set_tmp_ecdh(.)`) and OpenSSL 0.9.7 (which does not have `SSL_CTX_set_tmp_ecdh(.)`). However, both do not support DTLS anyway.

> we dont even support openssl less than 1.0.2 already

Where are such decisions documented?
I went for the same code ([change](https://github.com/freeswitch/sofia-sip/pull/31)) as within Sofia-SIP for symmetry reasons. Does it have the same minimum OpenSSL version already? We can go for the [code used](https://github.com/asterisk/asterisk/blob/master/res/res_rtp_asterisk.c#L1828) in Digium Asterisk instead. Or I drop the < 1.0.2 alltogether. Whatever you like. This is something, I cannot decide as an external contributor.

crienzo:
@mjerris any further comment on this?

andywolk (approved):

andywolk:
@traud Thank you for the patch!

PR #913, opened 2020-10-16
[configuration] Copy TLS ciphers from vars.xml even in Vanilla External

traud:
Previously, the configuration file vars.xml set the TLS Cipher Suites only for internal SIP profiles. External SIP profiles used the 'DEFAULT' set of the underlying OpenSSL. Now, external SIP profiles copy to the general setting of vars.xml as well. Of course, you can overwrite this anytime, for example, to use different sets for internal and external.

### DIFF
--- a/conf/vanilla/sip_profiles/external.xml
+++ b/conf/vanilla/sip_profiles/external.xml
@@ -92,5 +92,7 @@
     <param name="tls-verify-in-subjects" value=""/>
     <!-- TLS version ("sslv23" (default), "tlsv1"). NOTE: Phones may not work with TLSv1 -->
     <param name="tls-version" value="$${sip_tls_version}"/>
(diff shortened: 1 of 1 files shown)


traud:
The same must be done for the configuration file `external-ipv6.xml` because that does not copy the values from `external.xml` (which is just for IPv4). I am not sure, whether this change is accepted at all. However, even if you reject this Pull Request, this way it gets documented somehow. For example, I got totally crazy why I got `DEFAULT@SECLEVEL=2` as ciphers for my FreeSWITCH as TLS client.

traud:
Wupps. The same must be done for the the configuration file `internal-ipv6.xml` as well. This happened six years ago, in commit 3fae719 because of commit 57abca1.

Beside that, the internal and internal-ipv6 differ a lot. Mhm.

crienzo (approved):

PR #919, opened 2020-10-19
fix like memory leak

debugcommand:
the uuid duplicated do not be used.it is not necessary new duplicated.


andywolk:
Does not seem like a bug to me. It's duplicated in a channel session's pool. No need to free that.

PR #936, opened 2020-11-02
[mod_python] Fix PyObject leaks in eval_some_python function

sunlin7:
* [mod_python.c] eval_some_python: fix leaks for python objects.


sunlin7:
Hi @andywolk 
There are leaks in `mod_python.c`, according https://docs.python.org/2.7/extending/embedding.html, all the PyObjects should decrease the reference after usage. 
Please help to review the change, I tested in my local. 
Thank you.

crienzo (approved):

Issue #935, opened 2020-11-02
prosody-rate  paramter can not be set  negative  integer in mod_unimrcp


netaudio:
**Describe the bug**
prosody-rate  paramter can not be set  negative  integer in mod_unimrcp,but when use aliyun unimrcp server,the range of legal values is -500 - 500.

**To Reproduce**
play_and_detect_speech say:unimrcp:Aiqi:{Prosody-Volume=90,Prosody-Rate=-500}:123450987654321 detect:unimrcp {enable_bargin=false, start-input-timers=false, define-grammar=true, no-input-timeout=30000} builtin:speech/transcribe

crienzo:
Resolved

PR #939, opened 2020-11-03
fix issue #935 prosody-rate paramter can not be set negative integer …

netaudio:
…in mod_unimrcp

prosody-rate paramter can not be set negative integer in mod_unimrcp,but when use aliyun unimrcp server,the range of legal values is -500 - 500.


netaudio:
fix  #935  prosody-rate paramter can not be set negative integer in mod_unimrcp.
prosody-rate paramter can not be set negative integer in mod_unimrcp,but when use aliyun unimrcp server,the range of legal values is -500 - 500.

crienzo (dismissed):

crienzo (approved):

PR #951, opened 2020-11-11
Fix [switch_core_media.c]: SRTP media crypto keys are not working

ajay-sabat:
Comparing the user given cypto type string to both name and alias of the default crypto suites supported by the system.

### DIFF
--- a/src/switch_core_media.c
+++ b/src/switch_core_media.c
@@ -1761,7 +1761,7 @@ static void switch_core_session_parse_crypto_prefs(switch_core_session_t *sessio
 			int ok = 0;
 
 			for (j = 0; j < CRYPTO_INVALID; j++) {
-				if (!strcasecmp(fields[i], SUITES[j].name)) {
+				if (!strcasecmp(fields[i], SUITES[j].name) || !strcasecmp(fields[i], SUITES[j].alias)) {
 					smh->crypto_suite_order[k++] = SUITES[j].type;
 					ok++;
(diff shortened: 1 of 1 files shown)


andywolk (approved):

src/switch_core_media.c
```diff
@@ -1761,7 +1761,7 @@ static void switch_core_session_parse_crypto_prefs(switch_core_session_t *sessio
 			int ok = 0;
 
 			for (j = 0; j < CRYPTO_INVALID; j++) {
-				if (!strcasecmp(fields[i], SUITES[j].name)) {
+				if (!strcasecmp(fields[i], SUITES[j].name) || !strcasecmp(fields[i], SUITES[j].alias)) {
```
andywolk, line 1764: Shouldn't we use `switch_core_media_crypto_str2type()` here instead of the whole block?

andywolk (changes requested):

Five9Voice:
This page defines cyptro suites: https://freeswitch.org/confluence/display/FREESWITCH/rtp_secure_media.  Some of the encryption suite entries are aliases, some are not aliases.

For a TLS/SRTP call, if the user specifies a crypto suite that is an alias from the list mentioned in the document negotiation will fail.  This patch allows the user to specify either an encryption suite name OR and alias.

andywolk:
@Five9Voice Please resolve the requested change.

PR #953, opened 2020-11-17
[Core] Fix XML generation

NormB:
Correct multiple instances where the XML node offsets were set improperly.


crienzo (approved):

crienzo:
@andywolk I talked to norm about this- looks good to me and fixes his issue.

andywolk:
@NormB Thank you for the fix.

Issue #904, opened 2020-10-13
No need "precondition" in Supported:


nevian427:
Hi,

FreeSWITCH adds precondition option with 100rel enable. We don't need it to work with 100rel.
There is an old bug FS-6221 where precondition moved from mandatory to optional part but correct way to clean it out.
FS doesn't support preconditions at all as per sources.

1. Vanila config
2. Set profile  <param name="enable-100rel" value="true"/>
3. Register two users or simple gw
4. Make call between users or via gw

INVITE sip:+79641234567@176.xx.xx.3 SIP/2.0
Via: SIP/2.0/UDP 83.xx.xx.142;rport;branch=z9hG4bK2j3X353U8QQKc
Max-Forwards: 67
From: <sip:+79011234567@176.xx.xx.3>;tag=6H8K6vBXtH8UD
To: <sip:+79641234567@176.xx.xx.3>
Call-ID: 34f87d71-16d5-4eed-9523-0fa2fb217b73
CSeq: 26739475 INVITE
Contact: <sip:gw+tele2_79011234567@83.xx.xx.142:5060;transport=udp;gw=tele2_79011234567>
User-Agent: ES FS SBC
Allow: INVITE, ACK, BYE, CANCEL, OPTIONS, MESSAGE, INFO, UPDATE, PRACK, NOTIFY
Supported: precondition, 100rel, timer, path, replaces
Allow-Events: talk, hold, conference, refer
Privacy: none
Content-Type: application/sdp
Content-Disposition: session
Content-Length: 216
P-Asserted-Identity: <sip:+79011234567@176.xx.xx.3>

v=0
o=nobody 1602566261 1602566262 IN IP4 83.xx.xx.142
s=nobody
c=IN IP4 83.xx.xx.142
t=0 0
m=audio 20258 RTP/AVP 0 101
a=rtpmap:0 PCMU/8000
a=rtpmap:101 telephone-event/8000
a=fmtp:101 0-16
a=ptime:20

**Expected behavior**
Supported header w/o precondition option i.e.

Supported: 100rel, timer, path, replaces

**Package version or git hash**
commit 1010c6c55c35b9df738c67bc80cb353c873dab76

andywolk:
Fixed by https://github.com/signalwire/freeswitch/pull/956

PR #956, opened 2020-11-24
[mod_sofia] Remove 'precondition' from 'Supported:' w/ 100rel

trdenton:
This PR removes the "precondition" option from Supported: header when 100rel mode is enabled, addressing #904 

comparing RFC3312 and the freeswitch codebase, I see no support for the mechanisms that "precondition" is intended to support (eg. `a=des, a=curr`), in my opinion it makes sense to remove it until such a time that rfc3312 is supported

In more concrete terms - freeswitch (or it's included modules) do not maintain a precondition table which is a "MUST" for  RFC3312 "precondition" functionality (there are probably other unsatisfied MUSTs, that's just the first one I've found: https://tools.ietf.org/html/rfc3312#section-5.1)

### DIFF
--- a/src/mod/endpoints/mod_sofia/sofia.c
+++ b/src/mod/endpoints/mod_sofia/sofia.c
@@ -3165,7 +3165,7 @@ void *SWITCH_THREAD_FUNC sofia_profile_thread_run(switch_thread_t *thread, void 
 		goto end;
 	}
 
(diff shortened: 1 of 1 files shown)


trdenton:
Has been tested on a local install to ensure it meets the 'expected behaviour' listed in #904

mjerris:
I had thought this was handled down in the sofia-sip stack itself, not in freeswitch or mod_sofia.  You'd need to confirm thats correct or not.

crienzo (approved):
@andywolk the commit looks fine- just massage the commit message to follow the standard format.  Thanks

PR #967, opened 2020-12-03
switch_ivr_park_session will now ignore a park state change if the ch…

djgroves:
…annel is already in a CS_PARK state.


djgroves:
When a channel is in CS_PARK, it's executing ivr_park(). If a new park request comes in (from whatever source), the flag to indicate a state change is set, which results in it returning. The proposed fix is to ignore state changes to CS_PARK if we're already in CS_PARK, and not generate a state change from CS_PARK to CS_PARK.

crienzo:
@mjerris is this OK?

djgroves:
Closing due to no longer being valid against current version

PR #969, opened 2020-12-04
mod_curl: Added PATCH, Verify SSL certs with option insecure, Added proxy option

ar45:
- Handle `PATCH` http request
- Verify SSL Certs for https requests with an option `insecure` to not verify (**It's 2020!**)
- Added option `proxy` to proxy requests, I use this for security


snaptel (approved):

ar45:
@andywolk mind looking at this?

andywolk (approved):

andywolk (changes requested):
Let's `assert(options)` instead of `if (options) {` as we never call this func without options.

ar45:
@andywolk done.

andywolk:
Well. Here is the thing. We have not had secure before so insecure was our default. I don't say it's elegant, but we should keep default as insecure for backward compatibility or it is going to break those who updates to this version but using self-signed certificates.

andywolk:
Is that by a reason or there is a missing `NOT`?
```
curl_options_t options = { .insecure = globals.validate_certs };
```

ar45:
> Is that by a reason or there is a missing `NOT`?
> 
> ```
> curl_options_t options = { .insecure = globals.validate_certs };
> ```

Chang, yep. Updated.

ar45:
@andywolk mind having a look at this?

andywolk:
@ar45 Thinking. As we have configurable default now ... we should add `secure` option as well. So those who is insecure by default will be able to make secure requests.

ar45:
At the moment I can't do this. If anyone wants to add that later we can do that. Can we please merge it as is?

andywolk (approved):

ar45:
@andywolk @mjerris Can we get this merged?

mjerris (approved):

ar45:
👏

PR #982, opened 2020-12-11
[mod_opus] Do not hangup call on decode error - fix 86a5ee3509

italorossi:
Had this issue back in October with a specific Chrome version. When there's packet loss from the browser (tested with 10% using Linux netem (tc qdisc add dev ethx root netem loss 10%) and VMware fusion) mod_opus fails to decode the samples and return SWITCH_STATUS_FALSE and this makes switch_core_media hangup the call, changing to SWITCH_STATUS_NOOP fixed the issue and there's no, apparently, audio issues.

With latest chrome stable and canary I couldn't reproduce the issue anymore.

### DIFF
--- a/src/mod/codecs/mod_opus/mod_opus.c
+++ b/src/mod/codecs/mod_opus/mod_opus.c
@@ -914,7 +914,7 @@ static switch_status_t switch_opus_decode(switch_codec_t *codec,
 	if (samples < 0) {
 		switch_log_printf(SWITCH_CHANNEL_LOG, SWITCH_LOG_ERROR, "Decoder Error: %s fs:%u plc:%s!\n",
 						  opus_strerror(samples), frame_size, plc ? "true" : "false");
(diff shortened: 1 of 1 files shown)


dragos-oancea (approved):

andywolk:
@italorossi This PR needs to be reverted. We should not use the NOOP status in the case of an error.
```
SWITCH_STATUS_NOOP, NOOP indicates that the audio in is already the same as the audio out, so no conversion was necessary.
```
Do you remember any details about this problem? Because we are going to get back to returning FALSE as it was before.

a-rose:
Before this was merged, we used to use SWITCH_STATUS_BREAK to avoid hanging up, and did not notice issues.

italorossi:
@andywolk calls were dropping when chrome side had packet loss (>10%), you can reproduce it with netem on client side (chrome).

PR #988, opened 2020-12-15
[build] fix libpq configure check to use flags from autoconf/pkg-config

mjerris:


andywolk (approved):

MarioG-X:
Just an FYI, this mornings master does not contain this fix. I don't know how to tell when a fix actually goes into master on GitHub so I have to keep downloading master. Is there an easy way to know when this goes into master. Thank you.

PR #995, opened 2020-12-16
[mod_sms] add final chat function

seven1240:


crienzo (approved):

src/mod/applications/mod_sms/mod_sms.c
```diff
@@ -514,6 +514,13 @@ SWITCH_STANDARD_CHAT_APP(stop_function)
 	return SWITCH_STATUS_FALSE;
 }
 
+SWITCH_STANDARD_CHAT_APP(final_function)
+{
+	switch_event_add_header(message, SWITCH_STACK_BOTTOM, "final_delivery", "true");
+	switch_set_flag(message, EF_NO_CHAT_EXEC);
+	return SWITCH_STATUS_FALSE;
```
andywolk, line 521: This looks wrong and the one above looks wrong too. Shouldn't them return success?

andywolk (changes requested):

src/mod/applications/mod_sms/mod_sms.c
crienzo: returning false stops processing in mod_sms.  I think this is correct.

PR #1000, opened 2020-12-17
[mod_http_cache] Set binary mode when downloading files in Windows

crienzo:


andywolk (approved):

hheedfeld:
I did compile the patch and tested it - unfortunately the problem is still there

PR #1016, opened 2020-12-30
Update subagent.c

melekoktay:
Snmp module couldn't compile at Debian 10 compiler "gcc version 8.3.0" . After resizing appropriate buffers , bug has been resolved


crienzo:
@andywolk you have a similar patch already, right?

andywolk:
I think this is fixed. Please rebase so we could see if there is anything left.

PR #1029, opened 2021-01-11
[mod_sofia] Set the proper content-length for multipart messages

winsock:
# Context
This change ensures that the proper content length is set when setting a multipart header with custom headers.

# Questions
- Is Content-Length even required or needed in the multipart data reading this email chain it seems like it may be unnecessary and unused but allowed since it isn't explicitly disallowed. 
  - https://lists.cs.columbia.edu/pipermail/sip-implementors/2012-March/028216.html

# Overview
- Use strstr to find the boundary of the body `\r\n\r\n` ignore if there is no valid body
- Use the pointer found by strstr + 4(accounting for the boundary) to call strlen with to get the real size of the body. But add one to match existing behavior.


winsock:
Here is a packet capture of how this looks before and after:
[before_and_after_pcap.zip](https://github.com/signalwire/freeswitch/files/5798826/before_and_after_pcap.zip)

Issue #839, opened 2020-09-01
Output the SIP call-id with the FreeSWITCH logs


khou:
I'm trying to correlate FreeSWITCH application logs with SIP I'm capturing using Heplify.

If possible, an enhancement to logging where the call-id is logged with each line of the application log would make it much easier to identify what occurs in FreeSWITCH for a specific call.

I've attempted this with channel UUID and capturing the CDR, then correlating the SIP with the CDR, but as you can see, this is an extensive workaround and requires lots of custom logic.

Desired output sample:
```
2020-09-01T20:33:58Z INFO HEP packet:{Version:2,Protocol:17,SrcIP:192.168.63.171,DstIP:192.168.63.6,SrcPort:15060,DstPort:41963,Tsec:1598992438,Tmsec:270403,ProtoType:1,NodeID:2002,NodePW:,CID:,Vlan:0} with Payload:
SIP/2.0 200 OK
Via: SIP/2.0/UDP 192.168.63.6:41963;rport=41963;branch=z9hG4bKPj2911369f-24eb-4731-b915-7cde4dbfa3b9
From: <sip:1018@freeswitch-internal>;tag=cba407e1-3d14-497f-97b2-c0efae925577
To: <sip:1019@freeswitch-internal>;tag=v2K9tQpaUrQKj
Call-ID: dedf359c-fd65-4df4-a512-1b05747a1b6d
CSeq: 9211 BYE
User-Agent: FreeSWITCH-mod_sofia/1.10.2-release-14-f7bdd3845a~64bit
Allow: INVITE, ACK, BYE, CANCEL, OPTIONS, MESSAGE, INFO, UPDATE, REGISTER, REFER, NOTIFY, PUBLISH, SUBSCRIBE
Supported: timer, path, replaces
Content-Length: 0
Session-ID: 00000000000000000000000000000000

dedf359c-fd65-4df4-a512-1b05747a1b6d 2020-09-01 20:33:58.252944 [INFO] sofia_glue.c:545 Setting "Session-ID: 058092b0134449418b09e2df389e06a5" from partner leg
dedf359c-fd65-4df4-a512-1b05747a1b6d 2020-09-01 20:33:58.252944 [WARNING] sofia_glue.c:588 Session-ID: Fallback to RFC7329
[...]

PR #1034, opened 2021-01-19
[mod_sofia] add call-id to log line to link with session uuid

seven1240:
so it's easy to relate session id and sip call-id


seven1240:
for https://github.com/signalwire/freeswitch/issues/839

crienzo (approved):

crienzo:
@andywolk this PR is good

PR #1036, opened 2021-01-20
fix the color rendering in the chrome 84+

beibeiyhj:
fix the color rendering in the chrome 84+ that sRGB not support


crienzo:
@seven1240 is this OK?

PR #1045, opened 2021-01-23
[core] allow redirection of com modules

lazedo:
checks redirect-mod-load-to-com-module core variable
and redirects the load to com module if it exists.
ex: load mod_g729 will load mod_com_g729 if its available.

this patch adds

* switch_loadable_module_available_in_dir
  checks if module is available in directory
* switch_loadable_module_available
  calls switch_loadable_module_available_in_dir with default module dir

this patch changes
* switch_loadable_module_enumerate_available
  use created pool and freeing after enumeration for switch_dir operations.
  this avoids the switch_dir pool allocations (loadable_modules.pool) to stay in memory
  allowing freeswith to run for long periods without the need to restart
  because of available memory


crienzo:
I think you could accomplish this with a little preprocessor magic using `exec-set`?

PR #1048, opened 2021-01-25
[mod_http_cache] Generic HTTP API Authorization header support

tmnsur:
Current mod_http_cache implementation does not support HTTP authorization headers for generic (non aws/azure) HTTP APIs.

This patch provides a simple way to add a custom (like "API_KEY") or standard ("Authorization") HTTP API authorization header.


crienzo (changes requested):
Fix tab/spaces indentation mismatches and change the default profile configuration so it can accept multiple key/value pairs.

tmnsur:
I fixed the tab/spaces indentation mismatches and default profile configuration can now accept multiple key/value pairs.

I am not sure about memory allocation related parts though, are they going to be automatically freed when mod_http_cache is unloaded?

tmnsur (commented):
OK

crienzo:
Looks fine- yeah, if allocated off the module's pool it will get freed if the module is unloaded.

crienzo (approved):
@andywolk this should be OK.  Just need the commits squashed before merge.

PR #1050, opened 2021-01-26
[mod_callcenter] fix leak

seven1240:


crienzo (dismissed):

crienzo:
@andywolk this is OK

andywolk (changes requested):
Does not look right to me.
@seven1240 Please take a look at lines 
https://github.com/signalwire/freeswitch/blob/795f41e40c9348d298f4f33467e70ba77dbc7572/src/mod/applications/mod_callcenter/mod_callcenter.c#L1549
and https://github.com/signalwire/freeswitch/blob/795f41e40c9348d298f4f33467e70ba77dbc7572/src/mod/applications/mod_callcenter/mod_callcenter.c#L4344

andywolk:
Btw zstr() does not warn if a pointer is an empty string "" (not NULL) so switch_safe_free() may be reasonable here (unless allocated in a pool):
```
	if (zstr(globals.cc_instance_id)) {
		switch_safe_free(globals.cc_instance_id); // <----
		globals.cc_instance_id = strdup("single_box");
```

crienzo:
I'll fix it

crienzo (approved):

PR #1051, opened 2021-01-26
[mod_event_socket] Make ip connected which matched one of acls

azzbcc:


azzbcc:
found it duplicate in https://github.com/signalwire/freeswitch/blob/bc8e3da74c9a8f0ead6a3e6fcbdf396348b2a513/src/switch_core.c#L1391-L1413

close it.

PR #1054, opened 2021-01-28
[mod_sofia] fix empty profile name check

azzbcc:
Fix in which condition is always false.
Fix leak when nua start failed.
Add a test suite for sofia profile.


crienzo (changes requested):
Test needs to pass address sanitizer checks.

crienzo:
@andywolk is this OK once the conflicts get resolved?

andywolk:
@azzbcc Ping me when you have everything resolved.

Issue #1011, opened 2020-12-25
FreeSWITCH conference AGC doesn't work


figaro2015:
I wrote this bug against the conference AGC handling, but I think it's applicable to all other audio media processing applications. If I play a constant 'loud' tone (in my it's 440hz tone) into the conference I am getting an initial reduction in volume based on the AGC level setting and then constant tone in the output. 
Instead, if I am running my test with conference default values (whatever those values are), meaning before issuing any AGC setting API, I see the initial reduction in volume (which looks correct to me), followed by constant 8-second cycles of another reduction in volume (approximately -8db) and then going back to the previous level. 
If I am running the same test after I assume conference AGC setting API - I do get constant output, which level is EXACTLY the same as the input tone. In fact, they are identical. This means no AGC processing takes place at all.
Just in case my explanation seems confusing, I have recorded .wav files and screenshots of Audacity plots.

My setup is the following:
1. I have Windows PC configured to use 'Stereo Mix' as audio input.
2. I am dialing into conference with the Linphone. Linphone is set up to have Stereo Mix as input and I am using Windows Media Player to play the constant tone into the conference. Important to note that 'Enable echo cancellation' on Linphone has to be unchecked. Otherwise, Linphone does it's own AGC processing.
3. I am recording the conference into an audio .wav file for further analysis.

The expected behavior would be an initial reduction in volume based on the AGC setting followed by the constant tone.

The version of the FreeSWITCH I have been testing this with is the top of the master branch:
FreeSWITCH Version 1.10.6-dev+git~20201223T174526Z~62970aa998~64bit (git 62970aa 2020-12-23 17:45:26Z 64bit)

I built FreeSWICH from sourse:
commit 62970aa9987683a802a413393b93b2f687499272 (HEAD -> master, origin/master, origin/HEAD)

I am going to attach recorded .wav files and screenshots of Audacity plots:
1. The initial settings of the conference are: 
energy_level = 50
AGC = 500

2 API has been used to set AGC level to 1100.

![50_550](https://user-images.githubusercontent.com/11821347/103142258-25ac3600-46b5-11eb-8b0d-b8822ad0f071.png)

![50_1100_api](https://user-images.githubusercontent.com/11821347/103142261-2e047100-46b5-11eb-9c0d-fbe0b6ede5ae.png)

Please let me know if there are any other artifacts I can provide. Github won't allow upload of .WAV files, therefore I had to convert them to .MP4

[...]

PR #1064, opened 2021-02-04
[Core, mod_conference] Fixed oscillation in conference AGC processing. Especially during volume reduction processing

figaro2015:
https://github.com/signalwire/freeswitch/issues/1011


crienzo (changes requested):
Can this be fixed without changing switch_normalize_volume_granular()?

src/mod/applications/mod_conference/conference_api.c
```diff
@@ -1276,15 +1276,6 @@ void conference_api_set_agc(conference_member_t *member, const char *data)
...
 
-	if (argv[0]) {
-		tmp = atoi(argv[0]);
-
-		if (tmp > 0) {
-			member->agc_period_len = (1000 / member->conference->interval) * tmp;
-		}
-	}
-
-
```
crienzo, line 1287: This should be if (argv[3]) {

src/mod/applications/mod_conference/conference_api.c
```diff
@@ -1276,15 +1276,6 @@ void conference_api_set_agc(conference_member_t *member, const char *data)
 	}
 
 
-	if (argv[0]) {
```
crienzo, line 1279: if (argv[3]) {

src/mod/applications/mod_conference/conference_api.c
```diff
@@ -1276,15 +1276,6 @@ void conference_api_set_agc(conference_member_t *member, const char *data)
...
-		tmp = atoi(argv[0]);
```
crienzo, line 1280: tmp = atoi(argv[3]);

figaro2015:
Actually, my main change is not in the switch_normalize_volume_granular(), but in switch_change_sln_volume_granular().
The main reason for the oscillation is the fact that the step in the array of coefficients for the negative adjustment is too wide. In other words, it is not 'granular' enough. Which causes, on every run of the algorithm, to force the adjustment either to increase or decrease the volume. But the actual AGC processing algorithm is correct is working fine.
Also, I had to increase the margin value to 100. 
These adjustments seem to produce acceptable results in AGC processing. To accommodate even more fine margin (20, etc) the array of negative coefficients will have to be increased even further, with the smaller step between the elements.

crienzo:
How about calculating amplitude gain from dB like:
```
#define dB_gain_factor(gain_dB) (powf(10.0f, (float)gain_dB / 20.0f))
```
Then we get 2x perceived volume increase for every +10 increase in gain_dB.  Then you can be as granular as you want to be.

crienzo:
@figaro2015 I rescaled the tables for the granular volume.  Looks like they are still pretty close- can you verify this works for you?  Thanks.

figaro2015:
Chris, I believe the coefficients you generated are working fine. Please see the screenshot of the Audacity plot attached.

Latest_AGC

crienzo:
@andywolk I think we can take this now

crienzo (approved):

andywolk (approved):

MarioG-X:
I think it's possible that this change broke building make now fails. I was able to build on March 24, but not on April 3. Using git log the only change between the two is this issue. Please see issue #1145.

git log:
commit ec1a0215fedb56bcd6f73fb1e93f763ed5286cbb (HEAD -> master, origin/master, origin/HEAD)
Author: figaro2015 <figaro2015@users.noreply.github.com>
Date:   Fri Apr 2 12:18:16 2021 -0700

    [Core, mod_conference] Fixed oscillation in conference AGC processing. Especially during volume reduction processing ( 1 0 6 4)
    * Fixed oscillation in conference AGC processing. Especially during volume reduction processing    
    * Re-scaled granular volume to +/- 50 dB level change. 10 = double loudness, -10 = half loudness, 0 = no change
    Co-authored-by: Chris Rienzo <chris@signalwire.com>
commit 67cec5c3e80e653d4540b9ba2c4a0b38b92a9458
Author: Andrey Volk <andywolk@gmail.com>
Date:   Thu Mar 25 18:52:56 2021 +0300
    version bump

Issue #1033, opened 2021-01-16
FreeSWITCH displays misleading ERROR message if SOFIA profile could not be created due to bad certificate


figaro2015:
I was debating for some time whether to submit this as a bug report or a new feature request. Eventually decided to go with Feature Request.

FreeSWITCH displays misleading ERROR message is SOFIA profile could not be created due to bad/corrupted PEM certificate file:

2021-01-16 13:07:52.752287 [ERR] sofia.c:3260 Error Creating SIP UA for profile: webrtc (sip:mod_sofia@192.168.86.37:0;transport=udp,tcp) ATTEMPT 2 (RETRY IN 5 SEC)
2021-01-16 13:07:57.752289 [ERR] sofia.c:3260 Error Creating SIP UA for profile: webrtc (sip:mod_sofia@192.168.86.37:0;transport=udp,tcp) ATTEMPT 3 (RETRY IN 5 SEC)
2021-01-16 13:07:57.752289 [ERR] sofia.c:3270 Error Creating SIP UA for profile: webrtc (sip:mod_sofia@192.168.86.37:0;transport=udp,tcp)
The likely causes for this are:
1) Another application is already listening on the specified address.
2) The IP the profile is attempting to bind to is not local to this system.

The way to re-create the problem is simple: create bad/corrupted wss.pem file in ../certs directory and either try to start the FreeSWITCH with a profile that uses the certificate (for example profile for WebRTC calls) or unload and then load mod_sofia module. Once again, the problem exhibits itself only if there is a profile that uses the certificate.

The reason for this problem is obvious: FreeSWITCH doesn't get any feedback from the NUA protocol stack, other than the fact that it returns NULL of failure to create the stack.

This problem exists in all releases and on the top of the master branch:

FreeSWITCH Version 1.10.6-dev+git~20210116T161209Z~d498f6c723~64bit (git d498f6c 2021-01-16 16:12:09Z 64bit).

Needless to say, this could lead to hours of frustration while debugging the issue. Without any result, unless you start putting custom debug statements in the actual SOFIA stack code.

The proposed solution is to perform verification of the wss.pem certificate in the mod_sofia application code, similar to how it is does in the NUA stack, **before** calling the API to create the stack (**nua_create**).  And fail the profile with the appropriate error message if certification verification failed.

I have the solution implemented, and if the proposed solution above is acceptable - I am ready to submit the pull request.

PR #1067, opened 2021-02-05
Verification of the wss.pem certificate in mod_sofia application code. Display correct reason why sofia profile failed to create

figaro2015:
https://github.com/signalwire/freeswitch/issues/1033

### DIFF
--- a/src/mod/endpoints/mod_sofia/sofia.c
+++ b/src/mod/endpoints/mod_sofia/sofia.c
@@ -3149,11 +3149,14 @@ void *SWITCH_THREAD_FUNC sofia_profile_thread_run(switch_thread_t *thread, void 
 	switch_status_t st;
 	char qname [128] = "";
 
+#if defined(HAVE_OPENSSL)
 	char *key  = switch_core_sprintf(profile->pool, "%s/%s", profile->tls_cert_dir, "wss.pem");
 	char *cert  = switch_core_sprintf(profile->pool, "%s/%s", profile->tls_cert_dir, "wss.pem");
 	char *chain  = switch_core_sprintf(profile->pool, "%s/%s", profile->tls_cert_dir, "wss.pem");
 	SSL_CTX *ssl_ctx;
 	const SSL_METHOD *ssl_method = SSLv23_server_method();
+#endif
+
 	switch_bool_t ssl_error = SWITCH_FALSE;
 
 	switch_mutex_lock(mod_sofia_globals.mutex);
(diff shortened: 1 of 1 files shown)


src/mod/endpoints/mod_sofia/sofia.c
```diff
@@ -3185,6 +3192,37 @@ void *SWITCH_THREAD_FUNC sofia_profile_thread_run(switch_thread_t *thread, void
 		profile->tls_verify_in_subjects = su_strlst_dup_split((su_home_t *)profile->nua, profile->tls_verify_in_subjects_str, "|");
 	}
 
+	ssl_ctx = SSL_CTX_new((SSL_METHOD *)ssl_method);
```
crienzo, line 3195: This must be destroyed when finished.  Any use of SSL must check if `HAVE_OPENSSL` is defined and the correct version of openssl is available.

crienzo (changes requested):

src/mod/endpoints/mod_sofia/sofia.c
mjerris: also can we just make this a core helper function that does all what chris said properly and returns a result and use that function here

mjerris (changes requested):

figaro2015:
Based on the comments provided I added a check for HAVE_OPENSSL pre-processor directive define. And the release of previously allocated SSL_CTX object.
I don't believe any additional OpenSSL version check is required as 'SSLv23_server_method', according to the OpenSSL documentation, is version flexible and support all currently available version of SSL/TLS protocols.

"SSLv23_method(), SSLv23_server_method(), SSLv23_client_method()
These are the general-purpose version-flexible SSL/TLS methods. The actual protocol version used will be negotiated to the highest version mutually supported by the client and the server. The supported protocols are SSLv2, SSLv3, TLSv1, TLSv1.1 and TLSv1.2. Most applications should use these method, and avoid the version specific methods described below"

Please take a look at certificate validation in MOD_VERTO. It's pretty similar to what I am doing in MOD_SOFIA.

As far as creating a separate core library helper utility - there are at least 4 or 5 places throughout FreeSWITCH modules where similar certificates validation/verification performs. Therefore, creating such a centralized utility function, to really serve the purpose, would involve modifications in all those places.

src/mod/endpoints/mod_sofia/sofia.c
```diff
@@ -3267,9 +3313,16 @@ void *SWITCH_THREAD_FUNC sofia_profile_thread_run(switch_thread_t *thread, void
...
 	if (!profile->nua) {
-		switch_log_printf(SWITCH_CHANNEL_LOG, SWITCH_LOG_ERROR, "Error Creating SIP UA for profile: %s (%s)\n"
-						  "The likely causes for this are:\n" "1) Another application is already listening on the specified address.\n"
-						  "2) The IP the profile is attempting to bind to is not local to this system.\n", profile->name, profile->bindurl);
+		if (!ssl_error) {
+			switch_log_printf(SWITCH_CHANNEL_LOG, SWITCH_LOG_ERROR, "Error Creating SIP UA for profile: %s (%s)\n"
+							  "The likely causes for this are:\n" "1) Another application is already listening on the specified address.\n"
+							  "2) The IP the profile is attempting to bind to is not local to this system.\n", profile->name, profile->bindurl);
+		}
+		else {
```
crienzo, line 3321: } else {

crienzo:
Just a minor nit for me.  @mjerris is this OK otherwise?

src/mod/endpoints/mod_sofia/sofia.c
figaro2015: Corrected if/else curly braces formatting. Checked in.

crienzo (approved):

PR #1069, opened 2021-02-11
freeswitch.spec: Module kazoo renamed to event-kazoo

sergey-safarov:


sergey-safarov:
Could you look at this PR?
@andywolk 
@lazedo

PR #1074, opened 2021-02-11
[core] Do not block waiting for peer_channel after the bridge ended with CF_ATTENDED_TRANSFER

italorossi:
[core] Do not block waiting for peer_channel after the bridge ended due to CF_ATTENDED_TRANSFER

In some attended transfer scenarios the BYE was only being sent after
the call between peer_channel and his new party is finished. ref
https://freeswitch.org/jira/browse/FS-7242

This was happening because between the audio_bridge_thread finishing and
the while loop checking for peer_channel state, the peer_channel got
quickly bridged to another channel, keeping his state in CS_EXCHANGE_MEDIA
and holding the original channel in the CS_EXECUTE state. Since this
loop doesn't finish the session couldn't transition to CS_PARK to be
hung up with park_timeout (ATTENDED_TRANSFER) ref: mod_sofia
sofia_handle_sip_i_refer.

### DIFF
--- a/src/switch_ivr_bridge.c
+++ b/src/switch_ivr_bridge.c
@@ -1800,7 +1800,7 @@ SWITCH_DECLARE(switch_status_t) switch_ivr_multi_threaded_bridge(switch_core_ses
 			switch_channel_stop_broadcast(peer_channel);
 
 
-			while (switch_channel_get_state(peer_channel) == CS_EXCHANGE_MEDIA) {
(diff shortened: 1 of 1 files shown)


mjerris (changes requested):
I think what we really need here is to check if peer_channel has channel as its bridged partner.  The issue is this same race could theoretically happen from something like a uuid bridge to something else that happens right as the bridge is setting up,and have it set up before it gets to this code.  The problem is we need that to be a very inexpensive check as this will get called a bunch, do we have what we need to already have those objects without additional locking, and can do a binary uuid compare on the 2 uuid?

fetristan:
Hi,
The problem is always up to date, any news ?

PR #1077, opened 2021-02-15
[mod_sofia] add support to override Call-Info in INVITE

cauca-signalwire:


sergey-safarov:
We use Call-Info headers.
This is the reason why looked at this PR.

Could you describe the use-case and issue that solves this PR?

PR #1079, opened 2021-02-15
[mod_sofia] add a new uuid_refer_3pcc command and refer_3pcc in the dialplan

cauca-signalwire:


cauca-signalwire:
There's an issue in the BYE that is being sent, you can hold the review until I push again.

PR #1099, opened 2021-03-05
[mod_tts_commandline] check return of switch_system() against 127 too

dragos-oancea:


dragos-oancea:
this fixes the situation when "text2wave" utility is missing ("festival" package not installed) .  we should stop the dialplan execution in this case. 
eg: "2021-02-26 15:19:24.713040 [DEBUG] mod_tts_commandline.c:160 Executing:
echo 'This is an example of using tts_commandline' | text2wave -f 8000 >
'/tmp/0665a30d-2bb8-40fe-b5a6-9625bd58306e.tmp.wav'"

https://lists.freeswitch.org/pipermail/freeswitch-users/2021-March/134569.html

crienzo (approved):

Issue #1097, opened 2021-03-04
Coding error in switch_dso_open()


Paul-Mateer:
if the first call to LoadLibraryEx fails in switch_dso_open, it makes a second call with amended parameters but does not capture the return code. As a result this second call may succeed but the code will still report an error and return a  NULL value to the caller.

andywolk:
Patches are always welcome :-) @Paul-Mateer

Paul-Mateer:
Happy to oblige @andywolk . Bit of a git/github novice so do I need to clone the active branch (apr-osx-configure-ipv6) and submit the change there?

andywolk:
Press fork, create a branch in your fork, make a change, commit with a message, click to make a PR.

Paul-Mateer:
Thanks @andywolk. So I think that I have successfully forked, branched, modified the file and committed it using GitHub for Desktop. Not sure what I am supposed to click to make a PR however. Getting my head around Git after years of working with TFS is giving me a headache.

andywolk:
Fixed by https://github.com/signalwire/freeswitch/pull/1100

PR #1100, opened 2021-03-05
[Core] Capture return value of 2nd call to LoadLibraryEx and add a third call to widen the search to the module in switch_dso_open().

Paul-Mateer:
Modified to capture return value of 2nd call to LoadLibraryEx and added a third call to widen the search to the module.


Paul-Mateer:
#1097

andywolk (approved):

PR #1102, opened 2021-03-09
Fix Memory leak in ESL lib.

zooptwopointone:


libs/esl/src/esl.c
```diff
@@ -1392,6 +1392,7 @@ ESL_DECLARE(esl_status_t) esl_recv_event(esl_handle_t *handle, int check_q, esl_
 		if (!esl_safe_strcasecmp(hval, "text/disconnect-notice") && revent->body) {
 			const char *dval = esl_event_get_header(revent, "content-disposition");
 			if (esl_strlen_zero(dval) || strcasecmp(dval, "linger")) {
+				free(revent-body);
```
mjerris, line 1395: do you mean free(revent->body) ?

mjerris (changes requested):
the patch as is is incorrect.  Please correct your syntax issue and actually test the patch before updating the pull request.

zooptwopointone:
Oops yea look from the bug I fixed in opensips so didn’t pay too much attention here.  Thanks.

libs/esl/src/esl.c
zooptwopointone: Ok fixed.

zooptwopointone (commented):
Fixed free statement as requested

andywolk:
Please rebase to the latest master

PR #1103, opened 2021-03-10
[mod_amr] introduce new cfg setting force-oa to change between OA/BE when originating, show session in debug logs, update configs

dragos-oancea:
Sync with the amr-wb changes, so the modules would be easier to maintain.


crienzo:
@andywolk this is OK

PR #1104, opened 2021-03-10
[core,mod_conference] Fix memory leak on clearing down member call by changing way of flushing frame buffer

jakubkarolczyk:


jakubkarolczyk:
Would it be possible to merge this fix?

PR #1105, opened 2021-03-10
[core] Fix memory leak caused by persistence of AVPacket side_data

jakubkarolczyk:
When encoded frames cannot be written to session, the loop breaks and the remaining data cannot be processed in consume_h264_bitstream function.
This lead to situation when AVPacket is not wiped out and its side_data are not released from memory causing the leak.

With this fix we are not breaking the loop when we cannot write the encoded frame to session, giving the chance to process the whole frame and in consequence to wipe out the AVPacket and free side_data.

### DIFF
--- a/src/switch_core_media.c
+++ b/src/switch_core_media.c
@@ -14721,7 +14721,7 @@ SWITCH_DECLARE(switch_status_t) switch_core_session_write_video_frame(switch_cor
 			status = switch_core_session_write_encoded_video_frame(session, frame, flags, stream_id);
 		}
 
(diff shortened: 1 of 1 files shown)


jakubkarolczyk:
Would it be possible to merge this fix?

seven1240:
can test the status before write? e.g.
```
status = SWITCH_STATUS_SUCCESS

while() {
    if (status == SWITCH_STATUS_SUCCESS) switch_core_session_write( ....)
}
```

also, is there a better way to wipe the buffer in `codec_close` instead of tweak the core? since mod should manage it's buffer properly.

Issue #1107, opened 2021-03-11
Bug in switch_xml_parse_str_dynamic() 


pregalla:
**Issue**: **Double free Situation**

In switch_xml_parse_str_dynamic(_In_z_ char *s, _In_ switch_bool_t dup), 

when dup is SWITCH_TRUE, data is duped. data = dup ? switch_must_strdup(s) : s;

But irrespective of the value of 'dup', **root->dynamic = 1;**

if ((root = (switch_xml_root_t) switch_xml_parse_str(data, strlen(data)))) {
		**root->dynamic = 1;**		/* Make sure we free the memory is switch_xml_free() */
		return &root->xml;
	}

In switch_xml_parse_str(char *s, switch_size_t len)
     **root->m = s;**

And in switch_xml_free():
		if (root->dynamic == 1)
			**free(root->m)**;		/* malloced xml data */

All will be good when called as: switch_xml_parse_str_dynamic(str, SWITCH_TRUE); switch_xml_free()

But when we do: switch_xml_parse_str_dynamic(str, SWITCH_FALSE); switch_xml_free()
Here str will be free'd as root->dynamic was set to 1, even though it was called with **SWITCH_FALSE**

As the user did not ask 'str' to be duplicated, they will then call free(str), which will be a double free situation.

Please check once if I missed something here in understanding the current code. Once confirmed I can open a PR for fixing this.

pregalla:
Yes, below code crashes with double free.

SWITCH_STANDARD_STREAM(api_stream);
switch_api_execute("sofia", "xmlstatus", NULL, &api_stream);

char *line = (char *) api_stream.data;

xmlroot = switch_xml_parse_str_dynamic(line, SWITCH_FALSE);

switch_xml_free(xmlroot);
switch_safe_free(api_stream.data);  **//double free here**

andywolk:
You should call `switch_xml_parse_str_dynamic` with `SWITCH_TRUE` in that case. So that's not a bug.

PR #1108, opened 2021-03-11
FS-1107: free memory only if it was malloc'ed

pregalla:
**Ticket ID**: https://github.com/signalwire/freeswitch/issues/1107

**Changes**
In switch_xml_parse_str_dynamic(), root->dynamic = 1, only if the function was called with SWITCH_TRUE.

### DIFF
--- a/src/switch_xml.c
+++ b/src/switch_xml.c
@@ -1024,7 +1024,9 @@ SWITCH_DECLARE(switch_xml_t) switch_xml_parse_str_dynamic(char *s, switch_bool_t
 	data = dup ? switch_must_strdup(s) : s;
 
 	if ((root = (switch_xml_root_t) switch_xml_parse_str(data, strlen(data)))) {
-		root->dynamic = 1;		/* Make sure we free the memory is switch_xml_free() */
+		if (dup) {
+			root->dynamic = 1;	/* Make sure we free the memory is switch_xml_free() */
+		}
 		return &root->xml;
 	} else {
 		if (dup) {


andywolk:
Well, now I have a questions:
How about https://github.com/signalwire/freeswitch/blob/master/src/switch_xml.c#L1782
```
*node = *root = switch_xml_parse_str_dynamic(x, SWITCH_FALSE);	/* x will be free()'d in switch_xml_free() */
```
Isn't that behaviour expected?
What's the point of the `dup` argument of `switch_xml_parse_str_dynamic(char *s, switch_bool_t dup)`?

pregalla:
> What's the point of the `dup` argument of `switch_xml_parse_str_dynamic(char *s, switch_bool_t dup)`

One use case would be to have no dependency on the argument, s, that is passed to switch_xml_parse_str_dynamic(). User may free(s) whenever he wants to, provided that was malloc'ed before.

pregalla:
> How about https://github.com/signalwire/freeswitch/blob/master/src/switch_xml.c#L1782

About this, as it was malloc'd, it can be free'd in switch_xml_free(). There should be no issue as long as one does not free(x) again.

andywolk:
This PR breaks other things.

pregalla:
> This PR breaks other things.

Right. This one as you rightly pointed out. https://github.com/signalwire/freeswitch/blob/master/src/switch_xml.c#L1782

It needs to be changed to switch_xml_parse_str_dynamic(x, SWITCH_TRUE);

Do not see any more areas that needs this change as switch_xml_parse_str_dynamic() is not that widely used.

lazedo:
@pregalla it is widely used. the caller decides to pass SWITCH_TRUE if it didn't allocate memory already or SWITCH_FALSE if it did and wants that memory to be freed.

your example `switch_safe_free(api_stream.data); //double free here` is fundamentally wrong because `you decided` to pass SWITCH_FALSE and you are double freeing the memory, nothing wrong in `switch_xml.c`

Best

andywolk:
This is a behaviour change. I'm closing as that's not a bug.

pregalla:
> @pregalla it is widely used. the caller decides to pass SWITCH_TRUE if it didn't allocate memory already or SWITCH_FALSE if it did and wants that memory to be freed.
> 
> your example `switch_safe_free(api_stream.data); //double free here` is fundamentally wrong because `you decided` to pass SWITCH_FALSE and you are double freeing the memory, nothing wrong in `switch_xml.c`
> 
> Best

@lazedo Thanks for the clarification on the usage. I was confused from the description of that function.

andywolk:
@pregalla I can feel your pain and I agree, this func is not elegant.

pregalla:
> @pregalla I can feel your pain and I agree, this func is not elegant.

Thanks @andywolk. May be we modify the comment above that function, so other users are not confused too.

andywolk:
@pregalla I will be happy to review such a PR.

PR #1124, opened 2021-03-19
freeswitch.spec: renamed format-mod-shout to format-shout

sergey-safarov:


sergey-safarov:
Could you look this PR?
@andywolk

PR #1138, opened 2021-03-25
[mod_conference] add API command to get/set conference variables

htrendev:
The conference profiles already allow for setting conference level variables, but these are essentially read-only. This pull request adds two new API commands:

- `conference <name> servar <var-name> <var-value>`
- `conference <name> getvar <var-name>`

This enables read-write access to any existing conference level variables as defined in the profile and also allows setting new conference level variables on a specific conference instance once the conference is already in progress.

Additionally, a channel variable prefix `conference_var_` is defined and can be used to pre-set any conference level variables from the dial plan just prior to starting a conference. Everything which comes after the prefix will end up as a conference variable in the conference.

For example a channel variable `conference_var_My-Conference-Var=some data` will end up as a `My-Conference-Var: some data` conference level variable and will be sent like this in the conference ESL events.


andywolk:
Implemented. Please see latest master:
https://github.com/signalwire/freeswitch/commit/4c781e2b5fbfac52154a6ecbf7069439492083e0
https://github.com/signalwire/freeswitch/commit/4f123c46828d29d8d571f0cdf03fa35381e4ea0d

snaptel:
Can this be added to the wiki?

andywolk:
Sure. Please go ahead!

snaptel:
I would love to do it but I believe I don't have permission.

andywolk:
what is your jira account?

snaptel:
user snptl

andywolk:
@snaptel I've added you to the confluence-editors group. Please try again.

snaptel:
@andywolk thanks for the access I noticed the conference variables was not documented at all (and is not in vanilla configs either) I added it to the wiki, if you can review my changes I would appreciate it (I documented it based on the code).

PR #1161, opened 2021-04-13
[Build-System] Use a fresher gcc on RHEL based dists

sergey-safarov:


andywolk:
Doing some tests with this. Will let you know.

andywolk:
Ok. Looks good to me.

andywolk (approved):

PR #1167, opened 2021-04-16
[Core] switch_rtp: fix switch_rtp_set_max_missed_packets logging

NormB:
Warning message about missed packets is misleading.
Message indicates greater (>) however, the conditional is greater than or equals (>=).
The message prints the value of rtp_sesstion->missed_count twice instead of printing the value of rtp_session->max_missed_packets.


crienzo (approved):

crienzo:
@andywolk this is OK

PR #1169, opened 2021-04-20
fix compilation with newer glibc 32-bit

neheb:
glibc as opposed to musl uses a different define to indicate whether or
not time_t is 32 or 64 bit. Add support for it.

Signed-off-by: Rosen Penev <rosenp@gmail.com>

### DIFF
--- a/src/include/switch_platform.h
+++ b/src/include/switch_platform.h
@@ -270,7 +270,7 @@ typedef intptr_t switch_ssize_t;
 #if defined(__FreeBSD__) && SIZEOF_VOIDP == 4
 #define TIME_T_FMT "d"
 #else
-#if __USE_TIME_BITS64
+#if __USE_TIME_BITS64 || (__TIMESIZE == 64)
 #define TIME_T_FMT SWITCH_INT64_T_FMT
 #else
 #define TIME_T_FMT "ld"


mjerris (changes requested):
I think this can fail if __TIMESIZE is not defined.  are you sure its always defined in paths that can hit here ?

neheb:
...how can it fail?

musl 1.2.x sets __USE_TIME_BITS64 to 1 with all platforms by default. It evaluates to 0 if undefined.

glibc 2.32 sets __TIMESIZE to 32 or 64 based on when the platform was added. It evaluates to 0 when undefined, making 0 == 64 evaluate to 0 as well.

micmac1:
> ...how can it fail?
> 
> musl 1.2.x sets __USE_TIME_BITS64 to 1 with all platforms by default. It evaluates to 0 if undefined.

Future glibc releases may see _TIME_BITS=64 and then define __USE_TIME_BITS64 to indicate that time support is 64-bit rather than 32-bit, at least that is what [this website](https://sourceware.org/glibc/wiki/Y2038ProofnessDesign) suggests.

> 
> glibc 2.32 sets __TIMESIZE to 32 or 64 based on when the platform was added. It evaluates to 0 when undefined, making 0 == 64 evaluate to 0 as well.

It looks like __TIMESIZE is used by glibc *internally* since a long time. musl doesn't use __TIMESIZE, other libcs may not use it either (or worse, it may have a different meaning), so using __TIMESIZE here may be problematic.

neheb:
I’m still not following. In this if statement, the true or false branch is identical for a 64 bit system.

For a 32 bit system with 64-bit time_t, it must hit the first one. I fail to see how this will not be true after this commit.

mjerris:
is evaluates to 0 if undefined defined behavior in the c specs?  i didnt think that was true

micmac1:
Hi all,

This is fixed by #1409 as well. I committed the patch to openwrt/telephony yesterday and now the ARC target compiles fine.

Kind regards,
Seb

PR #1170, opened 2021-04-21
[mod_callcenter] fix queue periodic announcement

greenbea:
* Fixes announcement playing to channel even agent answered or member
broke or timed out from the queue
* Fixes frequency counted only when last announcement finished playing
so announcements don't overlap each other


src/mod/applications/mod_callcenter/mod_callcenter.c
```diff
@@ -2907,11 +2907,14 @@ void *SWITCH_THREAD_FUNC cc_member_thread_run(switch_thread_t *thread, void *obj
 
 		/* If Agent Logoff, we might need to recalculare score based on skill */
 		/* Play the periodic announcement if it is time to do so */
+		if (switch_channel_get_private(member_channel, queue->announce)) {
+		    last_announce = time_now;
+		}
+
 		if (announce_valid == SWITCH_TRUE && queue->announce && queue->announce_freq > 0 &&
```
ar45, line 2914: change to `else if`

ar45 (changes requested):

src/mod/applications/mod_callcenter/mod_callcenter.c
```diff
@@ -3225,11 +3225,15 @@ SWITCH_STANDARD_APP(callcenter_function)
 		agent_found = switch_true(p);
 	}
 
-	/* Stop member thread */
+	/* Stop member thread and announcement*/
 	if (h) {
 		h->running = 0;
 	}
 
+	if (switch_channel_get_private(member_channel, queue->announce)) {
```
ar45, line 3233: At this point, the channel has no media, so displace already stopped. No reason to do this.

laflammefrancois:
FYI the same happens when you set the Lua Script to announce members position as defined in the mod_callcenter documentation;
<action application="set" data="result=${luarun(callcenter-announce-position.lua ${uuid} example@default 10000)}"/>

The caller must wait until the Lua Script execution goes back to freeswitch.msleep(mseconds) because the execution of api:executeString(uuid_broadcast, is never stopped by mod_callcenter.

yois615:
@laflammefrancois - Have you tested this patch to see if it helps your issue?

laflammefrancois:
> @laflammefrancois - Have you tested this patch to see if it helps your issue?

No, I needed a solution for my production servers. I have instead implemented a patch in my lua script; I subscribe for the CUSTOM/Answer-State/CC-Queue event and send uuid_break to the call when the answered Answer-State event is received.

Sadly, I do not have the setup to test the fix, sorry about that.

greenbea:
I've been using this fix in production without any issues for a couple of months

yois615:
@greenbea - You're using it with uuid_broadcast in a LUA script?  I don't see how your fix addresses @laflammefrancois problem.

I recompiled FS with this patch:
https://github.com/signalwire/freeswitch/commit/acbe390da617e88e066884dfe2a92bc69b8c1eaf

But I haven't tested yet.

greenbea:
I have a lua script I use for announcing the queue position, but that takes quickly to announce, so I didn't notice any issues.

The solution would be to have mod_callcenter control any announcement to the member session.

yois615:
@greenbea Can you dynamically calculate the member position?

greenbea:
> @greenbea Can you dynamically calculate the member position?

A SQL query

yois615:
@laflammefrancois I tested my patch listed above (https://github.com/signalwire/freeswitch/commit/acbe390da617e88e066884dfe2a92bc69b8c1eaf) and it solves our problem.

src/mod/applications/mod_callcenter/mod_callcenter.c
```diff
@@ -3225,11 +3230,20 @@ SWITCH_STANDARD_APP(callcenter_function)
...
 	}
 
+	if (queue_announce) {
+		switch_ivr_stop_displace_session(member_session, queue_announce);
+	}
+	// // in the case queue was reloaded
+	if ((queue = get_queue(queue_name)) && queue->announce && switch_channel_get_private(member_channel, queue->announce)) {
+		switch_ivr_stop_displace_session(member_session, queue->announce);
+	}
+	queue_rwunlock(queue);
```
andywolk, line 3245: Our coding guidelines require a new line after an `if` block.

src/mod/applications/mod_callcenter/mod_callcenter.c
```diff
@@ -3225,11 +3230,20 @@ SWITCH_STANDARD_APP(callcenter_function)
 		agent_found = switch_true(p);
 	}
 
-	/* Stop member thread */
+	/* Stop member thread and announcement*/
```
andywolk, line 3233:
```
/* Stop member thread and announcement */
```

yois615:
Correction - it does not conflict as it solves a different issue

src/mod/applications/mod_callcenter/mod_callcenter.c
greenbea: I fixed the requested change

src/mod/applications/mod_callcenter/mod_callcenter.c
```diff
@@ -3170,6 +3171,10 @@ SWITCH_STANDARD_APP(callcenter_function)
 	h->t_member_called = t_member_called;
 	h->member_cancel_reason = CC_MEMBER_CANCEL_REASON_NONE;
 	h->running = 1;
+	// save to var in case queue gets reloaded with a new announcment
```
andywolk, line 3174: Please use /* */ comment style.

andywolk (changes requested):

src/mod/applications/mod_callcenter/mod_callcenter.c
```diff
@@ -3240,16 +3245,27 @@ SWITCH_STANDARD_APP(callcenter_function)
...
+	if (queue_announce) {
+		switch_ivr_stop_displace_session(member_session, queue_announce);
+	}
+
+	/* in the case queue was reloaded */
+	if ((queue = get_queue(queue_name)) && queue->announce && switch_channel_get_private(member_channel, queue->announce)) {
+		switch_ivr_stop_displace_session(member_session, queue->announce);
+	}
+
+	queue_rwunlock(queue);
```
andywolk, line 3262: rwunlock will crash if this part returns NULL
```
(queue = get_queue(queue_name))
```

Shouldn't we use the unlock inside the if block?

andywolk (changes requested):

src/mod/applications/mod_callcenter/mod_callcenter.c
greenbea: You're totally right. I fixed it.

src/mod/applications/mod_callcenter/mod_callcenter.c
```diff
@@ -3240,16 +3245,28 @@ SWITCH_STANDARD_APP(callcenter_function)
...
+	if (queue_announce) {
+		switch_ivr_stop_displace_session(member_session, queue_announce);
+	}
+
+	/* in the case queue was reloaded */
+	if ((queue = get_queue(queue_name)) {
+		if (queue->announce && switch_channel_get_private(member_channel, queue->announce)) {
+			switch_ivr_stop_displace_session(member_session, queue->announce);
+		}
+		queue_rwunlock(queue);
```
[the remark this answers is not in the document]
greenbea: Wouldn't it be better to include information about the announcement currently playing in the `member_thread_helper` struct? This way, the member thread would be able to update the file being played, allowing `callcenter_function` to always know which announcement is currently playing, even if the queue is reloaded.

PR #1178, opened 2021-04-28
[mod_sofia] release nat mapping when nua create failed

azzbcc:


andywolk (approved):

andywolk:
Thank you very much. Good catch!

PR #1193, opened 2021-05-11
Fixed bug with sofia gateways

ar45:
When using multiple profiles and with the same gateway name on both, freeswitch mixes up the two gateways and is unable to restart any of them unless restarting the profile.

- This fixes to not remove a gateway from the hash unless it matches the profile.
- Also when doing `sofia profile <ProfileName> killgw <GatewayName>` make sure to delete the gateway on that profile by using the qualified name as the hash key `ProfileName::GatewayName`  
- when loading/reloading a gateway, we add it using the unqualified name to the hash if there is no other gateway with that name on another profile.


dragos-oancea:
the memleak is not your bug.

kvishnivetsky:
Hi. Goog fix! Ineed needed.

ar45:
@andywolk Mind reviewing this?

seven1240:
does this also cover dial string like `sofia/gateway/gw-name/number` ?

ar45:
I think so

src/mod/endpoints/mod_sofia/sofia.c
```diff
@@ -3784,18 +3784,26 @@ static void parse_gateways(sofia_profile_t *profile, switch_xml_t gateways_tag,
 			goto skip;
 		}
 
-		if (gwname && strcmp(gwname, name)) {
+		if (gwname && strcmp(gwname, n
```
andywolk, line 3787: hmm

andywolk (changes requested):

src/mod/endpoints/mod_sofia/sofia.c
andywolk: There is a conflict with upstream master. Please resolve properly.

ar45: Resolved

PR #1215, opened 2021-06-02
[mod_blacklist] Fix use after free().

andywolk:


jungle-boogie:
Does this assume this module (probably) never worked properly?

jungle-boogie:
I think the commit that introduced writing to a file never worked.

https://github.com/signalwire/freeswitch/commit/1d5f5ec7f318c771463ae2ee5b39eb62f4cd5637#diff-eb05646fd013e056bca938750a7a86758ebd60580d492abf768c1dfe75002944

dragos-oancea (approved):

mjerris (changes requested):
This won't quite do it.  after we are out of that mutex-lock that pointer isnt for sure good.  similar issue with the other hash table

jungle-boogie:
What's the best way to fix this?

jungle-boogie:
This anywhere close to being fixed?

Issue #1217, opened 2021-06-05
The reload message of distributor should be same as others


hnimminh:
**Is your feature request related to a problem? Please describe.**
When reloading the FS modules, I'm using the output to detect whether module reloaded successfully or not. 
All module return as `+OK` except distributor, it return `+ok`

**Describe the solution you'd like**
distributor should return the consistent format like other module, The message should startswith `+OK`

**Describe alternatives you've considered**
no

**Additional context**
```shell
freeswitch@libresbc1> reloadxml
+OK [Success]

freeswitch@libresbc1> reloadacl
+OK acl reloaded

freeswitch@libresbc1> distributor_ctl reload
+ok reloaded.
```

**Package version or git hash**
- Version of FreeSWITCH [e.g. 1.10.6 and earlier]

PR #1218, opened 2021-06-05
[mod_distributor] Reload message should be same as others (+OK instead of +ok)

hnimminh:
PR for #1217 the reload message of distributor should be same as others


hnimminh:
@mjerris @crienzo Can take a bit time to review this PR.
Thanks

andywolk (approved):

PR #1220, opened 2021-06-07
[mod_python3] Adding support for Python3

mdemille:
(all credit to @mochouinard, I only fixed a conflict)


mjerris:
the removal of aquire lock seems like it will break things no?

mdemille:
> the removal of aquire lock seems like it will break things no?

See https://github.com/signalwire/freeswitch/pull/1222

PR #1224, opened 2021-06-10
[CORE] Fix video disruptions caused by dropping entire frame

jakubkarolczyk:
Currently when single packet is missed and there is subsequent packet already available in buffer, which belongs to the same frame or missed packet is marked, the entire frame is dropped.
Together with buffer resync, that behavior causes video disruptions.

With this fix, we don't drop any frame, just because of loss of single packet. We carry on and pass the broken frame to decoder, which does all it's best to conceals errors.
Because one step back we've requested new keyframe, video will recover anyway in a moment and there will be smooth transition from the bad state we've entered in.

### DIFF
--- a/src/switch_jitterbuffer.c
+++ b/src/switch_jitterbuffer.c
@@ -743,8 +743,6 @@ static inline switch_status_t jb_next_packet_by_seq(switch_jb_t *jb, switch_jb_n
 {
 	switch_jb_node_t *node = NULL;
 
- top:
-
 	if (jb->type == SJB_VIDEO) {
 		if (jb->dropped) {
 			jb->dropped = 0;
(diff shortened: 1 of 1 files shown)


jakubkarolczyk:
Requesting FIR for each packet lost was the part of original code. I've only added log message up there, which I'm going to remove.
In fact, in case of some network bursts, we are not sending FIR per each packet lost, because it's capped by VIDEO_REFRESH_FREQ. So it's not too bad. Either if we drop entire frame or we pass broken frame to the decoder, practically we need to request keyframe - unless we have a client who's encoder is set with relatively short GOP, so it will send keyframe quite quickly (not likely). We could also wait for scene-cut - would be too long.

PR #1225, opened 2021-06-10
[CORE] Parameterize Video Buffer size

jakubkarolczyk:
Video Buffer size (max_frames) is hardcoded to the value of 10, which leads to overflow the buffer quite quickly, especially for higher resolutions or harder to encoded video scenes (e.g. high motion ones).

With this change, video buffer size can be configured via video-jb-size-factor parameter in switch.conf within boundaries 10 - 100.

### DIFF
--- a/src/switch_rtp.c
+++ b/src/switch_rtp.c
@@ -89,6 +89,9 @@ static const switch_payload_t INVALID_PT = 255;
 								 * characters when sending, but an implementation MUST accept up to 256
 								 * characters when receiving." */
 
+#define MAX_VIDEO_BUFFER_SIZE_FACTOR 100
+#define MIN_VIDEO_BUFFER_SIZE_FACTOR 10
+
 static switch_port_t START_PORT = RTP_START_PORT;
 static switch_port_t END_PORT = RTP_END_PORT;
 static switch_mutex_t *port_lock = NULL;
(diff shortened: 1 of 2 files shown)


dragos-oancea:
frames and max_frames can be taken from config, without the need for the "factor". how was this useful to you ? and what was the right value ?

jakubkarolczyk:
In the original code, frames are multiplied by 10. To not interfere too much, I've get rid off hardcoded value replacing it with "video-jb-size-factor" core param. I defined the min/max boundaries as 10-100. In my environment (low motion 720p stream), value of 30 is enough. I believe max value of 100 gives us space for processing higher motion, higher resolution streams (subject to system memory limitations)

src/switch_rtp.c
```diff
@@ -4770,7 +4775,13 @@ SWITCH_DECLARE(switch_status_t) switch_rtp_set_video_buffer_size(switch_rtp_t *r
 	}
 
 	if (!max_frames || frames >= max_frames) {
-		max_frames = frames * 10;
+	    if (vb_size_factor && atoi(vb_size_factor) < MAX_VIDEO_BUFFER_SIZE_FACTOR+1 && atoi(vb_size_factor) > MIN_VIDEO_BUFFER_SIZE_FACTOR) {
```
dragos-oancea, line 4778: atoi(vb_size_factor) is used 3 times here. also the indentation is not good , eg: "MAX_VIDEO_BUFFER_SIZE_FACTOR+1" should be "MAX_VIDEO_BUFFER_SIZE_FACTOR + 1" . Again, where you added switch_core_set_variable() in switch_core.c,  it should be "switch_core_set_variable("video_jb_size_factor", val);" . yes, we're picky about these things. also, commits should be squashed into 1 commit.

jakubkarolczyk:
All done, hopefully I haven't miss anything :-)

PR #1226, opened 2021-06-10
[CORE] Fix handling of packet sequence reset and reaction on packet loss greater than buffer size

jakubkarolczyk:
There are situations, especially when NACKed packet or the original (delayed one) arrives after packet sequence switch occurs. Then the buffer is unnecessary reset, causing visible video glitch. That's all because of broken got-want subtraction and followed condition.
Also, even in normal scenario, when arriving packet is really greater (in sequence) than the desired one and the difference between them is greater than buffer size (even adjusted by strange value of 17), buffer reset occurs causing video glitch.

With this fix, packet sequence switch is handle properly, assuring there won't be any broken got-want subtractions.
Instead of resetting buffer when the difference between received and desired packet is greater than buffer size, the desired packet sequence is shifted to the value determined by received sequence and buffer size. With this approach, we try to collect last missing packets, but not more than the buffer size is.

### DIFF
--- a/src/switch_jitterbuffer.c
+++ b/src/switch_jitterbuffer.c
@@ -651,27 +651,6 @@ static inline void add_node(switch_jb_t *jb, switch_rtp_packet_t *packet, switch
 	jb_debug(jb, (packet->header.m ? 2 : 3), "PUT packet last_ts:%u ts:%u seq:%u%s\n",
 			 ntohl(jb->highest_wrote_ts), ntohl(node->packet.header.ts), ntohs(node->packet.header.seq), packet->header.m ? " <MARK>" : "");
 
-	if (jb->write_init && jb->type == SJB_VIDEO) {
-		int seq_diff = 0, ts_diff = 0;
-
-		if (ntohs(jb->highest_wrote_seq) > (USHRT_MAX - 100) && ntohs(packet->header.seq) < 100) {
-			seq_diff = (USHRT_MAX - ntohs(jb->highest_wrote_seq)) + ntohs(packet->header.seq);
-		} else {
-			seq_diff = abs(((int)ntohs(packet->header.seq) - ntohs(jb->highest_wrote_seq)));
-		}
-
-		if (ntohl(jb->highest_wrote_ts) > (UINT_MAX - 1000) && ntohl(node->packet.header.ts) < 1000) {
-			ts_diff = (UINT_MAX - ntohl(node->packet.header.ts)) + ntohl(node->packet.header.ts);
-		} else {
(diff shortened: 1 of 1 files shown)


src/switch_jitterbuffer.c
```diff
@@ -1255,31 +1239,65 @@ SWITCH_DECLARE(switch_status_t) switch_jb_put_packet(switch_jb_t *jb, switch_rtp
...
+		    jb_debug(jb, 2, "Non-ucycle Got:%d Want:%d\n",got,want);
 
-				for (i = want; i < got; i++) {
-					jb_debug(jb, 2, "MARK MISSING %u ts:%u\n", i, ntohl(packet->header.ts));
-					switch_core_inthash_insert(jb->missing_seq_hash, (uint32_t)htons(i), (void *)(intptr_t)1);
-				}
-			}
+		    if (got - want > jb->max_frame_len) {
+			minwant = got - jb->max_frame_len;
+			switch_log_printf(SWITCH_CHANNEL_SESSION_LOG(jb->session), SWITCH_LOG_DEBUG, "Non-ucycle Missed more frames than buffer size. Shift want seq from %d to %d\n",want,minwant);
```
dragos-oancea, line 1254: jb_debug()

src/switch_jitterbuffer.c
```diff
@@ -1255,31 +1239,65 @@ SWITCH_DECLARE(switch_status_t) switch_jb_put_packet(switch_jb_t *jb, switch_rtp
...
+		} else if (ucycle && want > got) {
+
+		    int minwant = 0;
+		    int subgotwant = got + (USHRT_MAX - want);
+
+		    jb_debug(jb, 2, "Ucycle Got:%d Want:%d\n",got,want);
+
+		    if (subgotwant > jb->max_frame_len) {
+			minwant = got - jb->max_frame_len;
+			switch_log_printf(SWITCH_CHANNEL_SESSION_LOG(jb->session), SWITCH_LOG_DEBUG, "Ucycle Missed more frames than buffer size. Shift want seq from %d to %d\n",want,minwant);
```
dragos-oancea, line 1278: jb_debug()

dragos-oancea:
wouldn't this whole patch lead to NACK storms from us ? since we mark so many packets missing (and we don't reset the JB anymore).

jakubkarolczyk:
The maximum number of requested NACKs is limited by maximum buffer size. In other PR (video/vb-configurable-size) I defined maximum as 100 (we can lower this cap if we want to), but in my environment I configured the buffer size to 30, which is enough. With such values, the number of  requested NACKs is not too high. There could be situations, when client won't be able to send all of them, because it's buffer is too small and that's fine too. This solution is to save us as much as possible.

btw. why unit-tests are failing randomly on sofia ?

dragos-oancea:
unit-tests are failing for reasons unrelated to these patches.

PR #1227, opened 2021-06-10
[CORE] Fix video glitches on video buffer overflow

jakubkarolczyk:
Whenever there are no existing nodes to use, new node is trying to be allocated. That can happen only when the number of allocated nodes doesn't reach the current maximum video buffer size. Otherwise video buffer is forcibly reset. That causes visible, unnecessary video glitch.

With this fix, when the number of allocated nodes reach the current maximum video buffer size, the packets, which we know they won't be used anymore are discarded from the buffer. Nodes are released and one of them is given to currently processed packet straightaway, making the handling of buffer overflow very smooth.

### DIFF
--- a/src/switch_jitterbuffer.c
+++ b/src/switch_jitterbuffer.c
@@ -188,7 +188,7 @@ switch_jb_node_t *sort_nodes(switch_jb_node_t *list, int (*cmp)(const void *, co
 				e->prev = tail;
 
 				tail = e;
-			}
+			};
 
 			/* now p has stepped `insize' places along, and q has too */
 			p = q;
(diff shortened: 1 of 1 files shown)


src/switch_jitterbuffer.c
```diff
@@ -314,11 +269,91 @@ static inline void hide_node(switch_jb_node_t *node, switch_bool_t pop)
...
+		}
+
+		if (jb->allocated_nodes > jb->max_frame_len * mult) {
+			jb_debug(jb, 2, "ALLOCATED FRAMES TOO HIGH! %d\n", jb->allocated_nodes);
+			if (jb->type == SJB_VIDEO){
+			    switch_log_printf(SWITCH_CHANNEL_SESSION_LOG(jb->session), SWITCH_LOG_DEBUG, "Allocated frames too high. Need to clean rubish packets %d / %d / %d / %d\n", jb->allocated_nodes,jb->max_frame_len,mult,jb->visible_nodes);
+			}
+
+			dropped = drop_lower_seq(jb, jb->target_seq);
+			switch_log_printf(SWITCH_CHANNEL_SESSION_LOG(jb->session), SWITCH_LOG_DEBUG, "Dropped %d rubish packets. Target sequence: %u , New Visible Count: %d\n",dropped,ntohs(jb->target_seq),jb->visible_nodes);
```
dragos-oancea, line 331: there is a particular way to print DEBUG log coming from JB: jb_debug()

PR #1233, opened 2021-06-16
Security enhancement

jmateep:
Modifications to resolve issue #694

Amended the event socket code to allow for a configuration entry called authentication-hash which takes a value of the form:

$algId$salt$hash

where algId is the id of the hashing algorithm used, as follows:

MD5 has an algId of 1
SHA-256 has an algId of 5
SHA-512 has an algId of 6

hash is the hash value generated when the specified hashing algorithm is applied to an input value which consists of the password to use prepended with the salt value specified.

The code will still support the presence of a plain text password entry, but will prefer any authentication-hash provided. Appropriate logging messages will be issued if:

1) Only a password entry can be located in the config
2) Both password and authentication-hash entries are present in the config.

The module has also been modified to NOT LISTEN on the specified port and to issue a Critical logging message if no authentication data can be established (i.e. the module will not fall-back to a default password)

### DIFF
--- a/src/mod/event_handlers/mod_event_socket/Makefile.am
+++ b/src/mod/event_handlers/mod_event_socket/Makefile.am
@@ -2,7 +2,7 @@ include $(top_srcdir)/build/modmake.rulesam
 MODNAME=mod_event_socket
 
 mod_LTLIBRARIES = mod_event_socket.la
-mod_event_socket_la_SOURCES  = mod_event_socket.c
+mod_event_socket_la_SOURCES  = mod_event_socket.c mod_event_socket_hash.c
 mod_event_socket_la_CFLAGS   = $(AM_CFLAGS)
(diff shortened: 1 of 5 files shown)


dragos-oancea:
we have functions to generate sha256 and sha512 in the core. you could use those perhaps. eg: switch_digest_string("sha512-256", digest, input, strlen((char *)input), outputlen);

jmateep:
Yes. Working on that at the moment (amongst other things)

jmateep:
I pushed further changes to the remote repo which I have checked in a Debian 10 build, so hopefully these changes will be available in the pull request which I had issued originally.

Paul

jmateep:
Does anyone know if the changes I pushed to my repo are automatically included in the pull request that was issued prior to the changes that were made, or do I need to somehow cancel the original pull request (can this be done) and then issue a new one?

Paul

PR #1241, opened 2021-06-23
[Core] Remove Error code from switch_mprintf

cdevelop:
if (xtype == etSQLESCAPE4) {
						if (ch == '\'' || (xtype == etSQLESCAPE3 && ch == '\\')) {
							bufpt[j] = (char) ch;
							bufpt[j-1] = (char) '\\';
							j++;
						}
					} else {
						if (ch == '\'' || (xtype == etSQLESCAPE3 && ch == '\\'))
							bufpt[j++] = (char) ch;
					}

### DIFF
--- a/src/switch_mprintf.c
+++ b/src/switch_mprintf.c
@@ -85,8 +85,8 @@
 #define etSRCLIST    14			/* a pointer to a SrcList */
 #endif
 #define etPOINTER    15			/* The %p conversion */
-#define etSQLESCAPE3 16
-#define etSQLESCAPE4 17
+#define etSQLESCAPE3 16			/* Strings with '\'' and '\\' doubled. %w */
+#define etSQLESCAPE4 17			/* Add '\\' in front of the string '\''.  %y */
 
 /*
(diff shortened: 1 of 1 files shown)


mjerris:
we’ll need a far more in depth description and unit tests to go with a change like this

seven1240:
@mjerris it looks obvious, the second `xtype == etSQLESCAPE3` will never match when it matches the first if `xtype == etSQLESCAPE4`

```
if (xtype == etSQLESCAPE4) {
					if (ch == '\'' || (xtype == etSQLESCAPE3 && ch == '\\')) {
```

mjerris:
> @mjerris it looks obvious, the second `xtype == etSQLESCAPE3` will never match when it matches the first if `xtype == etSQLESCAPE4`
> 
> ```
> if (xtype == etSQLESCAPE4) {
> 					if (ch == '\'' || (xtype == etSQLESCAPE3 && ch == '\\')) {
> ```

but this doesn't answer what this block of code was supposed to do or tests to address those cases that confirm this works as intended.

PR #1256, opened 2021-07-09
[Core] Fix bridge_early_media

andywolk:


eridiumdev:
I've been following this issue regarding bridge_early_media for quite some time now.

We have first discovered the problem when doing a major upgrage from 1.6 to 1.10, and shortly after we set `bridge_early_media=false` to avoid the problem of non-terminating b-legs. This however broke `originate_timeout` directives, since we cannot `ignore_early_media`, and we do not `bridge_early_media`.

We didn't want to implement some ad-hoc solution, so for the time being we decided to wait for this to be resolved. In the meantime we have built master branch + this PR fix and deployed it to a portion of our production customer base. It has been working perfectly fine for a week so far, no issues with `originate_timeout` and broken b-legs, so we are planning to roll out this fix to the rest of our cluster.

Looking forward for this to be accepted, I will try to report any findings (if any)

crienzo (approved):

remdragon:
Has this fix been included in a release? If so, how can I tell which version it got included in? Thanks!

andywolk:
You can always use unstable packages or build from master.

shaunjstokes:
> Has this fix been included in a release? If so, how can I tell which version it got included in? Thanks!

You can simply apply it as a patch for now.
```
cd /tmp
https://github.com/signalwire/freeswitch/commit/f6f3f38d252511dd6203cec2fc561cf2e8c7834d.diff
cd /usr/src/freeswitch
git checkout /usr/src/freeswitch/src/switch_ivr_originate.c
git apply /tmp/f6f3f38d252511dd6203cec2fc561cf2e8c7834d.diff
make
make install
```

shaunjstokes:
We have tested this patch and can confirm it works as expected, we will switch to using this patch in our production environment.

phamhieptel4vn:
> You can always use unstable packages or build from master.

I have the same problem in Freeswitch 1.10.10-release & FusionPBX 4.4 on Debian 10. Although the actual call has ended, the channel inside Freeswitch has not been hung up, so the CDR has not been updated.
channel

PR #1259, opened 2021-07-09
[mod_opusfile] add opc_encoder_drain before destroying handle

bmlkc:
not draining before destroying encoder causes loss at the end of the encoded file

### DIFF
--- a/src/mod/formats/mod_opusfile/mod_opusfile.c
+++ b/src/mod/formats/mod_opusfile/mod_opusfile.c
@@ -316,6 +316,7 @@ static switch_status_t switch_opusfile_close(switch_file_handle_t *handle)
 	}
 #ifdef HAVE_OPUSFILE_ENCODE
 	if (context->enc) {
+		ope_encoder_drain(context->enc);
 		ope_encoder_destroy(context->enc);
 	}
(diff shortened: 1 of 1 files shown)


dragos-oancea:
there was a crash in libopusenc with ope_encoder_drain() . Was it fixed in latest version perhaps ?

bmlkc:
not observed any crashes with master branch:

image

considering that last release at 2018 (0.2.1), we tried this after upgrading to master branch to see if it resolves losing the last part of the buffer. though mod_opusfile in deb repo comes without enc so someone who wants this needs to recompile it with enc enabled as well. also observed that `HAVE_OPUSFILE_ENCODE` is not reflected to build when libopusenc is on include though just hardcoded a `#define HAVE_OPUSFILE_ENCODE 1` to bypass it.

also a similar issue exist on call setup while reading an opus file immediately after an answer, like half a second to 1 second of playback is lost. subsequent reads have no issue.

dragos-oancea:
you could give this a try: https://github.com/signalwire/freeswitch/pull/1262

dragos-oancea (approved):

andywolk (approved):

PR #1262, opened 2021-07-12
[mod_opusfile] Improvements.

andywolk:


dragos-oancea (approved):

bmlkc:
tested this with master branch of libopusenc on current stable FS release and issues mentioned for playback/record #1259 still exists and strangely adding ope_encoder_drain no longer fixes the recording buffer flush. I will give it a shot when I find some time to try with master FS and last (old) release of libopusenc.

dragos-oancea:
can you explain a bit more what you're doing and how ? my understanding is that you're try to record in .opus files and it misses some samples when the call is hungup. about the rate, I think we needed to hardcode the 48K in the module.

bmlkc:
> can you explain a bit more what you're doing and how ? my understanding is that you're try to record in .opus files and it misses some samples when the call is hungup. 

This is the extensions we call to test, there is nothing else before interacting with the channel on dial-plan:

image

Basically, recording a file then listening it again (local). The file is recorded fine, but missing some samples when the call is hang up. When calling again to listen it, it misses some samples at the start too. With the stable branch, adding `ope_encoder_drain` solves the issue of missing samples on the recorded file but not on this PR. The playback issue exists on both cases but after call is setup, the playback issue is gone i.e. playing different files back to back only first file after `answer` has the playback issue. Neither of the aforementioned problems observed when .wav is used.

More details on FS environment:
Fs version: 1.10.6 Debian
Profile: https://gist.github.com/bmlkc/c1c0a5908a61b638f60272b666f3a9fa
Modules: https://gist.github.com/bmlkc/49de007d4c11fc3a1942a2435da8e5c9
Invite received from `internal` profile.
We call from zoiper (UDP or TCP signal) or from a WebRTC phone (signal proxied from a WebRTC gateway) with same results.
There is NAT on signaling, but no issues on normal calls/presence/message etc. TLS/SRTP is on.

PR #1272, opened 2021-07-19
mod_sofia is processing the SIP BYE with delay of 5 seconds for a bridge call scenario

adil-mafzool:
mod_sofia is processing the SIP BYE with delay of 5 seconds for a bridge call scenario https://github.com/signalwire/freeswitch/issues/1268

Session queue signal data delays the handling of the SIP BYE

### DIFF
--- a/src/mod/endpoints/mod_sofia/sofia.c
+++ b/src/mod/endpoints/mod_sofia/sofia.c
@@ -2654,7 +2654,7 @@ void sofia_event_callback(nua_event_t event,
 		goto end;
 	}
 
-	if (sofia_private && sofia_private != &mod_sofia_globals.destroy_private && sofia_private != &mod_sofia_globals.keep_private) {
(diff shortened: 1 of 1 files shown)


mjerris:
not doing the queue on the bye request fixing this issue means there is a much bigger issue in processing messages somewhere else.  even on the queue it should never take more than a few hundred ms under load.  either sofia is just processing far too many messages or there is another issue to be fixed.  either way, this is not the right fix for this problem, and makes the opening to even worse DOS attacks on freeswitch.  we can’t accept this patch and i recommend you don’t run it either

adil-mafzool:
@mjerris I under what you are saying but there was single call active which was being hanged up. can you look the freeswitch log: https://github.com/signalwire/freeswitch/files/6829922/freeswitch.log to identify the cause or Is there any extra debug available to add for monitoring queue signal data?

PR #1276, opened 2021-07-21
[freetdm] Add pulse dialing support to ftmod_analog

hewittc:
This pull request adds pulse dialing support to ftmod_analog. Tested with an AEX410.


noahmehl:
Looks like freetdm has moved here: https://github.com/freeswitch/freetdm

PR #1278, opened 2021-07-21
Reducing Read RTP function blocking timeout from 5 seconds to 200ms. …

adil-mafzool:
Read RTP function blocking timeout is 5 seconds which can block SIP BYE process for 5 seconds which means the caller SIP phone may have to re-transmit SIP BYE with SIP timer. 

Changing to 200ms would help to reduce the rate of re-transmission. This would mean that SIP phone's SIP BYE will not be delayed as seen in https://github.com/signalwire/freeswitch/issues/1268


adil-mafzool:
Yes it was tested

> was this tested ?

PR #1279, opened 2021-07-21
[mod_sofia] Fix memory pool swell under high load.

andywolk:
See https://github.com/freeswitch/sofia-sip/blob/fbab3b7b3d077105378585d75d61a0a9c6fb7a4e/libsofia-sip-ua/nua/nua_stack.c#L369


src/mod/endpoints/mod_sofia/sofia.c
```diff
@@ -2666,9 +2666,18 @@ void sofia_event_callback(nua_event_t event,
...
+
  end:
 	//switch_cond_next();
 
+	/* 
+	 * This sofia_event_callback() function is called by nua_application_event()
+	 * which strictly requires application to call nua_handle_destroy()
+	 * Since we did not queue the message we should call nua_handle_destroy() now so handles don't pool swell.
+	 */
+	nua_handle_destroy(nh);
```
mjerris, line 2679: there are codepaths that have already called this above:

https://github.com/signalwire/freeswitch/pull/1279/files#diff-67b0315fd33384afe878af99d4ba8853b2ddf8a200d4545b236e52c2a29fa350R2455

also do we need to do this on the other side for queued events?

PR #1283, opened 2021-07-26
[Core] Fix FreeSWITCH crashes caused by race conditions when working with Secure RTP.

andywolk:


sergey-safarov:
> #define READ_DEC(rtp_session) rtp_session->reading--; switch_mutex_unlock(rtp_session->read_mutex)

Probably need lock and decrement.

mjerris (approved):

PR #1294, opened 2021-08-05
[mod_json_cdr] add timeout param.

zhouhailin:
add timeout param.


mjerris (dismissed):
please remove the white space changes from your patch.  we don’t allow mixed white space and code change pr’s like this, further the current formatting is fine for a one liner

zhouhailin:
> please remove the white space changes from your patch. we don’t allow mixed white space and code change pr’s like this, further the current formatting is fine for a one liner

crienzo (approved):

PR #1299, opened 2021-08-09
[Core] Add switch_safe_atol() and switch_safe_atoll() functions. Add a unit-test.

ssahoo-five9:
Add following 2 utility functions in switch_utils.h:
- static inline long int switch_safe_atol(const char *nptr, long int dft)
- static inline long long int switch_safe_atoll(const char *nptr, long long int dft)

### DIFF
--- a/src/include/switch_utils.h
+++ b/src/include/switch_utils.h
@@ -831,14 +831,38 @@ static inline char *switch_clean_name_string(char *s)
 
 
 /*!
-  \brief Turn a string into a number (default if NULL)
+  \brief Turn a string into an integer (default if NULL)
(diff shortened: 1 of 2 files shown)


Len-PGH:
@ssahoo-five9 Question:  why **dft** is **int** and not **long int** and **long long int** ?

ssahoo-five9:
> @ssahoo-five9 Question: why **dft** is **int** and not **long int** and **long long int** ?
Good point. We should give flexibility for user to set the default to a long/long long value. I will change it.  Thanks!

andywolk (approved):

PR #1305, opened 2021-08-18
[mod_conference] move conference name assignment

jamesjieye:
Move the assignment of conference name ahead of conference_video_parse_layouts() to fix the conference_name unable to pass to xml_curl gateway URL.


seven1240:
LGTM.

PR #1309, opened 2021-08-19
ignore CLion IDE configs.

zhouhailin:


dragos-oancea:
libsndfile has been removed from the FS repo, FS build scripts are expected to find it on the machine as prerequisite.

Issue #758, opened 2020-07-27
Cant start freeswitch .:. ERROR: Failed to set SCHED_FIFO scheduler (Operation not permitted)


Svenchenchen:
I have installed freeswitch on a fresh centos7.8 and I always get the error message.
freeswitch was installed from the rpm package according to the instructions.

```
ERROR: Failed to set SCHED_FIFO scheduler (Operation not permitted)
ERROR: Could not set nice level
```

Sometimes there is an entry:
`Cannot Initialize [[error near line 8161]: unexpected closing tag </condition>]`
in the journal

mjerris:
this is a permissions issue of some sort.  how are you starting freeswitch?

Svenchenchen:
Hello I'm starting Freeswitch as a service. I started it in the foreground and always get the same error message

Svenchenchen:
under root

ontur00:
Hi that someone did with that problem or maybe have advice what to do ?

[root@localhost ~]# fs_cli
[ERROR] fs_cli.c:1691 main() Error Connecting []

[root@localhost ~]# service freeswitch status
Redirecting to /bin/systemctl status freeswitch.service
● freeswitch.service - FreeSWITCH
   Loaded: loaded (/usr/lib/systemd/system/freeswitch.service; enabled; vendor preset: disabled)
   Active: failed (Result: exit-code) since pon 2020-08-24 12:20:04 CEST; 1h 46min ago
 Main PID: 11988 (code=exited, status=255)

sie 24 12:20:03 localhost systemd[1]: Started FreeSWITCH.
sie 24 12:20:04 localhost freeswitch[11988]: ERROR: Failed to set SCHED_FIFO scheduler (Operation not permitted)
sie 24 12:20:04 localhost freeswitch[11988]: ERROR: Could not set nice level
sie 24 12:20:04 localhost freeswitch[11988]: Cannot Initialize [[error near line 8810]: unexpected closing tag </condition>]
sie 24 12:20:04 localhost systemd[1]: freeswitch.service: main process exited, code=exited, status=255/n/a
sie 24 12:20:04 localhost systemd[1]: Unit freeswitch.service entered failed state.
sie 24 12:20:04 localhost systemd[1]: freeswitch.service failed.

dragos-oancea:
Solve the xml config issue first: "Cannot Initialize [[error near line 8810]: unexpected closing tag ]"

ontur00:
[...]

PR #1321, opened 2021-08-26
Assignment of User in CentOS systemd unit file

whosgonna:
Fix signalwire/freeswitch#758

It has been stated by Freeswitch developers in multiple places that rather than starting Freeswitch with a non-root user, it should be started as root, and passed a user with lower permissions as a parameter. Functionally, this is doing this:

`sudo freeswitch -u freeswitch_user`

Instead of:

`sudo -u freeswitch_user freeswitch`

The problem is that the systemd unit file for the service sets the user as it's parameter, rather than a parameter passed to freeswitch.

This PR addresses this by using the same logic that is used in the Debian systemd unit file for setting the user. There may be some other beneficial things to be gained from using even more of the Debian systemd unit file, but for now this commit is very short and to the point.

### DIFF
--- a/build/freeswitch.service
+++ b/build/freeswitch.service
@@ -6,11 +6,12 @@ After=postgresql.service postgresql-9.3.service postgresql-9.4.service mysqld.se
 [Service]
 EnvironmentFile=-/etc/sysconfig/freeswitch
 Environment="USER=freeswitch"
+Environment="GROUP=freeswitch"
 # RuntimeDirectory is not yet supported in CentOS 7. A workaround is to use /etc/tmpfiles.d/freeswitch.conf
 #RuntimeDirectory=/run/freeswitch
 #RuntimeDirectoryMode=0750
 WorkingDirectory=/run/freeswitch
-ExecStart=/usr/bin/freeswitch -nc -nf -u ${USER} $FREESWITCH_PARAMS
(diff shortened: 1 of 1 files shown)


micmac1:
If you specify the group this might cause issues. Like, when your freeswitch user is added to another group, like "dialout", so it can access some USB modem dongle, and then you start freeswitch with "-g $GROUP", then the dialout permission will be lost:

```
cat /proc/1603/status 
Name:	freeswitch
Umask:	0022
State:	S (sleeping)
Tgid:	1603
Ngid:	0
Pid:	1603
PPid:	1
TracerPid:	0
Uid:	385	385	385	385
Gid:	385	385	385	385
FDSize:	32
Groups:	 <<<< Nothing in here
NStgid:	1603
```

When on the other hand you only specify "-u $USER" it seems to take advantage of the user's groups as well:

```
cat /proc/1720/status 
Name:	freeswitch
Umask:	0022
State:	S (sleeping)
Tgid:	1720
Ngid:	0
Pid:	1720
PPid:	1
TracerPid:	0
Uid:	385	385	385	385
Gid:	385	385	385	385
FDSize:	32
Groups:	20 385 385 <<< 20 is the group id of "dialout" group
NStgid:	1720
```

Related discussion [here](https://github.com/openwrt/telephony/issues/685)  (for asterisk, really, but it's the same issue.

PR #1338, opened 2021-09-10
[mod_kazoo] Treat tag in the from tuple as an opaque and not as a ref.

xadhoom:
Fixes compatibility with OTP-24 which introduced EEP-53 (aliases).

In OTP-24 the {Pid, Ref} is {Pid, Tag} where Tag is not very well defined (seems a list with an atom). By treating it as an opaque (as it should be) no compatibiity problem arise with OTP-24 and is backwards compatible.

### DIFF
--- a/src/mod/event_handlers/mod_kazoo/kazoo_node.c
+++ b/src/mod/event_handlers/mod_kazoo/kazoo_node.c
@@ -1244,7 +1244,8 @@ static switch_status_t handle_mod_kazoo_request(ei_node_t *ei_node, erlang_msg *
 	} else if (arity == 3 && !strncmp(atom, "$gen_call", 9)) {
 		switch_status_t status;
 		ei_send_msg_t *send_msg = NULL;
-		erlang_ref ref;
+		int ref_start, ref_len = 0;
+		char *ref = NULL;
 
 		switch_malloc(send_msg, sizeof(*send_msg));
 		ei_x_new_with_version(&send_msg->buf);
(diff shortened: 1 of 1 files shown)


lazedo:
@xadhoom did you build freeswitch with otp-24 ? is this for compat between otp-24  erlang node and freeswitch built with otp-23 or lower ? can you provide a erlang snippet to test this ?

xadhoom:
> did you build freeswitch with otp-24 ?

yes


> is this for compat between otp-24 erlang node and freeswitch built with otp-23 or lower ? 

No, this is for an otp-24 erlang node speaking with fs built with otp-24.
But the fix is compatible with previous versions too, so an otp-23 node with fs built with otp-23 works too, since the patch considers the Ref as an opaque data type instead of checking it as a reference

> can you provide a erlang snippet to test this ?

Well, sort of... Elixir is ok for you? (we use mod_kazoo with a custom Elixir app, and prefer mod_kazoo instead of mod_erlang because is more complete)

But to be 100% honest there's no need for a snippet, just start an erlang node with distribution active, try to `net_adm:ping/1` the FS node from it. Without the patch, the ping fails (and mod_kazoo spits out an error), with the patch, it works.

lazedo (approved):

lazedo (approved):

toraritte:
@xadhoom @voughtdq A heads-up that `mod_kazoo` was removed from FreeSWITCH in 70d76bcc0e9970f1d0f49d16413f3e061735d87f

voughtdq:
Yep, it's [over here](https://github.com/freeswitch/mod_kazoo) now!

xadhoom:
Yep already noticed that. Unfortunately my patch that was even approved by kazoo maintainers was never merged, so I don't bother reporting it again on the new repo.

I've also tdm patches on the now out of tree freetdm, and are happily ignored.

Basically out of tree stuff is the same as being abandoned 🤷

For the records this patch still applies and is working on production since years, so is stable for me.

Should work also on OTP 25 and 26 but I've not tested it yet.

PR #1343, opened 2021-09-15
[mod_verto] fix websocket PING

seven1240:
when a ping packet is received, it `goto again` then blocks reading and expects new packets, but when the other side only sends ping in certain intervals, the reading is blocked until it receives new packets, it will block write, e.g. subscribered events cannot been delivered before the other side send any real data packets.


seven1240:
ws is replaced with kws, should be fixed libks.

PR #1355, opened 2021-09-24
allow smaller size for ECDSA certs

rjm2718:
ECDSA certificates come with much smaller keys compared to RSA.  256 bits versus 4096, for example.  FS code currently checks for minimum key length of 4096 regardless of type, which effectively precludes use of ECDSA.  This patch checks the certificate type and sets a lower limit as needed.

### DIFF
--- a/src/switch_core_cert.c
+++ b/src/switch_core_cert.c
@@ -365,7 +365,8 @@ SWITCH_DECLARE(switch_bool_t) switch_core_check_dtls_pem(const char *file)
 	}
 
 	bits = EVP_PKEY_bits(pkey);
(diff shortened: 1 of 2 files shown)


src/switch_core_cert.c
```diff
@@ -336,6 +336,7 @@ SWITCH_DECLARE(switch_bool_t) switch_core_check_dtls_pem(const char *file)
 	FILE *fp = NULL;
 	EVP_PKEY *pkey = NULL;
 	int bits = 0;
+	int min_cert_size_bits = 0;
```
dragos-oancea, line 339: don't assign 0 to it, it will bother scan-build.

f-vicente:
In 2026 I feel that EDCSA usage has increased. I hope this could be approved/fixed.

PR #1363, opened 2021-09-30
GPG key server update after keyservers.net shutdown, fixes #1362

### LINKED ISSUE #1362: Freeswitch Dockerfile broken by SKS keyserver shutdown since June of 2021
**Describe the bug**
https://sks-keyservers.net/ was the news, the dockerfile still tries to use keyfiles from it. Will submit a PR to fix.

**To Reproduce**
Try to build the dockerfile, it will fail (host cannot be resolved on key add).

**Expected behavior**
Not fail:)

mitchcapper:
fixes #1362

Both dockerfiles have been unbuildable since June of 2021 when keyservers.net shutdown: https://sks-keyservers.net/

This updates the fetching of the keys to just pull from ubuntu's server.  Also added it to the main step to avoid an extra layer.

### DIFF
--- a/docker/master/Dockerfile
+++ b/docker/master/Dockerfile
@@ -8,10 +8,10 @@ FROM debian:jessie
 RUN groupadd -r freeswitch --gid=999 && useradd -r -g freeswitch --uid=999 freeswitch
 
 # grab gosu for easy step-down from root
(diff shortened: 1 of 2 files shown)


kantum (approved):
This looks pretty good

PR #1385, opened 2021-10-15
[Configuration] Fix Brazil daylight’s in xml files

Fernandojdk:
Fixs wrong daylight saving on America/Sao_Paulo Timezone in all config files.


andywolk:
@Fernandojdk Ping

PR #1409, opened 2021-10-27
[build] properly fix time_t issues

ericonr:
aa71d87528643fd1b3897a64ecec8c11e92b5b55 tried fixing the issue by
adding yet another case to the TIME_T_FMT definition, but hardcoding
values as hinted at by internal variables or by platform is not future
proof or allows for improvements.

The most portable fix for time_t handling is to always cast it to
(long long) for printing and use "lld" for formatting.


micmac1:
Run-tested on musl 1.2.2 on 32 bit target. The other issue I had was unrelated to FS.

Thanks @ericonr !!

Kind regards,
Seb

crienzo:
@mjerris can you look at this?

PR #1413, opened 2021-10-31
[mod_sofia] Randomize OPTIONS Ping interval

### LINKED ISSUE #1132: OPTIONS pings are sent out all at once
**Describe the bug**
When you turn on the configuration to send out options pings (all-reg-options-ping and similar) the first set of OPTIONS pings get sent out all at once. There is code to randomise the interval between one OPTIONS ping and the next, unfortunately the same random offset is applied to all the user agents who are due an OPTIONS ping. So from the point of view of a single user agent, the OPTIONS pings arrive at the randomly-varied intervals specified, but FreeSWITCH still sends out all the OPTIONS pings at the same time.

The code (sofia_reg_check_ping_expire() in sofia_reg.c) checks the ping_expires field in the sip_registrations table to identify user agents whose ping is about to expire. It then sends OPTIONS pings to those destinations. Then it chooses the next ping_expires time by adding the configured time +/- a random amount. (This was the subject of FS-6400.) It then updates those registrations with the new ping_expires time with a single sql update, meaning they all get the same value.

**To Reproduce**
1. In a system with a number of registered user agents, turn the options ping option on.
2. Options pings will continue to be sent out to all user agents at the configured interval and random offset, but that interval will apply to all user agents.

**Expected behavior**
The randomisation in the timing of sending OPTIONS pings was added with [FS-6400](https://freeswitch.org/jira/browse/FS-6400) and the clear intention was to spread out sending of OPTIONS pings over time to reduce load on FreeSWTICH processing the replies and to be nicer to an upstream sbc or proxy.

 - Version 1.10.5 from packages

### LINKED ISSUE #1133: OPTIONS pings that are separated in time gradually converge
**Describe the bug**
When investigating https://github.com/signalwire/freeswitch/issues/1132, I manually set the ping_expires times in the sip_registrations table to all different values. About 20 minutes later they had converged on to mainly one value.

**To Reproduce**
Steps to reproduce the behaviour:
1. Set ```<param name="all-reg-options-ping" value="true"/>```
2. Manually separate the ping_expires times: I did: ```update sip_registrations set ping_expires=ping_expires+round(random()*60);```
3. Wait. The times converge.

**Expected behaviour**
I would expect the times to remain separated out.

 - Version 1.10.5

**Database output**
ping_expires is a Unix timestamp. The starting point shown here has expires times between 11:35:24 & 11:36:17
```
sofia_reg_internal=# select sip_user, ping_expires from sip_registrations order by ping_expires;
 sip_user | ping_expires
----------+--------------
 001      |   1616499324
 001      |   1616499325
 8855     |   1616499327
 8835     |   1616499328
 001      |   1616499328
 8833     |   1616499329
 001      |   1616499330
 8836     |   1616499330
 401      |   1616499331
 8855     |   1616499333
 8838     |   1616499334
 8833     |   1616499335
 001      |   1616499335
 263      |   1616499335
 8855     |   1616499335
 001      |   1616499339
 8895     |   1616499341
 002      |   1616499344
 003      |   1616499344
 002      |   1616499345
 833      |   1616499348
 001      |   1616499350
 002      |   1616499350
 836      |   1616499350
 001      |   1616499353
 835      |   1616499353
 403      |   1616499354
 001      |   1616499354
 838      |   1616499355
 8833     |   1616499356
 8833     |   1616499358
 402      |   1616499364
 001      |   1616499364
 8836     |   1616499364
 003      |   1616499377
(35 rows)
```
20 minutes later many of the ping_expires are the same: (11:55:03)
```
sofia_reg_internal=# select sip_user, ping_expires from sip_registrations order by ping_expires;
 sip_user | ping_expires
----------+--------------
 8835     |   1616500503
 8836     |   1616500503
 8895     |   1616500503
 8838     |   1616500503
 838      |   1616500503
 835      |   1616500503
 263      |   1616500503
 403      |   1616500503
 001      |   1616500503
 002      |   1616500503
 8833     |   1616500503
 001      |   1616500503
 8855     |   1616500503
 836      |   1616500503
 001      |   1616500503
 002      |   1616500503
 402      |   1616500503
 001      |   1616500503
 001      |   1616500503
 002      |   1616500503
 8833     |   1616500503
 8855     |   1616500503
 001      |   1616500503
 833      |   1616500503
 8833     |   1616500503
 401      |   1616500503
 8855     |   1616500503
 003      |   1616500503
 001      |   1616500503
 001      |   1616500505
 001      |   1616500514
 003      |   1616500514
 001      |   1616500514
 8833     |   1616500514
 8836     |   1616500523
(35 rows)
```

yois615:
In FS-6400, the attempt was made to randomize OPTIONS
packets to be sent at a random interval.  The same random
interval is applied to all endpoints so this doesn't work.

Furthermore, rounding withing the code will ultimately
make the ping times converge over time.

This commmit will apply the random interval only during
registration of the endpoint.  All subsequent pings will
be incremented with the actual value of ping-mean-interval.

Fixes #1132, Fixes #1133

### DIFF
--- a/src/mod/endpoints/mod_sofia/sofia.c
+++ b/src/mod/endpoints/mod_sofia/sofia.c
@@ -2949,6 +2949,9 @@ void *SWITCH_THREAD_FUNC sofia_profile_worker_thread_run(switch_thread_t *thread
 
 	sofia_set_pflag_locked(profile, PFLAG_WORKER_RUNNING);
 
+	/* Seed PRNG for functions within worker thread */
+	srand((unsigned)((intptr_t) switch_thread_self() + switch_micro_time_now()));
+
 	while ((mod_sofia_globals.running == 1 && sofia_test_pflag(profile, PFLAG_RUNNING))) {
 
 		if (tick) {
--- a/src/mod/endpoints/mod_sofia/sofia_presence.c
+++ b/src/mod/endpoints/mod_sofia/sofia_presence.c
@@ -3755,7 +3755,6 @@ void sofia_presence_handle_sip_i_subscribe(int status,
 
 	if ((sub_max_deviation_var = profile->sip_subscription_max_deviation)) {
 		int sub_deviation;
-		srand( (unsigned) ( (unsigned)(intptr_t)switch_thread_self() + switch_micro_time_now() ) );
 		/* random negative number between 0 and negative sub_max_deviation_var: */
 		sub_deviation = ( rand() % sub_max_deviation_var ) - sub_max_deviation_var;
 		if ( (exp_delta + sub_deviation) > 45 ) {
--- a/src/mod/endpoints/mod_sofia/sofia_reg.c
+++ b/src/mod/endpoints/mod_sofia/sofia_reg.c
@@ -884,7 +884,6 @@ long sofia_reg_uniform_distribution(int max)
 	int result;
 	int range = max + 1;
 
-	srand((unsigned)((intptr_t) switch_thread_self() + switch_micro_time_now()));
 	result = (int)((double)rand() / (((double)RAND_MAX + (double)1) / range));
 
 	switch_log_printf(SWITCH_CHANNEL_LOG, SWITCH_LOG_DEBUG9, "Generated random %ld, max is %d\n", (long) result, max);
@@ -894,8 +893,7 @@ long sofia_reg_uniform_distribution(int max)
 void sofia_reg_check_ping_expire(sofia_profile_t *profile, time_t now, int interval)
 {
 	char *sql;
-	int mean = interval / 2;
-	long next, irand;
+	long next;
 	char buf[32] = "";
 	int count;
 
@@ -952,8 +950,7 @@ void sofia_reg_check_ping_expire(sofia_profile_t *profile, time_t now, int inter
 		/* only update if needed */
 		if (count) {
 			switch_log_printf(SWITCH_CHANNEL_LOG, SWITCH_LOG_DEBUG9, "Updating ping expires for profile %s\n", profile->name);
-			irand = mean + sofia_reg_uniform_distribution(interval);
-			next = (long) now + irand;
+			next = (long) now + interval;
 
 			sql = switch_mprintf("update sip_registrations set ping_expires = %ld where hostname='%q' and profile_name='%q' and ping_expires <= %ld ",
 								 next, mod_sofia_globals.hostname, profile->name, (long) now);
@@ -1765,7 +1762,6 @@ uint8_t sofia_reg_handle_register_token(nua_t *nua, sofia_profile_t *profile, nu
 			     (( exp_max_deviation_var = profile->sip_expires_max_deviation )) ) {
 				if (exp_max_deviation_var > 0) {
 					int exp_deviation;
-					srand( (unsigned) ( (unsigned)(intptr_t)switch_thread_self() + switch_micro_time_now() ) );
 					/* random number between negative exp_max_deviation_var and positive exp_max_deviation_var: */
 					exp_deviation = ( rand() % ( exp_max_deviation_var * 2 ) ) - exp_max_deviation_var;
 					exptime += exp_deviation;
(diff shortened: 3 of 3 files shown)


yois615:
As a newcomer to this project, I don't know how FreeSWITCH handles a full restart, or a restart of just mod_sofia.  If registrations persist a restart, then this patch as proposed here could result in all the OPTIONS being sent out at the same time if ping_expires has passed for all of the registrations while the code was not running, since the randomization only occurs upon REGISTER.

This can be easily fixed by implementing the same database update a few lines later during registration updates, but I'm leaving as is pending review from someone more familiar with how all this works.

yois615:
After testing, I decided that this needs to be done in the registration updates as well.  I have this tested and running on my prod servers with what appears to be good results.

dragos-oancea:
it looks this changes default behavior. can you put it under a mod_sofia flag ? And what is the gain with the random interval ?

yois615:
This patch is really to implement the intended behavior of the original code. See the linked issues for details.

dragos-oancea:
so it does not "Randomize OPTIONS Ping interval" , but it scatters the initial OPTIONS requests across an interval so we don't send out all at once.

yois615:
Exactly

dragos-oancea:
any inclination to unit-test this ?

yois615:
I was thinking it's not directly necessary.  I didn't review the current test suite, but assuming that we're testing for OPTIONS, the fact that the frequency changes doesn't impact on the test.

The only test that may prove useful (but I still think it's unnecessary) is to ensure that the interval is consistent over 3 pings or so after the randomness is set, so we ensure that we continue to ping after the initial OPTIONS.

BTW - This is also in the commit message, but an important code change here is where SRAND is located.  A problem with randomization was that the code was constantly reseeding - which breaks randomization.  I would appreciate another set of eyes on that part of this to ensure nothing else is broken.  (I have had this running on my production systems for the last 2 months with no issues though).

src/mod/endpoints/mod_sofia/sofia.c
```diff
@@ -2949,6 +2949,9 @@ void *SWITCH_THREAD_FUNC sofia_profile_worker_thread_run(switch_thread_t *thread
 
 	sofia_set_pflag_locked(profile, PFLAG_WORKER_RUNNING);
 
+	/* Seed PRNG for functions within worker thread */
+	srand((unsigned)((intptr_t) switch_thread_self() + switch_micro_time_now()));
```
dragos-oancea, line 2953: you could use switch_time_now() instead of switch_micro_time_now() for better seed.

yois615: Can you explain why? The seed is longer with micro.

Thanks for the review!

dragos-oancea: it's just clock precision, switch_micro_time_now() does not increment that often practically as switch_time_now(). however, it seems it's fine like this.

dragos-oancea (dismissed):

yois615:
@dragos-oancea - Next steps to merge?

yois615:
@themsley-voiceflex I have this running on production systems for about 3 years with no ill effect, and I can confirm it works as intended by watching the ping packets in sngrep.

I still run 1.10.7 so it's possible something else conflicts with this. I don't understand how it's possible for this commit to cause what you're seeing to be occurring. What you're describing is the behavior that existed before and what this commit is trying to correct.

themsley-voiceflex:
It calculates a single value for the new expiry time based on now + interval and then updates all rows with ping_expires < now with the same ping_expires value. If you have 1000 rows that expired in the last 'interval' then with this patch all 1000 of them now expire at once. Next time round we will find the same 1000 as before (because the interval has expired again) and will now update 1000 plus however many others have expired in the meantime. You can see the result by running `select hostname,ping_expires, count(*) as HM from sip_registrations group by hostname,ping_expires order by HM desc;`

themsley-voiceflex:
https://github.com/signalwire/freeswitch/commit/01c96285cacf18c4cdc97c2ff93e2b1d54d354b0#diff-81784d87973e12d6a7a36b72b1dae9dd2f35190018848a1ec5a155eddd6f69ebL954

yois615:
Yes, this is the desired behavior. Since the registrations already happened randomly, we don't need to add another randomization here. Just keep the interval for that entire group.

To be clear, are you seeing convergence happen with random registrations? Or are you seeing that a group of registrations will keep pinging at the same time? 

If you read back through my initial comment on this commit, and the two linked issues, I think it'll give you a better understanding of what this commit was trying to fix. The way randomization was being handled before, it would cause a ping convergence of every single registration regardless of when it came online.

Can I ask you what behavior you would like to see?

themsley-voiceflex:
The query I posted above results in this sort of distribution:
`+----------+--------------+------+
| hostname | ping_expires | HM   |
+----------+--------------+------+
| host1    |   1729013255 | 6210 |
| host2    |   1729013257 |  457 |
| host1    |   1729013253 |  204 |
| host1    |   1729013270 |  109 |
`

This happens because the code calculates the new expiry time based on now + interval so anything expired since last time will be set to the same value. I think instead of that it should use

`update sip_registrations set ping_expires = ping_expires + %d where ping_expires > 0 and ping_expires <= %ld and hostname='%q' and profile_name='%q'`

This then preserves the original expiry time and adds 30s to it (or whatever interval comes from the profile). This will not have the same effect of flattening all expiry times to the same one over time.

themsley-voiceflex:
[freeswitch-1.10.12-ping-expires.PATCH](https://github.com/user-attachments/files/17382490/freeswitch-1.10.12-ping-expires.PATCH)

yois615:
Yup, I get it.  Your problem isn't happening in my environment, I suppose because all my registrations renew before they expire so the amount of expired registrations stays fairly constant.  I'm not sure why you would have convergence of expirations but that's a moot point here.  I also wonder how your patch would handle a profile restart that takes longer than `interval`.

Raise a PR, I'll try to merge it (to my repo, I'm not a project maintainer) and test.  If you don't mind, keep the same order in the WHERE clause so it's easier to read exactly what you're changing - it's really only one thing.

yois615:
Wait, I just remembered why we need to use NOW (and this is what I was thinking when I wrote it):

If you enable NAT-PING when extensions are already registered, they won't get a ping until they first completely de-register, since there's no existing time to increment from.

themsley-voiceflex:
The order in the WHERE clause seems to make a difference to which indexes mariadb uses to work out which rows to lock. If ping_expires is last then it uses the sr_hostname findex and locks all rows belonging to that hostname. If it's first and ping_expires has a decent spread of values then it'll only lock those rows when doing the update and get less deadlocks. That's part of the point of the change to this code, to stop it locking _all the rows_. That part of this patch came first and then only then did I find they were all the same value so all get locked anyway.

I now have a max of 57 ping_expires identical rows, down from 457 so the patch seems to be working for me.

themsley-voiceflex:
In the NAT-PING case is the initial ping_expires == 0?

yois615:
0 or NULL, I don't think it's set and I haven't tested.

themsley-voiceflex:
With that patch applied, my distribution of ping_expires values for the updated system is now spread fairly evenly across each of 30 values, 1 per second. I have not yet applied this to the busier system because I need to wait to see how 1.10.12 behaves on the one that I have updated. 

My initial thought is that we should be doing 2 sets of updates here, one for things that already have a ping_expires value and one for ones that do not (0 or NULL) but I very rarely see any rows come up on a query that have ping_expires == 0 and if I do then they've gone by the time I can recall that query and re-run it - but those may be coming from the unpatched system which would update them away the first time it ran that code.

themsley-voiceflex:
I have now applied the patch to our busiest system which has roughly 6500 rows in sip_registrations for the internal profile. Around 95% of entries in sip_registrations all had the same ping_expires value before this patch. After the patch that became 30 rows with approx 6500/30 rows per entry (+/- a fudge factor). I then added `<param name="ping-mean-interval" value="120"/>` as I figured that even trying to ping 220 things a second is a bit too much. That gave me 120 separate ping_expires values with between 40-75 rows each. It's been in and running there for 3 or 4 days so far.

I am 100% sure that the current update statement is incorrect and actually does the opposite of the intent of the patch and tends to coalesce all ping_expires to the same value. My current patch stops this from happening completely on my systems and I have no rows with ping_expires == 0 or any value much less than the current unix timestamp. Is it possible for you to apply this to your systems that showed a similar problem before and see if they still show the same problems as before? I'd like to see any anomalies you have from this to find out why your systems are different.
[freeswitch-1.10.12-ping-expires-latest.PATCH](https://github.com/user-attachments/files/17523092/freeswitch-1.10.12-ping-expires-latest.PATCH)

yois615:
Before doing this, I want to clarify what happens if options pinging is off on a profile and then enabled.  The current version will assign a value based on now, so the value will be correct immediately.  With this patch I'm not sure what will happen.

This should be easy to test - create a new profile on one of your systems, leave options ping off for that profile, register an endpoint, and run your sql query after a few minutes (`select hostname,ping_expires, count(*) as HM from sip_registrations group by hostname,ping_expires order by HM desc;` add a WHERE so you get only this profile...) to see what ping_expires is.  Then turn on options ping for that profile (does that need a profile restart?), and track ping_expires to see what happens  Since this code loops every second it may require getting enough of a differential between ping_expires and now to see what's happening.  Also test to see what happens when changing the mean interval both shorter and longer.

I'm pretty sure your code is still good because it looks to me like ping_expires is set for every endpoint regardless of whether pinging is on or off, but it needs to be tested first.

I appreciate your work on this.

PR #1431, opened 2021-11-05
[mod_sofia] Include Callee ID name in presence NOTIFY

### LINKED ISSUE #1410: Retrieve effective_caller_id_name for presence dialog-info+XML
In this line here:
https://github.com/signalwire/freeswitch/blob/bf62bd8a7fd01339f4251cc23311b95ed768b109/src/mod/endpoints/mod_sofia/sofia_presence.c#L2995
We use clean_to_user for the identity and the contact.  The display name can be retrieved from the switch to display the effective_caller_id_name of the extension to make this information more accurate and complete.

yois615:
Retreive user_data from the switch to display the callee
ID name on the endpoint receiving the NOTIFY if it is
available

Fixes #1410


src/mod/endpoints/mod_sofia/sofia_presence.c
```diff
@@ -2987,12 +2987,29 @@ static int sofia_presence_sub_callback(void *pArg, int argc, char **argv, char *
...
 
+				char *cmd;
+				char *callee_id_name = NULL;
+				switch_stream_handle_t user_data_stream = { 0 };
+
 				clean_to_user = switch_mprintf("%s", sub_to_user ? sub_to_user : to_user);
 				clean_from_user = switch_mprintf("%s", from_id ? from_id : from_user);
 
+				SWITCH_STANDARD_STREAM(user_data_stream);
+				cmd = switch_mprintf("%s@%s var effective_caller_id_name", clean_to_user, host);
```
andywolk, line 2998: Where is `cmd` freed?

yois615: When it falls out of scope?  cmd is declared on the stack at 2990

andywolk: is memory allocated by `switch_mprintf` freed?

yois615: Ack

src/mod/endpoints/mod_sofia/sofia_presence.c
```diff
@@ -2987,12 +2987,30 @@ static int sofia_presence_sub_callback(void *pArg, int argc, char **argv, char *
...
+				if (switch_api_execute("user_data", cmd, NULL, &user_data_stream) == SWITCH_STATUS_SUCCESS) {
+					if (strncmp("-ERR", user_data_stream.data, 4)) {
+						callee_id_name = switch_mprintf("%s", (char *)user_data_stream.data);
```
andywolk, line 3001: This is a memory leak.

yois615: I changed this to switch_copy_string and pushed again.

andywolk: Why? Could you just free it later?

yois615: It has been freed at 3083.

I thought you were concerned about a dangling pointer with the reassignment of callee_id_name, and TBH didn't understand your comment here.

Anyway switch_copy_string doesn't null terminate a truncated string so the code shouldn't be merged as it is now regardless.

Shall I revert to switch_mprintf and can you clarify what the memory leak is, or I have a lab up now with switch_snprinf and will push once I confirm it works.

andywolk: I could have github not updating the change to me properly. So please revert and I will take a look again.

src/mod/endpoints/mod_sofia/sofia_presence.c
```diff
@@ -2987,12 +2987,27 @@ static int sofia_presence_sub_callback(void *pArg, int argc, char **argv, char *
...
+				switch_stream_handle_t user_data_stream = { 0 };
+
 				clean_to_user = switch_mprintf("%s", sub_to_user ? sub_to_user : to_user);
 				clean_from_user = switch_mprintf("%s", from_id ? from_id : from_user);
 
+				SWITCH_STANDARD_STREAM(user_data_stream);
+				cmd = switch_mprintf("%s@%s var effective_caller_id_name", clean_to_user, host);
+				if (switch_api_execute("user_data", cmd, NULL, &user_data_stream) == SWITCH_STATUS_SUCCESS) {
+					if (strncmp("-ERR", user_data_stream.data, 4)) {
+						switch_snprintf(callee_id_name, sizeof(callee_id_name), "%s", (char *) user_data_stream.data);
```
andywolk, line 3001: do you really need to copy?

yois615: This looked cleaner to my eye but I will revert as you requested earlier.

yois615: Without me reverting, you can see that commit here:
https://github.com/signalwire/freeswitch/commit/8ea0af693294d675fb494c5c923853f0389b2c7c

andywolk: The question above remains.

yois615: switch_safe_free(callee_id_name); on 3083 doesn't help that?

andywolk: the question is do you really need to copy?

yois615: Sorry I'm not getting it and I'm sure it's on my end :)

You're saying even with `switch_mprintf` there's another way to do this by directly referring to `user_stream_data.stream` below?  What if I return -ERR?

I could refactor the whole thing to be in the `if (is_dialog) `branch and pass directly to stream.write_function... Is that what you're suggesting?

I appreciate your time and patience.

andywolk: You can set `callee_id_name` to `user_data_stream.data` when you need and forget about freeing `callee_id_name` as you free the stream anyways.

yois615: I thought it's a memory leak to reassign a pointer to another pointer since the initial memory will remain unfreed.  Looking at it now, user_data_stream isn't a pointer...

OK, I pushed that.  I left out the typecast of (char *)user_data_stream because I don't know how it's possible to point to a cast (beyond my pay grade).

andywolk: you need to cast on the right.

src/mod/endpoints/mod_sofia/sofia_presence.c
```diff
@@ -2987,12 +2987,26 @@ static int sofia_presence_sub_callback(void *pArg, int argc, char **argv, char *
...
 				clean_from_user = switch_mprintf("%s", from_id ? from_id : from_user);
 
+				SWITCH_STANDARD_STREAM(user_data_stream);
+				cmd = switch_mprintf("%s@%s var effective_caller_id_name", clean_to_user, host);
+				if (switch_api_execute("user_data", cmd, NULL, &user_data_stream) == SWITCH_STATUS_SUCCESS) {
+					if (strncmp("-ERR", user_data_stream.data, 4)) {
+						callee_id_name = (char *)user_data_stream.data;
+					}
+				}
+				switch_safe_free(cmd);
```
andywolk, line 3004: This line does not follow the coding guidelines. Please see https://github.com/signalwire/libks/blob/master/CODING_GUIDELINES#L36

yois615:
@Len-PGH

src/mod/endpoints/mod_sofia/sofia_presence.c
```diff
@@ -2987,12 +2987,27 @@ static int sofia_presence_sub_callback(void *pArg, int argc, char **argv, char *
...
+				char *cmd;
+				char *callee_id_name = NULL;
+				switch_stream_handle_t user_data_stream = { 0 };
+
 				clean_to_user = switch_mprintf("%s", sub_to_user ? sub_to_user : to_user);
 				clean_from_user = switch_mprintf("%s", from_id ? from_id : from_user);
 
+				SWITCH_STANDARD_STREAM(user_data_stream);
+				cmd = switch_mprintf("%s@%s var effective_caller_id_name", clean_to_user, host);
+				if (switch_api_execute("user_data", cmd, NULL, &user_data_stream) == SWITCH_STATUS_SUCCESS) {
```
andywolk, line 2999: I have concerns about the need of calling the `user_data` API. Shouldn't we have all the information in the `helper->event` ?
`switch_event_get_header(helper->event, ...)`

yois615: I don't have good knowledge of that object. But I would wonder why the effective_caller_id_name would be included. We are handling an inbound call and need to retrieve userdata for the notifying extension. Up until this point in the call flow this data is irrelevant.

yois615:
I found a problem but it's somewhat unrelated to this actual code change.
The longer display name of Callee ID info in dialog-info+XML can cause the packet to exceed 1514 bytes, and only the first part of the fragmented packet is being sent.  Any suggestions how to fix?  I'm noticing that the top of the SIP header seems excessive for a NOTIFY:

`Allow: INVITE, ACK, BYE, CANCEL, OPTIONS, MESSAGE, INFO, UPDATE, REGISTER, REFER, NOTIFY, PUBLISH, SUBSCRIBE
Supported: timer, path, replaces
Event: dialog
Allow-Events: talk, hold, conference, presence, as-feature-event, dialog, line-seize, call-info, sla, include-session-description, presence.winfo, message-summary, refer`

src/mod/endpoints/mod_sofia/sofia_presence.c
yois615: I am reverting back to copying.
I noted below that we potentially exceed MTU in UDP packets.  Since CallerID is only supposed to contain 15 characters, the string should be truncated.  We copy with switch_snprintf and truncate to 16 bytes.  This will help alleviate at least some of the problem.

src/mod/endpoints/mod_sofia/sofia_presence.c
yois615: I enabled the prescence debug that does a DUMP_EVENT(helper->event) and can confirm that the information I'm looking for is not available without calling the API.

yois615:
Oh my, IDK why that happened

yois615:
@andywolk - Looking at the compare it shows a whole bunch of garbage but the commit looks clean.  Should I rebase?

Issue #1440, opened 2021-11-15
Compilation with GCC 11.2.1


ROBERT-MCDOWELL:
Hi,

had a compile error on Fedora 35 (CentOS equivalent) Gcc 11.2.1

  CC       libs/libnatpmp/libfreeswitch_la-getgateway.lo
  CXX      src/switch_cpp.lo
  CXXLD    libfreeswitch.la

*** Warning: Linking the shared library libfreeswitch.la against the
*** static library libs/libvpx/libvpx.a is not portable!
Making all in .
make[2]: Entering directory '/home/src/freeswitch'
  CC       src/freeswitch-switch.o
  CCLD     freeswitch
/usr/bin/ld: ./.libs/libfreeswitch.so: undefined reference to `pthread_mutexattr_setrobust                                                                                                                                                    _np'
collect2: error: ld returned 1 exit status
make[2]: *** [Makefile:1994: freeswitch] Error 1
make[2]: Leaving directory '/home/src/freeswitch'
make[1]: *** [Makefile:3716: all-recursive] Error 1
make[1]: Leaving directory '/home/src/freeswitch'
make: *** [Makefile:1448: all] Error 2

Any idea?

thanks

ROBERT-MCDOWELL:
apparently it comes from libvpx

andywolk:
What's your linux kernel version?

ROBERT-MCDOWELL:
kernel 5.15.2-rt

ROBERT-MCDOWELL:
I just recompiled the kernel but still gets the same error not with libvpx but with others

/usr/bin/ld: ./.libs/libfreeswitch.so: undefined reference to `pthread_mutexattr_setrobust                                                                                                                                                    _np'
collect2: error: ld returned 1 exit status
make[2]: *** [Makefile:2077: tone2wav] Error 1
make[2]: *** Waiting for unfinished jobs....
/usr/bin/ld: ./.libs/libfreeswitch.so: undefined reference to `pthread_mutexattr_setrobust                                                                                                                                                    _np'
collect2: error: ld returned 1 exit status
make[2]: *** [Makefile:2038: fs_encode] Error 1
/usr/bin/ld: ./.libs/libfreeswitch.so: undefined reference to `pthread_mutexattr_setrobust                                                                                                                                                    _np'
collect2: error: ld returned 1 exit status
make[2]: *** [Makefile:2071: fs_tts] Error 1
[...]

PR #1443, opened 2021-11-16
Fix gcc-11.x.x compilation warnings considered as fatal

ROBERT-MCDOWELL:
...


src/mod/databases/mod_pgsql/mod_pgsql.c
```diff
@@ -774,6 +774,11 @@ switch_status_t pgsql_next_result_timed(switch_pgsql_handle_t *handle, switch_pg
 		case PGRES_COPY_IN:
 		case PGRES_COMMAND_OK:
 			break;
+//#if POSTGRESQL_MAJOR_VERSION >= 14
```
andywolk, line 777: not going to build on older versions

ROBERT-MCDOWELL: Do you know how to put this condition out of the switch since gcc-11.2.x does not allow it.

ROBERT-MCDOWELL: ok, I fixed the conflicts with master, but anyhow it cannot compile with gcc-11.x.x for now

ROBERT-MCDOWELL: ok my bad, it seems that now the issue went away. consider this patch ok.
however the other issue I posted is still persisting (#1440 )

ROBERT-MCDOWELL: I sent another PR that fixes the problem on the #if #endif POSTGRESQL_MAJOR condition into the switch.
gcc 11.2 does not allow a condition into a switch unless the "default: break;" is specified.
everything is compiling fine now (with all the patches I sent)

ROBERT-MCDOWELL (commented):
fix compliation with gcc 11.2.x

dragos-oancea:
we merged  gcc-11 related fixes, but only for the core (b4a7cf40597a7746e10d11e5bc6b3b6ee3fb0af6) . please approach the module fixes in a separate PR if you wish.

PR #1444, opened 2021-11-16
[mod_sofia] fail set up of STIR/SHAKEN verification service if configured CA directory fails to load

crienzo:


andywolk (approved):

crienzo:
@andywolk this is all set

PR #1458, opened 2021-11-23
[mod_sofia] Rewrite rport even if endpoint specifies port in contact

### LINKED ISSUE #1457: NDLB-force-rport broken when endpoint specifies port in contact
**Describe the bug**
Yealink T48S phone registers to FS from behind NAT.  The device includes a Contact in the request like so:
Contact: <sip:102@192.168.6.103:5060>

The actual received IP:port are 22.33.44.55:1035.  NDLB-force-rport and NDLB-connectile-dysfunction are enabled on the profile/endpoint.

FS responds to the request by rewriting the IP address to the received IP, but the port is not rewritten, so the packet flow looks like this:

22.33.44.55:1035 ------->  FS ---------> 22.33.44.55:5060

**Expected behavior**
The rport should be rewritten even if the endpoint specifies a port

 - Version [1.10.7]

yois615:
If NDLB-force-rport is enabled, we rewrite the rport regardless
of what the endpoint thinks the port should be.  We do this by
checking the status of server_rport_level before writing the port.

Fixes #1457


yois615:
@Len-PGH

briankwest:
You're actually breaking the entire purpose of this feature.  Your devices are capable of behaving correctly, and what your devices thinks, vs what the NAT in the middle thinks differ.

yois615:
@briankwest Just making sure we're both on the same page:

if (port) is checking if the end device is including port info, and if it does it will use it regardless of the rport.  That behavior is incorrect if NDLB is enabled.  We should not honor the device's incorrect port announcement, rather use rport.  Therefore, we add a check to see if the flag is set before committing that port number.

briankwest:
Can you outline your network topology?!?

yois615:
FS -> NAT Router -> Yealink device

Yealink sends
`Contact: sip:102@192.168.6.103:5060`

FS rewrites the IP but not the port

See #1457

briankwest:
You don't need to do this for a Yealink make that device do its job, STUN and what not and enable rport and things will just work.  Is a public IP on FS or is it behind aws or gcp?

yois615:
Public IP.

I agree the device is stupid; I'm trying to handle stupid devices. We honor rport regardless of what the device says.
Can you re-explain the problem you had in the first comment here?  I'm not imagining where this breaks anything. (and if it does I'll obviously drop the PR)

PR #1480, opened 2021-12-09
[core] extend count of acl entries (per SOFIA_MAX_ACL).

dragos-oancea:


dragos-oancea:
https://lists.freeswitch.org/pipermail/freeswitch-users/2021-December/135396.html

andywolk (approved):

src/switch_core.c
```diff
@@ -1405,7 +1405,7 @@ SWITCH_DECLARE(switch_bool_t) switch_check_network_list_ip_port_token(const char
 	} else if (strchr(list_name, '/')) {
 		if (strchr(list_name, ',')) {
 			char *list_name_dup = strdup(list_name);
-			char *argv[32];
+			char *argv[100]; /* MAX ACL */
```
nneul, line 1408: Is there a reason why we're hardcoding this at all? If it's speed sensitive - it should be getting cached. If it's not, just dynamically create the list?

dragos-oancea: it's already cached in switch_load_network_lists() . it's good to have bounds with user-provided cfg/data. only issue is that MAX_ACL should have been defined somewhere in the core, and perhaps increased. the original ACL code is from 2008, when the needs where different.

PR #1505, opened 2021-12-29
[mod_pgsql] Add support for PostgreSQL 14.x.

s3rj1k:
Signed-off-by: s3rj1k <evasive.gyron@gmail.com>

### DIFF
--- a/src/mod/databases/mod_pgsql/mod_pgsql.c
+++ b/src/mod/databases/mod_pgsql/mod_pgsql.c
@@ -779,6 +779,13 @@ switch_status_t pgsql_next_result_timed(switch_pgsql_handle_t *handle, switch_pg
 		case PGRES_PIPELINE_SYNC:
 			break;
 #endif
+//#if POSTGRESQL_MAJOR_VERSION >= 14
+		case PGRES_PIPELINE_SYNC:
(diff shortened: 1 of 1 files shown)


src/mod/databases/mod_pgsql/mod_pgsql.c
```diff
@@ -779,6 +779,13 @@ switch_status_t pgsql_next_result_timed(switch_pgsql_handle_t *handle, switch_pg
 		case PGRES_PIPELINE_SYNC:
 			break;
 #endif
+//#if POSTGRESQL_MAJOR_VERSION >= 14
```
mjerris, line 782: why is this commented out?

andywolk: why is `PGRES_PIPELINE_SYNC` mentioned twice?

src/mod/databases/mod_pgsql/mod_pgsql.c
```diff
@@ -779,6 +779,13 @@ switch_status_t pgsql_next_result_timed(switch_pgsql_handle_t *handle, switch_pg
 		case PGRES_PIPELINE_SYNC:
```
mjerris, line 779: shouldn't this be removed as its added back right below?  also seems that this doesn't require 14 if its already there so it shouldn't be in the #if

PR #1514, opened 2022-01-07
[core] fix cannot open error in switch_simple_email()

greenbea:


src/switch_utils.c
```diff
@@ -1288,6 +1288,9 @@ SWITCH_DECLARE(switch_bool_t) switch_simple_email(const char *to,
 				goto end;
 			}
 		}
+	} else {
+		err = "cannot open tmp file\n";
+		rval = SWITCH_FALSE;
```
mjerris, line 1293: shouldn’t there be a goto end here?

greenbea:
was closed by mistake

PR #1520, opened 2022-01-14
[core] Fix wrong reason code for group call

fx02:
Patch by Christian Berger/Peter Steinbach from FreeSWITCH-users mailing list


src/switch_ivr_originate.c
```diff
@@ -1795,7 +1795,13 @@ SWITCH_DECLARE(switch_status_t) switch_ivr_enterprise_originate(switch_core_sess
 		if (cancel_cause && *cancel_cause > 0) {
 			handles[i].cancel_cause = *cancel_cause;
 		} else {
-			handles[i].cancel_cause = SWITCH_CAUSE_LOSE_RACE;
+            if ((hp!=NULL) && (hp->cause==SWITCH_CAUSE_SUCCESS)) { //Was this call taken by another destination?
```
dragos-oancea, line 1798: it's wrongly indented. Also the comments should be C-style: /* */

fx02: let me know if its ok now

dragos-oancea:
commit message must say "[core]" instead of "[switch]"

dragos-oancea (approved):

andywolk:
> Patch by Christian Berger/Peter Steinbach from FreeSWITCH-users mailing list

@fx02 Can we have a link to the discussion on the mailing list?

src/switch_ivr_originate.c
andywolk: Are you sure checking `hp`'s cause is right here?

fx02:
https://lists.freeswitch.org/pipermail/freeswitch-users/2022-January/135477.html

gregoriusus:
I can confirm that this is working. We have this patch on production and it is working. What needs to be done to include this patch in main tree?

gregoriusus:
Guys, just checking this commit. What needs to be done, so that this commit would be merged. We build a solution with this patch and it is working. We are using this patch in production for quite some time now. We just do not want to do out build everytime. If I understand correctly, Andrey needs to review it?

src/switch_ivr_originate.c
```diff
@@ -1795,7 +1795,13 @@ SWITCH_DECLARE(switch_status_t) switch_ivr_enterprise_originate(switch_core_sess
 		if (cancel_cause && *cancel_cause > 0) {
 			handles[i].cancel_cause = *cancel_cause;
 		} else {
-			handles[i].cancel_cause = SWITCH_CAUSE_LOSE_RACE;
+			if ((hp!=NULL) && (hp->cause==SWITCH_CAUSE_SUCCESS)) { /* Was this call taken by another destination? */
```
andywolk, line 1798: Please address bad spaces.

andywolk (changes requested):

src/switch_ivr_originate.c
fx02: Not sure where the issue is, I guess after comment?

andywolk: How about some space?
```
if (hp != NULL && hp->cause == SWITCH_CAUSE_SUCCESS) {
```
instead of 
```
if ((hp!=NULL) && (hp->cause==SWITCH_CAUSE_SUCCESS)) {
```

fx02: oh, ok, fixed

dragos-oancea (approved):

andywolk (approved):

PR #1523, opened 2022-01-19
[core] Add support to phrase play-file to set a volume attribute

mochouinard:
This add support to specify volume="-10" in phrase action

the switch_file_handle only supported the ->vol flag which have a total of 8 volume correction position.
So I've added the ->volgranular which is based on the switch_resample code which is similar to the ->vol, but provide 100 volume setting in a much larger range of value.

### DIFF
--- a/src/include/switch_module_interfaces.h
+++ b/src/include/switch_module_interfaces.h
@@ -356,6 +356,7 @@ struct switch_file_handle {
 	switch_size_t samples_in;
 	switch_size_t samples_out;
 	int32_t vol;
+	int32_t volgranular;
 	switch_audio_resampler_t *resampler;
 	switch_buffer_t *buffer;
 	switch_byte_t *dbuf;
--- a/src/switch_ivr_play_say.c
+++ b/src/switch_ivr_play_say.c
@@ -251,7 +251,17 @@ SWITCH_DECLARE(switch_status_t) switch_ivr_phrase_macro_event(switch_core_sessio
 								  module_name);
 
 				if (!strcasecmp(func, "play-file")) {
-					status = switch_ivr_play_file(session, NULL, odata, args);
+					char *volume_str = (char *) switch_xml_attr_soft(action, "volume");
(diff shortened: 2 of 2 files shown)


anthmFS (commented):
Can't you just convert the existing vol to use the granular func instead of adding a different val?

mochouinard:
I was about to do that but realized it would cause issues with potential existing implementation.

If I did it, the people who implemented the ->vol will have different volume level value for number between 1 and 4 and -1 and -4, so it will affect the volume playback of people who upgrade. (And also change the behavior of 4 to -4 in term of volume control of the granular resample code)

I could probably do some kind of hack that the value of 5 is equal to 1 granular and use the same ->vol, but I though it would be cleaner to add a separate variable at this point.

mochouinard:
Or maybe I could create a new function that convert the value of the old vol to granular vol (which are not 100% perfect, but could be close enough), and replace everywhere in the code where the ->vol could be used.  It would require much more code change with more potential of things to break with existing install, for example : Maybe people use those function externally from FS where I can't see the code. So to be safe, I didn't go that route.

But if you prefer I change it so ->vol, I'll rewrite it this way !

Another option : If the goal is to make things cleaner in the long term, we could just add it like I did, plus changing everywhere in the code that use the ->vol, but add a warning in the log when ->vol isn't equal to 0 and notify them to use a specific function to change the value of the file descriptor.  So in a few years, we could remove the ->vol altogether and have a single variable which is only controlled via functions so make it easier to change how it implemented in the future (Maybe we switch to store variable to an event instead of adding lot of parameters to the file handler struct).

andywolk:
Okay. Just make the new one. Mark the old one deprecated, do not remove the old code. 
The last hunk that implements in ivr_play_file the switch_change_sln_volume_granular, we should move that up one block and make the old vol part an else/if so if both are set only the granular happens.

mochouinard:
Latest commit is based on the discussed changes.

anthmFS (approved):

PR #1525, opened 2022-01-21
[Core] Fix recording by clearing an unused bug info from the channel when recording is stopped

zuixuewosha:
@Description
    if record to same file and open the RECORD_APPEND, will show "Already recording xxxx",need clear the unused bug info if stop recording


andywolk:
So the question is why it is not done here https://github.com/signalwire/freeswitch/blob/966aa92ef0c214a9744f93d58df89896598f2698/src/switch_ivr_async.c#L1597

zuixuewosha:
sorry,I add some code caused this code not work in my repo,will close this request

PR #1531, opened 2022-01-27
fix(mod_conference): no audio output when client uses a=recvonly

### LINKED ISSUE #1437: mod_conference no audio output when smode is SWITCH_MEDIA_FLOW_SENDONLY
**Describe the bug**
conference_loop_output is not invoked when smode is SWITCH_MEDIA_FLOW_SENDONLY. I believe this logic in mod_conference is the problem.

		switch_media_flow_t audio_flow = switch_core_session_media_flow(session, SWITCH_MEDIA_TYPE_AUDIO);

		if (switch_channel_test_flag(channel, CF_AUDIO) && (audio_flow == SWITCH_MEDIA_FLOW_SENDRECV || audio_flow == SWITCH_MEDIA_FLOW_RECVONLY)) {
			conference_loop_output(&member);
		} else {

Originally, I believed that SWITCH_MEDIA_FLOW_RECVONLY should be SWITCH_MEDIA_FLOW_SENDONLY switch_core_session_media_flow reads the smode (send mode). 

I now believe we should ignore the audio_flow direction and invoke conference_loop_output regardless because it both starts the input thread and handles the output processing. For SENDONLY, we need output processing. For RECVONLY we need the input thread. For SENDRECV we need both. Alternatively someone could break up starting of input from output and be more intelligent about what to start.

**To Reproduce**
When setting up a call, pass in an sdp offer with recvonly. FreeSWITCH responds with an answer of sendonly, but never sends audio. 

**Expected behavior**
When FreeSWITCH responds with an answer of sendonly it should originate audio to the client.

FreeSWITCH Version 1.10.8-dev~64bit ( 64bit)

**Trace logs**
Problem seems to have been introduced with [this commit](https://github.com/signalwire/freeswitch/commit/badd59d609b5d5583c01d543e0d6b3a9cb365df1?diff=split
). Line 2481.

**backtrace from core file**
N/A

mariogasparoni:
We now call loop output for any direction
Closes #1437

### DIFF
--- a/src/mod/applications/mod_conference/mod_conference.c
+++ b/src/mod/applications/mod_conference/mod_conference.c
@@ -2477,9 +2477,7 @@ SWITCH_STANDARD_APP(conference_function)
 
 	/* Run the conference loop */
 	do {
-		switch_media_flow_t audio_flow = switch_core_session_media_flow(session, SWITCH_MEDIA_TYPE_AUDIO);
-		
-		if (switch_channel_test_flag(channel, CF_AUDIO) && (audio_flow == SWITCH_MEDIA_FLOW_SENDRECV || audio_flow == SWITCH_MEDIA_FLOW_RECVONLY)) {
+		if (switch_channel_test_flag(channel, CF_AUDIO)) {
 			conference_loop_output(&member);
 		} else {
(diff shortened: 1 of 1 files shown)


dragos-oancea (approved):

andywolk:
Closing this in favor of https://github.com/signalwire/freeswitch/pull/1568

PR #1540, opened 2022-02-03
[core] set completion cause upon failed recording

zayim:
Hi,

This PR sets recording completion cause for failures that previously did not have cause set.
Without this, it's impossible from CHANNEL_EXECUTE_COMPLETE event to distinguish if `record_session` ended without failure or did not even start because of failure.

I tested it and it worked fine. Thank you.


dragos-oancea (approved):

zayim:
@dragos-oancea I see this PR is now OK for merge, can you merge it when you have time? Thanks!

andywolk (approved):

PR #1574, opened 2022-03-14
mod_verto: add bind presence in/probe events

wmasilva:
bind to events PRESENCE_IN and PRESENCE_PROBE

This allow to send custom presence IN events from an esl script to subscribe channels.
Also when subscribe to a channel detect if is registered or not.


wmasilva:
the default behaviour is monitoring only the active calls for the subscribe channel and is not affected by this commit.
This add the possibility to get the registered status and as well any presence event for channel.

wmasilva:
ok, thanks.

PR #1584, opened 2022-03-20
[mod_callcenter] Fix SWITCH_EVENT_PRESENCE_PROBE unbind

greenbea:
`globals` was getting initialized to null after `switch_event_bind_removable` binding causing `globals.nodes` to get set to NULL causing `switch_event_unbind` not to be able to unbind the `SWITCH_EVENT_PRESENCE_PROBE` event when unloading the module. Besides not being able to unbind the event, this was also causing a segfault  upon receiving a `PRESENCE_PROBE` event when unloading or reloading the module


greenbea:
was caused by this commit https://github.com/signalwire/freeswitch/commit/647ed310e54ce8bcab683ea6d37fbd1a9a22577b

dragos-oancea (approved):

PR #1603, opened 2022-03-29
set rtp flag for rtcp events after flags zeroed out

rjm2718:
SEND_RTCP_MESSAGE events not fired even when fire_rtcp_events channel variable is set, because SWITCH_RTP_FLAG_AUDIO_FIRE_SEND_RTCP_EVENT rtp flag is not set in the right place.

Code was introduced in v1.10.7 via https://github.com/signalwire/freeswitch/pull/698
Unit test didn't catch this problem because switch_core_media_set_rtp_flag() sets that rtp flag explicitly.

In src/switch_core_media.c, rtp flags
SWITCH_RTP_FLAG_AUDIO_FIRE_SEND_RTCP_EVENT and
SWITCH_RTP_FLAG_VIDEO_FIRE_SEND_RTCP_EVENT
 are set early in switch_core_media_activate_rtp(), before
`memset(flags, 0, sizeof(flags))`
is used (in different places depending if session will be audio or video).

Therefore setting of those flags must happen in separate places, after the flags init/memset statements.

### DIFF
--- a/src/switch_core_media.c
+++ b/src/switch_core_media.c
@@ -8681,13 +8681,6 @@ SWITCH_DECLARE(switch_status_t) switch_core_media_activate_rtp(switch_core_sessi
 		goto end;
 	}
 
-	if (switch_channel_var_true(session->channel, "fire_rtcp_events")) {
-		flags[SWITCH_RTP_FLAG_AUDIO_FIRE_SEND_RTCP_EVENT] = 1;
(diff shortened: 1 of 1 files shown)


dragos-oancea:
we'll remove the memset.

dragos-oancea:
30d5d58e6e243fad31b27a1559f644c68f242018

dragos-oancea:
thanks.

PR #1626, opened 2022-04-11
[mod_sofia] refine CLI "sofia global" usage messages

dilyanpalauzov:


dilyanpalauzov:
I do not understand the difference in the help output on `sofia<ENTER>` between `<x>` and `[x]`.  What is optional what, is mandatory?  Or rather, what must be used literally, what is a placeholder needing substitution?

dilyanpalauzov:
E.g. here
> "-ERR Usage: siptrace <on|off>|capture <on|off>|watchdog <on|off>|debug <sla|presence|none>| standby <on|off>"

In practice the parameter after `debug` is optional, after the other global-commands it is not optional.  This is not evident from the "-ERR Usage:" message.

PR #1633, opened 2022-04-16
[mod_sofia] remove leftovers from stun-auto-disable and stun-enabled

dilyanpalauzov:
These STUN parameters disappeared in commit 0fffc839e4a7c5a226572f.


dilyanpalauzov:
Can somebody explain me, which test does fail because of this change, and how can I find it at https://public-artifacts.signalwire.cloud/drone/signalwire/freeswitch/456/log_run-tests_src!mod!endpoints!mod_sofia!test!test_run_sipp.sh.html ?

PR #1652, opened 2022-05-11
[mod_curl] The Reason-Phrase can be a empty string

GangZhuo:
See [RFC-2616 Section 6.1](https://datatracker.ietf.org/doc/html/rfc2616#section-6.1)
```
Status-Line = HTTP-Version SP Status-Code SP Reason-Phrase CRLF
Reason-Phrase  = *<TEXT, excluding CR, LF>
```


GangZhuo:
Sorry, maybe dup with https://github.com/signalwire/freeswitch/pull/1556/files

PR #1669, opened 2022-05-26
[mod_shout] Replace space with tab in Makefile.am

Clock00Smith:
Space indent will cause a error when running ./configure when HAVE_MP3LAME is false.


briankwest:
Commit message needs to match the PR description, please update.

PR #1672, opened 2022-06-01
[core] Add loop with waitpid to avoid zombies process

budocay:
Hello Dear,

We observed zombie process creation during use of fs_ivrd when many call are made.

Indeed a lot of zombies process are created and never clean. So we investigate and saw you were using wait without loop on it.

So we change wait to waitpid with WNOHANG flag and we added loop in order to be sure all son process are ended.

If you have any questions be sure I'll be there

Regards


libs/esl/ivrd.c
```diff
@@ -41,8 +41,12 @@
 static void handle_SIGCHLD(int sig)
 {
 	int status = 0;
+	int pid = 0;
+
+	do {
+		pid = waitpid(-1, &status, WNOHANG);
+	} while (pid > 0);
```
andywolk, line 48: Is this going to burn CPU? Do we need a small sleep?

budocay: No, there is no CPU burning because this function is only called when SIGCHILD is found and with WNOHANG flag we prevent from blocking process, it will return current process status and if there is an error it will return immediatly this error (PID .

We already test on our platform this patch (3000 simultaneous call) and no CPU burning was found with our monitoring

budocay:
Up :)

PR #1675, opened 2022-06-02
mod_sms: add app transfer

wmasilva:
Allow to transfer chat messages between contexts. 
Simple use case:
<context name="default">
	<extension name="customerX">
		<condition field="to" expression="^.*$">
			<action application="transfer" data="customerX"/>
		</condition>
	</extension>
</context>

<context name="customerX">
	<extension name="200">
		<condition field="to" expression=".*">
			<action application="reply" data="Hello, you said: ${_body}"/>
		</condition>
	</extension>
</context>

### DIFF
--- a/src/mod/applications/mod_sms/mod_sms.c
+++ b/src/mod/applications/mod_sms/mod_sms.c
@@ -610,6 +610,30 @@ SWITCH_STANDARD_CHAT_APP(reply_function)
 	return SWITCH_STATUS_SUCCESS;
 }
 
+
+SWITCH_STANDARD_CHAT_APP(transfer_function)
+{
+	char *context;
+
+	if (!data) return SWITCH_STATUS_SUCCESS;
+
+	context = strdup(data);
+	if (!context) return SWITCH_STATUS_SUCCESS;
+
+	switch_log_printf(SWITCH_CHANNEL_LOG, SWITCH_LOG_DEBUG, "SMS transfer to %s \n", data);
+
+	switch_event_del_header(message, "context");
+	switch_event_add_header_string(message, SWITCH_STACK_BOTTOM, "context", context);
+
+	switch_core_chat_send("GLOBAL", message);
+
+
+	return SWITCH_STATUS_SUCCESS;
+}
+
+
+
+
 /* Macro expands to: switch_status_t mod_sms_load(switch_loadable_module_interface_t **module_interface, switch_memory_pool_t *pool) */
 SWITCH_MODULE_LOAD_FUNCTION(mod_sms_load)
 {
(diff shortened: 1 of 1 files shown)


src/mod/applications/mod_sms/mod_sms.c
```diff
@@ -610,6 +610,30 @@ SWITCH_STANDARD_CHAT_APP(reply_function)
...
 }
 
+
+SWITCH_STANDARD_CHAT_APP(transfer_function)
+{
+	char *context;
+
+	if (!data) return SWITCH_STATUS_SUCCESS;
+
+	context = strdup(data);
```
andywolk, line 620: Where is context freed?

wmasilva: is not... my bad.  fixing it now.

thanks

src/mod/applications/mod_sms/mod_sms.c
```diff
@@ -610,6 +610,30 @@ SWITCH_STANDARD_CHAT_APP(reply_function)
 	return SWITCH_STATUS_SUCCESS;
 }
 
+
+SWITCH_STANDARD_CHAT_APP(transfer_function)
+{
+	char *context;
+
+	if (!data) return SWITCH_STATUS_SUCCESS;
```
dragos-oancea, line 618: is it really supposed to return SWITCH_STATUS_SUCCESS if no data ? it seems wrong in multiple places in this file.

wmasilva: not sure, i guess the correct way is print to user the missing data and return failure.  
and If failure it should stop executing the next actions, no?

src/mod/applications/mod_sms/mod_sms.c
```diff
@@ -610,6 +610,30 @@ SWITCH_STANDARD_CHAT_APP(reply_function)
...
+	if (!context) return SWITCH_STATUS_SUCCESS;
+
+	switch_log_printf(SWITCH_CHANNEL_LOG, SWITCH_LOG_DEBUG, "SMS transfer to %s \n", data);
+
+	switch_event_del_header(message, "context");
+	switch_event_add_header_string(message, SWITCH_STACK_BOTTOM, "context", context);
+
+	switch_core_chat_send("GLOBAL", message);
+
+
```
briankwest: Extra lines

src/mod/applications/mod_sms/mod_sms.c
```diff
@@ -610,6 +610,30 @@ SWITCH_STANDARD_CHAT_APP(reply_function)
...
+	return SWITCH_STATUS_SUCCESS;
+}
+
+
```
briankwest, line 634: Extra Lines

PR #1689, opened 2022-06-13
When using 'as json', make sure response is terminated

xrobau:
Super trivial bug I noticed when using 'as json', but it's a bug 8)

There is no \n automatically appended to a JSON response, so
there wasn't any way to determine the result was completed.
You can see the 'rows:0' result below has a \n, and that worked
correctly.

Signed-Off-By: Rob Thomas <xrobau@gmail.com>


mjerris:
why would you need a \n to know it’s complete?  you get the entire thing as one message, it can’t fragment, and if it was truncated you would know due to malformed json

xrobau:
Yep, Sorry @mjerris you're right - I had misread the API, and have now realised I was wrong to be expecting a \n, as it's not required!

PR #1691, opened 2022-06-15
[mod_amqp] Added xml handler via amqp

alexey-khabulyak:
Added ability to retrieve XML config via mod_amqp

mod_amqp creates one connection for sending requests and multiple connections for retrieving responses(had to do it because librabbitmq-dev doesn't allow to share connection between threads)

When fs needs xml config, it takes a free connection from a pool(or creates a new one)
Then it sends a request via main connection and waits for a response via  auxiliary connection. 
After fs receives XML it releases auxiliary connection, so next thread can use it now.

image

Backend example:
```
import pika, sys, os, json

BODY='''<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<document type="freeswitch/xml">
  <section name="directory" description="">
    <domain name="default">
      <groups>
        <group name="default" description="">
          <users>
            <user id="1000">
              <variables>
                <variable name="asdf" value="true"/>
              </variables>
            </user>
          </users>
        </group>
      </groups>
    </domain>
  </section>
</document>
'''
def main():
    credentials = pika.PlainCredentials('guest', 'guest')
    parameters = pika.ConnectionParameters(credentials=credentials, host='127.0.0.1')
    connection = pika.BlockingConnection(parameters)
    channel = connection.channel()


    def callback(ch, method, props, body):
        print(" [x] Received %r" % body)
        print(" [x] props %r" % props)
        channel.basic_publish(
                exchange="",
                routing_key=props.reply_to,
                properties=pika.BasicProperties(correlation_id = props.correlation_id),
                body=BODY)

    channel.basic_consume(queue='test_amqp', on_message_callback=callback, auto_ack=True)

    print(' [*] Waiting for messages. To exit press CTRL+C')
    channel.start_consuming()

if __name__ == '__main__':
    try:
        main()
    except KeyboardInterrupt:
        print('Interrupted')
        try:
            sys.exit(0)
        except SystemExit:
            os._exit(0)
```

image

### DIFF
--- a/src/include/switch_core.h
+++ b/src/include/switch_core.h
@@ -147,7 +147,7 @@ typedef void(*switch_device_state_function_t)(switch_core_session_t *session, sw
 
 #define DTLS_SRTP_FNAME "dtls-srtp"
 #define MAX_FPLEN 64
-#define MAX_FPSTRLEN 192
+#define MAX_FPSTRLEN 193
 
 typedef struct dtls_fp_s {
 	uint32_t len;
--- a/src/switch_rtp.c
+++ b/src/switch_rtp.c
@@ -3233,8 +3233,13 @@ static int dtls_state_setup(switch_rtp_t *rtp_session, switch_dtls_t *dtls)
 	if ((dtls->type & DTLS_TYPE_SERVER)) {
 		r = 1;
 	} else if ((cert = SSL_get_peer_certificate(dtls->ssl))) {
-		switch_core_cert_extract_fingerprint(cert, dtls->remote_fp);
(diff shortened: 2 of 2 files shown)


crienzo (changes requested):
Fix scan-build errors.

alexey-khabulyak:
@crienzo 
i've fixed unit-tests issues, made some code refactoring and added description how it works.

alexey-khabulyak:
@crienzo hi! I've fixed all bugs and memory leaks. also, i added description how it works and an example.

PR #1693, opened 2022-06-16
Add Fanvil to vendors allowed to update.

markjcrane:
Problem:
Pickup calls from Valet Park shows feature code rather than the Caller's original Caller ID.

Solution:
Add Fanvil to the update_allowed condition resolves this problem.

This is a useful improvement for people using Fanvil with FreeSWITCH.


kidjake28:
This would be very helpful to us Fanvil users. Please accept pull request.

andywolk (approved):

Issue #1698, opened 2022-06-22
uuid_transfer will cause the channel to remain in the CS_RESET state


mingfireyu:
**Describe the bug**
if Aleg is bridged with Bleg, then we call api
uuid_transfer(Bleg_uuid, -both conference:conference_name inline) to transfer  A-leg and B-leg to conference.
Because of mutex problem , Bleg may be remain in the CS_RESTE state.

**Code Problem**

<img width="1462" alt="image" src="https://user-images.githubusercontent.com/20698955/175004511-bcdbe7b1-efd4-4ca7-b84c-360840c3e0de.png">
In function `switch_ivr_session_transfer` , api uuid_transfer will set Bleg channel state to CS_ROUTING.
But before doing Bleg transfer, switch_ivr_session_transfer will first transfer Aleg to "conference" application. Aleg will unbrige with Bleg now. So Bleg session thread will do this in function `switch_ivr_multi_threaded_bridge`:
<img width="1426" alt="image" src="https://user-images.githubusercontent.com/20698955/175005125-42da59cd-e582-401b-b040-bbe3bbfbaf65.png">
Since switch_channel_get_state is not execute with lock, it may get a  state that is not CS_ROUTING, and then other thread (thread execute uuid_transfer) switch_ivr_session_transfer set Bleg to CS_ROUTING,  then Bleg session thread set channel state to CS_RESET. 
Now, no thread can set Bleg to CS_ROUTING.

**Expected behavior**
I think we should make 
```c
state = switch_channel_get_state(caller_channel);
if (!(state == CS_RESET || state == CS_PARK || state == CS_ROUTING || state == CS_EXECUTE)) {
       switch_channel_set_state(caller_channel, CS_RESET);
}
```
to be in a transaction with lock . So other thread modify channel state can be saw by session thread.

**Package version or git hash**
 - master branch

PR #1718, opened 2022-07-14
[core] When unbridge, set the channel state atomically

mingfireyu:
for issue https://github.com/signalwire/freeswitch/issues/1698

after we get channel state at line 1995 in switch_ivr_bridge.c, it may be modified by other thread, so line 1956 "if judgement" maybe invalid.


mingfireyu:
personal issue

PR #1736, opened 2022-07-29
Added global-db-lock setting in mod_sofia

sharmilas2704:
**Purpose**
This fixes the handling of global-db-lock in mod_sofia has configurable.

**Overview**
As the Freeswitch use global DB locks to ensure a module will execute just one query at a time creates a bottleneck and it is required only for sqlite , not required for Mariadb or postgresql as it can handle multiple concurrent queries. To overcome this issue added and enabled the global-db-lock settings for mutex lock conditions in sofia_glue.c file.

**Enhancements**
1.To make the code backward compatibility added a new setting global-db-lock in suitable files to enable or disable mutex lock and unlock.
2.Added a new header for global-db-lock.

**Problems faced**
Creates bottleneck

Changes done
1.Modified sofia_glue.c file mutex conditions with enabling global_db_lock settings.
2.Added a header for global-db-lock in mod_sofia.h.

### DIFF
--- a/conf/vanilla/autoload_configs/sofia.conf.xml
+++ b/conf/vanilla/autoload_configs/sofia.conf.xml
@@ -5,6 +5,7 @@
     <!-- <param name="abort-on-empty-external-ip" value="true"/> -->
     <!-- <param name="auto-restart" value="false"/> -->
     <param name="debug-presence" value="0"/>
+    <!-- <param name="global-db-lock" value="0"/>-->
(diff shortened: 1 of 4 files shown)


mjerris:
this for sure should not be a config setting… this should come from api knowing if we need it or not based on different engines.  also must ensure there isnt anything we’ve done that assumes ordering as a side effect of these locks

PR #1740, opened 2022-08-03
[mod_sofia] Add db-spin-up-wait-ms profile parameter.

ukolovda:


ukolovda:
Delay 1000 milliseconds for each profile is too much for current computers. I tested with 100 ms. Start time reduce from 215 to 35 seconds (200 profiles).

src/mod/endpoints/mod_sofia/sofia.c
```diff
@@ -6351,7 +6354,11 @@ switch_status_t config_sofia(sofia_config_t reload, char *profile_name)
 						launch_sofia_profile_thread(profile);
 						if (profile->odbc_dsn) {
 							switch_log_printf(SWITCH_CHANNEL_LOG, SWITCH_LOG_NOTICE, "Connecting ODBC Profile %s [%s]\n", profile->name, url);
-							switch_yield(1000000);
+							start_delay = profile->odbc_start_delay;
+							if (start_delay == 0) {
+							  start_delay = 1000;
```
andywolk, line 6359: We use tabs and not 2 spaces. Please correct.

src/mod/endpoints/mod_sofia/sofia.c
```diff
@@ -6351,7 +6354,11 @@ switch_status_t config_sofia(sofia_config_t reload, char *profile_name)
...
+							}
+							switch_yield(start_delay * 1000);
```
andywolk, line 6361: Please make an empty line after an `if` block. https://github.com/signalwire/libks/blob/master/CODING_GUIDELINES#L36

src/mod/endpoints/mod_sofia/sofia.c
```diff
@@ -4788,6 +4789,8 @@ switch_status_t config_sofia(sofia_config_t reload, char *profile_name)
 						}
 					} else if (!strcasecmp(var, "odbc-dsn") && !zstr(val)) {
 						profile->odbc_dsn = switch_core_strdup(profile->pool, val);
+					} else if (!strcasecmp(var, "odbc-start-delay") && !zstr(val)) {
+					  profile->odbc_start_delay = atoi(val);
```
andywolk, line 4793: Please fix spacing.

src/mod/endpoints/mod_sofia/sofia.c
```diff
@@ -4788,6 +4789,8 @@ switch_status_t config_sofia(sofia_config_t reload, char *profile_name)
 						}
 					} else if (!strcasecmp(var, "odbc-dsn") && !zstr(val)) {
 						profile->odbc_dsn = switch_core_strdup(profile->pool, val);
+					} else if (!strcasecmp(var, "odbc-start-delay") && !zstr(val)) {
```
andywolk, line 4792: Please edit vanila conf putting that param in a commented out form. Give it a comment about defaults.

andywolk (changes requested):
Thanks for the PR. There are a few minor things we need to improve. That should not be hard for you.

ukolovda:
@andywolk  I make fixes. Please, check.

I don't sure about parameter name `odbc-start-delay`. May be we should name it `odbc-after-start-delay`, what do you think?

andywolk:
> @andywolk I make fixes. Please, check.
> 
> I don't sure about parameter name `odbc-start-delay`. May be we should name it `odbc-after-start-delay`, what do you think?

`db-spin-up-wait-ms` may be?

andywolk:
```
<!-- Default delay after starting profile with odbc-dsn is 1000 milliseconds, you can wish reduce it to 100 or 200 ms -->
```
does not sound right. Let's try
```
<!-- By default each profile will give the database 1000 ms to spin-up on load. -->
```

ukolovda:
@andywolk  Fixed parameter name and comment.
Also I modify code for make simpler.

andywolk (approved):

PR #1784, opened 2022-08-30
[mod_kazoo] Fix building (map to list changes)

fx02:


lazedo:
@fx02 what’s the purpose of this?

fx02:
For me, mod_kazoo will not build without this fix.

lazedo:
What otp version are you using? JSON encoding is done with tupple or map. You need a otp version that supports maps

fx02:
You mean this? erlang-otp_mibs-R16B-03.18.el7.x86_64

lazedo:
You need an otp version that supports maps, at least otp17.

fx02:
thanks, will look into that

Issue #1481, opened 2021-12-10
Too long Reason header swallowed


dcw67:
**Describe the bug**

A BYE message with a Reason: header that is longer than 128 characters get responded to on the B leg with a 200 OK as expected but is not forwarded across a bridge back to the A leg.

**To Reproduce**

Send a BYE packet containing a Reason head with a long text description. For example:

BYE sip:mod_sofia@103.26.173.43:5060 SIP/2.0
Via: SIP/2.0/UDP 139.162.48.4:5060;branch=z9hG4bK-10866-1-6
Route: <sip:103.26.173.62;lr;ftag=9ra9N222HeQ8e>
From: <sip:617xxxxxxxx@139.162.48.4>;tag=10866SIPpTag011
To: "614xxxxxxxx" <sip:614xxxxxxxx@103.26.173.43>;tag=9ra9N222HeQ8e
Call-ID: 690b9544-d34d-123a-a98a-b2ad0b93b34a
CSeq: 2 BYE
Contact: sip:sipp@139.162.48.4:5060
Max-Forwards: 70
Content-Length: 0
Reason: Q.850;cause=16;text="The call has ended because there is no longer any parties left in the call. This is an exceptionally long reason header and will probably break things"
User-Agent: SONUS SBCSWeLite 9.0.4v316 Azure Ribbon

**Expected behaviour**
Expected the BYE to pass back through the bridge to the A leg. This does not occur resulting in the A leg remaining up until the A leg hangs up.

**Package version or git hash**

FreeSWITCH Version 1.10.7-release-19-883d2cb662~64bit (-release-19-883d2cb662 64bit)

**Trace logs**
FS log and PCAPS attached

Network topology is:

offnet ->   routing node
routing_node -> FS SBC (2 x interfaces)
FS SBC -> proxy
proxy -> CPE

In attached PCAPS:

offnet -> 192.168.153.38
192.168.153.38 -> 192.168.153.39 (private interface)
103.26.173.42 (public interface) -> 103.26.173.62
103.26.173.62 -> 139.162.48.4 

In LongReasonHeaderCustomerLeg.pcap the BYE arrives at SBC
In LongReasonHeaderInternalLeg.pcap te BYE is not sent to the routing node

[LongReasonHeader.zip](https://github.com/signalwire/freeswitch/files/7690282/LongReasonHeader.zip)

**backtrace from core file**
n/a

gilles0606:
Hello, has anyone made a patch? 
[...]

Issue #1293, opened 2021-08-05
A leg not being closed down when B ends call


mase2hot:
Hi All

I have an issue where the BYE is not being sent to the A leg once the B leg closes down the call. Now this only happens on calls where there were multiple re-invites caused by the B leg putting the far end on hold. When this doesn't take place there is only 1 re-invite from the B leg and the call closes down fine.

This is happening 99% of the time, I managed to grab a good and bad call. I can not see any obvious errors and FS is reporting that it will be sending a BYE to the A leg, but it does not. Please see attached debug from FS.

From what I can see on the call that fails to send the BYE, there are some state change extra things going on, no idea why?

Package version or git hash
 - Version 1.10.6

processor       : 23
vendor_id       : GenuineIntel
cpu family      : 6
model           : 63
model name      : Intel(R) Xeon(R) CPU E5-2620 v3 @ 2.40GHz
stepping        : 2
microcode       : 0x31
cpu MHz         : 1201.312
cache size      : 15360 KB
physical id     : 1
siblings        : 12
core id         : 5
cpu cores       : 6
apicid          : 27
initial apicid  : 27
fpu             : yes
fpu_exception   : yes
cpuid level     : 15
wp              : yes
flags           : fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush dts acpi mmx fxsr sse sse2 ss ht tm pbe syscall nx pdpe1gb rdtscp lm constant_tsc arch_perfmon pebs bts rep_good nopl xtopology nonstop_tsc cpuid aperfmperf pni pclmulqdq dtes64 monitor ds_cpl vmx smx est tm2 ssse3 sdbg fma cx16 xtpr pdcm pcid dca sse4_1 sse4_2 x2apic movbe popcnt aes xsave avx f16c rdrand lahf_lm abm cpuid_fault epb invpcid_single pti tpr_shadow vnmi flexpriority ept vpid ept_ad fsgsbase tsc_adjust bmi1 avx2 smep bmi2 erms invpcid cqm xsaveopt cqm_llc cqm_occup_llc dtherm ida arat pln pts
bugs            : cpu_meltdown spectre_v1 spectre_v2 spec_store_bypass l1tf mds swapgs itlb_multihit
bogomips        : 4800.08
clflush size    : 64
cache_alignment : 64
address sizes   : 46 bits physical, 48 bits virtual
power management:
root@FS-4:~# uname -a
[...]

PR #1789, opened 2022-09-02
if the reason text field received is too long, the field is no longer correct.

gilles0606:
if the reason text field received is too long, the field is no longer correct.
Too long Reason header swallowed #1481
 A leg not being closed down when B ends call #1293

### DIFF
--- a/src/mod/endpoints/mod_sofia/sofia.c
+++ b/src/mod/endpoints/mod_sofia/sofia.c
@@ -8576,7 +8576,11 @@ static void sofia_handle_sip_i_state(switch_core_session_t *session, int status,
 				if (!tech_pvt) goto done;
 
 				if (tech_pvt->mparams.num_codecs) {
-					match = sofia_media_negotiate_sdp(session, r_sdp, SDP_TYPE_RESPONSE);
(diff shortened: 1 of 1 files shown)


briankwest:
I think this needs to be fixed in libsofia, nea_server.c reason_buf is 64, and not in mod_sofia.

gilles0606:
i Brian,

Or else I can do the test myself.
what do you recommend? replace 64 with 128?
or 256? I see that some SBC sometimes can exceed 128 in the text cause.

gilles0606:
Brian, 

sorry for my not very good english.
I tried with 256 in nea_server.c reason_buf but that didn't fix the problem.

we are exactly in line 501 of mod_sofia.c
the complete sip header: "reason" is copied into a memory of 128.
(char reason[128] = "";)

if this header is truncated, it is no longer correct when the ending quote is missing.

if we enlarge the (char reason[256] = "";) it works, but the next freeswitch that receives the call will have the same problem.

that's why I made this "pull request"

of course, you know freeswitch thousands of times better than me, that's why I'm explaining to you what I saw and understood, but maybe I'm wrong...

thank you very much for your opinion

PR #1791, opened 2022-09-02
[mod_sofia] Fix sip_invite_to_uri without host

greenbea:
`sofia_outgoing_channel()` sets the `tech_pvt->dest_to` and handles if sip_invite_to_uri doesn't have the host part in it, but that was being overwritten in `sofia_glue_do_invite` that sets the to_str to the sip_invite_to_uri and overwrites the `tech_pvt->dest_to`

technically the whole ternary isn't needed as `tech_pvt->dest_to` is already being set in `sofia_outgoing_channel` and `tech_pvt->dest_to` will have the value of `sip_invite_to_uri` if it's set on the channel


greenbea:
was closed by mistake

PR #1796, opened 2022-09-07
 fix socket bind error under windows 

makefriend8:
Function esl_listen_threaded cannot start correctly under windows. It lacked function WSAStartup.


seven1240:
can you squash into one commit?

makefriend8:
> can you squash into one commit?

Let me try.

PR #1797, opened 2022-09-07
Fixed remote SSRC

DoduRH:
Changed `atol` to `atoll` to correctly read SSRC values above `2,147,483,647`.

According to the spec, SSRC is a 32-bit unsigned integer.  `atol` only supports $\pm$ `2,147,483,647` meaning when the client sends an SSRC above this, behaviour is undefined.

Spec:
https://www.rfc-editor.org/rfc/rfc3550.html#section-2.4


andywolk:
@DoduRH Please take a look if that's the only example in that file.

Issue #1809, opened 2022-09-18
fs_cli hang on debian 11 whith stable v1.10.7 (reproducible) 


gilles0606:
Hello, On a brand new Debian 11, I don't install any packages during install. 

Then I install the following packages :
apt-get install 
autoconf libtool pkg-config libcurl4-openssl-dev uuid-dev libssl-dev build-essential libtool-bin unzip libsqlite3-dev 
libpcre3-dev libspeex-dev libspeexdsp-dev libedit-dev python3-dev libsndfile-dev liblua5.2-dev libsqlite3-dev
yasm zip libtiff-dev vim net-tools ssh

Using git I take the sources of version v1.10.7 Then I take the sources of sofia and spandsp

cd sofia-sip
./bootstrap.sh
./configure
make
make install

cd spandsp-master            
./bootstrap.sh               
./configure                  
make                         
make install                 

cd freeswitch
./bootstrap.sh -j

(cp modules.conf freeswitch)

./configure
make
make install

LD_LIBRARY_PATH=/usr/local/lib
export LD_LIBRARY_PATH

and run :

/usr/local/freeswitch/bin/freeswitch -nonat -u root

I don't do anything else, I keep the default config, I don't change anything. Just compiling...

After launching freeswitch, I connect with fs_cli

After the connection, hat the first command: show calls for example
randomly, 50% of time, fs_cli freezes and becomes unresponsive.
I did it, on several different servers
 

here are the weirdest details... :

For fs_cli hanging, **there must be no activity on freeswitch.**
I run an fs_cli and it freezes. 
I run on a second console fs_cli and it freezes.
I launch another fs_cli which does not freeze… and I do a “log 1 hello”
**This makes a trace, and all the fs_cli unfreeze….**

With gdb I saw, fs_cli hangs in mutex locks

I reproduce this on several debian 11.5, but the same source files compiling and work fine on debian 10.13

Do you have any ideas ??

Thanks 

Additional Information 
If anyone wants to reproduce. 
Here is the list of my module.conf

applications/mod_commands
applications/mod_conference
applications/mod_distributor
applications/mod_dptools
applications/mod_esf
applications/mod_expr
applications/mod_hash
applications/mod_spandsp
applications/mod_test
dialplans/mod_dialplan_xml
endpoints/mod_loopback
endpoints/mod_sofia
event_handlers/mod_cdr_csv
event_handlers/mod_event_socket
formats/mod_local_stream
formats/mod_native_file
[...]

PR #1812, opened 2022-09-24
Change Thread Create location in fs_cli.c

gilles0606:
Move Thread Create after all direct esl_send_recv.  avoid deadlocks.
see :  fs_cli hang on debian 11 whith stable v1.10.7 (reproducible) #1809

### DIFF
--- a/libs/esl/fs_cli.c
+++ b/libs/esl/fs_cli.c
@@ -1732,12 +1732,6 @@ int main(int argc, char *argv[])
 	global_handle = &handle;
 	global_profile = profile;
 
-	if (esl_thread_create_detached(msg_thread_run, &handle) != ESL_SUCCESS) {
-		printf("Error starting thread!\n");
-		esl_disconnect(&handle);
-		return 0;
-	}
-
 	snprintf(cmd_str, sizeof(cmd_str), "api switchname\n\n");
 	esl_send_recv(global_handle, cmd_str);
 	if (global_handle->last_sr_event && global_handle->last_sr_event->body) {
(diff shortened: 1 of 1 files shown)


andywolk:
Please show all troubleshooting info so we could analyze this.

gilles0606:
hello andrey.

https://github.com/signalwire/freeswitch/issues/1809

You can reproduce the problem with a Debian 11.

In fact, there is a deadlock between the ESL event processing routine, and the fact of sending commands directly through the same channel.

After setting up the processing routine (esl_thread_create_detached)
The program sends commands directly, which creates unwanted locks.

I moved the esl_thread_create_detached routine after the direct commands to avoid conflicts.

Conflicts do not trigger in Debian 10, but appear in Debian 11.

This is what I put in my report

After the connection, hat the first command: show calls for example
randomly, 50% of time, fs_cli freezes and becomes unresponsive.
I did it, on several different servers
 
here are the weirdest details... :
For fs_cli hanging, there must be no activity on freeswitch.
I run an fs_cli and it freezes.
I run on a second console fs_cli and it freezes.
I launch another fs_cli which does not freeze… and I do a “log 1 hello”
This makes a trace, and all the fs_cli unfreeze….
With gdb I saw, fs_cli hangs in mutex locks
I reproduce this on several debian 11.5, but the same source files compiling and work fine on debian 10.13

I kept a virtual server with the problem, tell me what you need I can make a trace under gdb?

tell me what you need and i will do it quickly..

Otherwise, for my change, here is the explanation:

fs_cli launches three direct commands at startup:
- api switchname
- api hostname
-log

I moved the esl_thread_create_detached after the last direct command.

Thanks...

jerry6290 (approved):

jerry6290:
> Please show all troubleshooting info so we could analyze this.

Please review

megrooms:
I'm seeing this problem happen frequently on Debian 13 which is quite frustrating. Applying this patch resolves the issue.

PR #1838, opened 2022-10-24
[Build-System] Disable static build by default on Linux.

andywolk:


mjerris (approved):

MarioG-X:
I already tested that (see #1840) and it builds fine.

Issue #1476, opened 2021-12-06
Unable to hear MOH, after uuid_hold


GangZhuo:
**Describe the bug**
A call from 1000 to 1001, when execute uuid_hold on bleg, unable hear MOH on aleg, but when uuid_hold is executed on aleg, MOH can be hear on bleg.

**To Reproduce**
1. Using default configurations
2. Dial from 1000 to 1001
3. Execute uuid_hold <bleg uuid>
4. Should be hear MOH on phone 1000

**Expected behavior**
Hold on either legs can be heard MOH.

**Package version or git hash**
 - FreeSWITCH version: 1.10.8-dev+git~20211205T095551Z~e632759ae8~64bit (git e632759 2021-12-05 09:55:51Z 64bit)

**Trace logs**
```
2021-12-06 17:10:36.921187 99.03% [NOTICE] switch_channel.c:1123 New Channel sofia/internal/1000@192.168.56.21 [9995f03c-5fb1-4ac1-8ff5-ba9a48ab377b]
2021-12-06 17:10:36.921187 99.03% [INFO] sofia.c:10462 sofia/internal/1000@192.168.56.21 receiving invite from 192.168.56.1:56474 version: 1.10.8-dev git e632759 2021-12-05 09:55:51Z 64bit call-id: 17f998546b414470bf8ac785dcca759c
2021-12-06 17:10:36.939192 99.03% [INFO] sofia.c:10462 sofia/internal/1000@192.168.56.21 receiving invite from 192.168.56.1:56474 version: 1.10.8-dev git e632759 2021-12-05 09:55:51Z 64bit call-id: 17f998546b414470bf8ac785dcca759c
2021-12-06 17:10:36.958862 99.03% [INFO] mod_dialplan_xml.c:639 Processing 1000 <1000>->1001 in context default
EXECUTE [depth=0] sofia/internal/1000@192.168.56.21 set(open=true)
EXECUTE [depth=0] sofia/internal/1000@192.168.56.21 hash(insert/192.168.56.21-spymap/1000/9995f03c-5fb1-4ac1-8ff5-ba9a48ab377b)
EXECUTE [depth=0] sofia/internal/1000@192.168.56.21 hash(insert/192.168.56.21-last_dial/1000/1001)
EXECUTE [depth=0] sofia/internal/1000@192.168.56.21 hash(insert/192.168.56.21-last_dial/global/9995f03c-5fb1-4ac1-8ff5-ba9a48ab377b)
[...]

PR #1861, opened 2022-11-11
fix couldn't play moh when execute uuid_hold on bleg

eighty-zou:
fix issue [#1476](https://github.com/signalwire/freeswitch/issues/1476)

### DIFF
--- a/src/switch_core_media.c
+++ b/src/switch_core_media.c
@@ -4753,7 +4753,7 @@ static void switch_core_media_set_rmode(switch_core_session_t *session, switch_m
 
 	if (switch_core_session_get_partner(session, &other_session) == SWITCH_STATUS_SUCCESS) {
 
-		if (sdp_type == SDP_TYPE_RESPONSE && (switch_channel_test_flag(other_session->channel, CF_REINVITE) || switch_channel_direction(session->channel) == SWITCH_CALL_DIRECTION_OUTBOUND)) {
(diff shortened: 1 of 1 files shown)


briankwest:
Can you outline the use case because uuid_hold is as if FreeSWITCH itself pressed the hold, FS isn't involved in playing any hold music when you call it.  It's use case is for an IVR needing to put something on hold:

Example:

PHONE -> FS1 -> FS2, FS can uuid_hold directed at FS1 and FS1 will play music to PHONE.

I just want to make sure you understand what this feature was for and it has little use in many cases, unless you're using FreeSWITCH as a phone itself.

eighty-zou:
> Can you outline the use case because uuid_hold is as if FreeSWITCH itself pressed the hold, FS isn't involved in playing any hold music when you call it. It's use case is for an IVR needing to put something on hold:

our use case:
In our callcenter we use browser which support webrtc as a phone. Our webrtc phone don't have hold function(In other words we dont't send any sip messages to freeswitch when agent click hold button on webrtc phone,but just send a request to freeswitch which will call uuid_hold on this channel) .

first : pstn->freeswitch->webrtc phone
second : click hold button on webrtc phone
third :call uuid_hold on the channel which associate webrtc phone


in freeswitch version 1.10.6 and before we can hear hold music but in version1.10.8 we can't.

gpavinteractiv:
@briankwest 
I have also posted our use case related to this regression in https://github.com/signalwire/freeswitch/issues/1476
Please advise.

cqlmq:
uuid_hold [outbound's uuid] , other channel couldn't play moh

nathanleyton:
We have also just faced this bug with uuid_hold, similar use case to the other cases mentioned above. We need to perform the hold by api not sip. We are looking at a workaround using uuid displace but it would be good to see this fix in a future release. Thanks :-)

ruhnet:
@eighty-zou @gpavinteractiv @nathanleyton why can't you use the `soft_hold` application instead of `uuid_hold`?

PR #1867, opened 2022-11-20
[core] fix segfault when switch_channel_get_private gets a NULL value

greenbea:
When `switch_channel_get_private` gets a NULL value, it will segfault at https://github.com/signalwire/freeswitch/blob/master/src/include/switch_hashtable.h#L230

### DIFF
--- a/src/switch_channel.c
+++ b/src/switch_channel.c
@@ -1081,6 +1081,9 @@ SWITCH_DECLARE(switch_status_t) switch_channel_set_private(switch_channel_t *cha
 SWITCH_DECLARE(void *) switch_channel_get_private(switch_channel_t *channel, const char *key)
 {
 	void *val;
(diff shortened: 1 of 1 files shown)


anthmFS (approved):
This is ok or so is asserting the key is not NULL since it wasn't designed to accept NULL but not crashing is probably better. Consider a warning.

mjerris:
what was calling this with a null key?  this shouldn’t happen and we should fix that upstream as well probably as it doesn’t make sense to look up a bill hash key.  The only concern is that doing this instead of assert will make it harder for static analyzers to find real problems upstream

greenbea:
I think you're right. I had a NULL value being passed when working on pull request https://github.com/signalwire/freeswitch/pull/1170. At line 3233 `queue->announce` isn't always set

Issue #1862, opened 2022-11-16
uuid_media_reneg API fails to add audio component to existing video only call leg


figaro2015:
**Describe the bug**
The user is not able to add an audio component to the existing video-only call leg using the **uuid_media_reneg** API while passing both video and audio codecs as parameters to the API.

**To Reproduce**
1. Create video-only call leg using **originate** API while having _outbound_codec_prefs_ variable set to video only codec (VP8). Example: originate sofia/external/1004@192.168.86.66 3000 XML public. For the far end, it is recommended to use a video-capable SIP softphone, like Linphone. For the dial plan configuration, this call can be placed into a conference.
2. Answer the phone call on the client side.
3. Try to add an audio component to the previously created video-only call using the uuid_media_reneg API by specifying both audio and video codecs as parameters to the API (VP8 and OPUS). Example: uuid_media_reneg <channel_uuid> =VP8,OPUS
4. Observe that the audio media component doesn't get added to the SDP in the INVITE the API generates and the call remains to be video-only.

**Expected behavior**
Newly generated SDP in the INVITE should contain both audio and video media components that are using OPUS and VP8 codecs correspondently. Example:

v=0
o=FreeSWITCH 1668598456 1668598460 IN IP4 192.168.86.35
s=FreeSWITCH
c=IN IP4 192.168.86.35
t=0 0
m=audio 22410 RTP/AVP 103 101
a=rtpmap:103 opus/48000/2
a=fmtp:103 useinbandfec=1; maxaveragebitrate=30000; maxplaybackrate=48000; ptime=20; minptime=10; maxptime=40
a=rtpmap:101 telephone-event/48000
a=fmtp:101 0-15
a=ptime:20
a=sendrecv
m=video 26018 RTP/AVP 102
b=AS:3072
a=rtpmap:102 VP8/90000
a=sendrecv
a=rtcp-fb:102 ccm fir
a=rtcp-fb:102 ccm tmmbr
a=rtcp-fb:102 nack
a=rtcp-fb:102 nack pli

**Package version or git hash**
 FreeSWITCH Version 1.10.9-dev+git~20221107T233635Z~bb682fc5b4~64bit (git bb682fc 2022-11-07 23:36:35Z 64bit)

**Analysis**
[...]

PR #1868, opened 2022-11-20
Fix for the uuid_media_reneg API bug which prevents audio to be added to video-only call (bug #1862)

figaro2015:


src/switch_core_media.c
```diff
@@ -13126,6 +13126,11 @@ SWITCH_DECLARE(switch_status_t) switch_core_media_receive_message(switch_core_se
 													   "codec_string", "=%s", switch_channel_get_variable(session->channel, "ep_codec_string"));
 				}
 				
+				// Required fix to allow audio to be added to video
```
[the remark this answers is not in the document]
figaro2015: Done!

src/switch_core_media.c
```diff
@@ -13126,6 +13126,12 @@ SWITCH_DECLARE(switch_status_t) switch_core_media_receive_message(switch_core_se
 													   "codec_string", "=%s", switch_channel_get_variable(session->channel, "ep_codec_string"));
 				}
 				
+				/* Required fix to allow audio to be added to video
```
andywolk, line 13129: `Required fix` sounds a bit off. If merged there is no issue to fix anymore.

andywolk (changes requested):

src/switch_core_media.c
figaro2015: so, what exactly do you want me to change?

figaro2015: Updated misleading comment

PR #1878, opened 2022-11-27
Fix misplaced brackets inside ifs with assignment + comparison

tomek-o:
This fixes misplaced brackets in two occurrences of "if" with assignment + comparison.
I don't think this would be important functionally but with misplaced brackets all non-success statuses were changed to 1 = SWITCH_STATUS_FALSE.


andywolk:
Please take a look at mod_db in a case there is another one

tomek-o:
Yes, it looks this is another (and maybe last) one.

PR #1908, opened 2022-12-28
Update to python_fetch function in mod_python3.c

AdrianFretwell:
When python_fetch is called for configuration binding there is often no associated switch event.
This commit creates an event if it is missing, but crucially, adds the previously unused variables: section, tag_name, key_name and key_value so these variables can be accessed within the python script.


AdrianFretwell:
@andywolk Thank you for looking at this, I have updated the PR with a second commit based on your patch.

andywolk (approved):

PR #1914, opened 2023-01-04
Fix ice delay

kalmik:
* Changing ICE candidate based on BINDING_RESPONSE

### DIFF
--- a/src/include/switch_rtp.h
+++ b/src/include/switch_rtp.h
@@ -103,6 +103,8 @@ typedef struct icand_s {
 	switch_port_t rport;
 	char *generation;
 	uint8_t ready;
+	switch_time_t last_binding_response;
+	switch_sockaddr_t *addr;
 } icand_t;
 
 #define MAX_CAND 50
--- a/src/switch_core_media.c
+++ b/src/switch_core_media.c
@@ -4234,6 +4234,14 @@ static switch_status_t check_ice(switch_media_handle_t *smh, switch_media_type_t
 				engine->ice_in.cands[engine->ice_in.cand_idx[cid]][cid].priority = atol(fields[3]);
 				engine->ice_in.cands[engine->ice_in.cand_idx[cid]][cid].con_addr = switch_core_session_strdup(smh->session, con_addr);
 				engine->ice_in.cands[engine->ice_in.cand_idx[cid]][cid].con_port = (switch_port_t)atoi(fields[5]);
+				engine->ice_in.cands[engine->ice_in.cand_idx[cid]][cid].last_binding_response = 0;
+
+				switch_sockaddr_new(
+						&engine->ice_in.cands[engine->ice_in.cand_idx[cid]][cid].addr,
+						engine->ice_in.cands[engine->ice_in.cand_idx[cid]][cid].con_addr,
+						engine->ice_in.cands[engine->ice_in.cand_idx[cid]][cid].con_port,
+						smh->session->pool
+				);
 
 				j = 6;
 
--- a/src/switch_rtp.c
+++ b/src/switch_rtp.c
@@ -543,6 +543,9 @@ static void switch_rtp_change_ice_dest(switch_rtp_t *rtp_session, switch_rtp_ice
 
 	ice->ice_params->cands[ice->ice_params->chosen[ice->proto]][ice->proto].con_addr = switch_core_strdup(rtp_session->pool, host);
 	ice->ice_params->cands[ice->ice_params->chosen[ice->proto]][ice->proto].con_port = port;
+	ice->ice_params->cands[ice->ice_params->chosen[ice->proto]][ice->proto].addr->hostname = switch_core_strdup(rtp_session->pool, host);
+	ice->ice_params->cands[ice->ice_params->chosen[ice->proto]][ice->proto].addr->port = port;
+
 	ice->missed_count = 0;
 
 	if (is_rtcp) {
@@ -796,7 +799,7 @@ static int rtp_common_write(switch_rtp_t *rtp_session,
 							rtp_msg_t *send_msg, void *data, uint32_t datalen, switch_payload_t payload, uint32_t timestamp, switch_frame_flag_t *flags);
 
 
-static switch_status_t ice_out(switch_rtp_t *rtp_session, switch_rtp_ice_t *ice)
+static switch_status_t ice_out(switch_rtp_t *rtp_session, switch_rtp_ice_t *ice, icand_t *cand)
 {
 	uint8_t buf[256] = { 0 };
 	switch_stun_packet_t *packet;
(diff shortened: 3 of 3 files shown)


italorossi:
We've seen some delays to establish media on verto/webrtc calls, especially when the device using the browser (on our tests chrome) had a VPN interface that does not reach the internet or has some weird network path that could successfully send a binding request to FreeSWITCH but never receives our binding request response due to the network topology or the router discarding the packet (not NAT'd for example). So, I've been reading the code and understanding how the ice handling works inside our RTP stack and came out with a change (big?) that works perfectly for these cases reducing the media delay from 20, sometimes 30, seconds to start to less than 1 sec. The changes remove (and change) a lot of code that was handling binding requests and sticking with a wrong(bad) candidate causing the delays (if you look at the handle_ice function, all the wrong_addrs, and elapsed logic).

Basically, we changed the negotiation behavior to be more dynamic, following the RFC as soon as we receive a binding request from the remote peer we reply with a binding request response (that part was there already) and now we also send a bind request back to this candidate(before the change we were taking too long to do this), if this candidate replies with a binding request response we get the response we mark this cand as a good one storing the timestamp we received the response in the icand_s struct, the next step is getting the current candidate and if it doesn't have a request-response received we swap this current candidate with the candidate we just received the request-response (receiving a request-response is a signal that we can reach the remote peer and successfully send RTP/RTCP to it). 

We also created new candidates as soon as they are discovered (candidates not previously sent in the SDP), when we receive a binding request from a candidate that was not on the candidates list we insert it there and send a bind request, following the behavior described above. (before the change, a new candidate caused the current ice destination to be changed but not created in the candidates list, possibly overwriting an already existing candidate).

This new behavior worked perfectly for us and looks like it's a short path to implement trickle ice (maybe?).

matthew-swell:
Has this patch been tested pretty well? I have production call center that suffers from this very issue and would be willing to do additional testing of this patch assuming the author believes it to be production worthy.

andywolk:
@italorossi @kalmik We were notified that this PR may make a crash.
We also know that reverting this PR prevents the crash from happening https://pastebin.freeswitch.org/view/d5b45375
Any thoughts?

italorossi:
Which thread crashed in that Pastebin file? There's only a thread apply all bt there. We're running this patch in production for months with hundreds of calls every day and no crashes so far.

andywolk:
https://pastebin.freeswitch.org/view/d5b45375#L2777
```
Thread 1 (Thread 0x7f04d1cb2700 (LWP 498798)):
#0  find_free_frame (fb=0x5fc2ff39ebb42, orig=0x7f0520e19720) at src/switch_utils.c:144
#1  0x00007f053b939b28 in switch_frame_buffer_dup (fb=<optimized out>, orig=<optimized out>, clone=clone@entry=0x7f04d1caf7a0) at src/switch_utils.c:261
#2  0x00007f053b8e5385 in perform_write (session=session@entry=0x7f0520e18d18, frame=frame@entry=0x7f0520e19720, flags=flags@entry=0, stream_id=stream_id@entry=0) at src/switch_core_media.c:7093
#3  0x00007f053b915460 in switch_core_session_write_frame (session=0x7f0520e18d18, frame=frame@entry=0x7f04d1cafb30, flags=flags@entry=0, stream_id=stream_id@entry=0) at src/switch_core_media.c:16234
#4  0x00007f0538364f7b in conference_loop_output (member=member@entry=0x7f04d1cb1020) at conference_loop.c:1556
#5  0x00007f0538352990 in conference_function (session=<optimized out>, data=<optimized out>) at mod_conference.c:2480
#6  0x00007f053b8d013f in switch_core_session_exec (session=session@entry=0x7f0520e18d18, application_interface=application_interface@entry=0x560d9a5a73c8, arg=arg@entry=0x7f0254182820 "419520867@cdquality") at src/switch_core_session.c:2959
#7  0x00007f053b8d04fb in switch_core_session_execute_application_get_flags (session=session@entry=0x7f0520e18d18, app=0x7f0254182810 "conference", arg=0x7f0254182820 "419520867@cdquality", flags=flags@entry=0x0) at src/switch_core_session.c:2819
#8  0x00007f053b8d1d70 in switch_core_standard_on_execute (session=session@entry=0x7f0520e18d18) at src/switch_core_state_machine.c:350
#9  0x00007f053b8d4dc8 in switch_core_session_run (session=0x7f0520e18d18) at src/switch_core_state_machine.c:647
#10 0x00007f053b8cd54c in switch_core_session_thread (thread=<optimized out>, obj=0x7f0520e18d18) at src/switch_core_session.c:1726
#11 0x00007f053b8c89e4 in switch_core_session_thread_pool_worker (thread=0x7f0520e29fc0, obj=<optimized out>) at src/switch_core_session.c:1790
#12 0x00007f053bc05399 in dummy_worker (opaque=0x7f0520e29fc0) at threadproc/unix/thread.c:151
#13 0x00007f053b812609 in start_thread (arg=<optimized out>) at pthread_create.c:477
#14 0x00007f053b737133 in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
```

italorossi:
Looking..

italorossi:
I've read a few times the patch again and it's not clear where it can break. 

Do you have the steps to reproduce? Is there anything particular about the call path other than running a conference? (we don't run conferences or video calls). 

One thing that we may improve in the patch is whenever we receive an ice candidate that's not currently on the candidate list we create it and add it to the end but with a fixed transport string (udp).

ffdixon:
Hi @italorossi,

We've been using FreeSWITCH within BigBlueButton for many years.  We run 100's of BigBlueButton servers each day.  In BigBlueButton, we're using FreeSWITCH for
  - accepting incoming WebRTC audio calls (via sip.js)
  - accepting incoming calls through SIP Trunking
  
> Do you have the steps to reproduce? Is there anything particular about the call path other than running a conference? (we don't run conferences or video calls).

All audio occurs in `mod_conference`.

You can try it out here https://demo.bigbluebutton.org/.  (You'll hear the familiar "you are are currently the only one in this conference" audio message.

On those servers, we had this patch applied.  We really liked the functionality of this patch -- it significantly reduced the time for connecting for some of our users to connect.

Our servers will often have over 250 channels with FreeSWITCH.  We started experiencing crashes after deploying this version.  After experiencing a number of crashes, we removed the patch, and the crashes stopped.

We don't have a way to reproduce this issue; other than currently running it in production until a crash occurs (which is less than ideal).

Again, we *really* liked this patch.  Is our use of `mod_conference` possibly related to the issue we are encountering?

italorossi:
It's probably related to your environment, mod_conference, or some specific scenario with your users that triggers the crash.
Our WebRTC users are using mod_verto instead of sipjs but not sure if this changes anything.

If you could isolate a call flow that triggers the bug I can probably fix it.

jakubkarolczyk:
@italorossi @kalmik do you mind to put this feature under chan var and do not remove the existing code?

wesmu:
Hi, we are very very interested on this patch. It's the solution we were looking for. We face with some customers 5-6 seconds of delay with Chrome and 40-60 seconds with Firefox because symmetric NATs in the customer side.
Is there any update about this patch and the crashes? If I understand well, the problem seems to be focused only on conferences? We only use webrtc voice calls using sip.js
Thanks

italorossi:
@wesmu No updates yet, we couldn't reproduce this issue, and this makes me believe that the bug is only triggered when using mod_conference.

@jakubkarolczyk we'll do it as soon as we find time.

jakubkarolczyk:
closing in favour of https://github.com/signalwire/freeswitch/pull/2253

PR #1925, opened 2023-01-17
[mod_curl] SWITCH_STANDARD_API need to return true

greenbea:


briankwest:
Why does it need to return true?

greenbea:
> Why does it need to return true?

If it returns false when called from esl it will output `"Command not found!`

anthmFS:
That's right. The return of false from an API command is only to indicate the command is not properly configured, not to indicate the result of the requested function.

greenbea:
was closed by mistake

PR #1954, opened 2023-02-06
[Core] Fix race condition of session_table hash in switch_core_session_request_uuid()

andywolk:


andywolk:
PR is incomplete. Need some more work.

jakubkarolczyk (approved):

PR #1970, opened 2023-02-22
urlencode body string of esl event as same as headers

wuxz:
PLAY_AND_DETECT_SPEECH event will store asr result in the header, but as DETECT_SPEECH event, the asr result is in the body . Current code will urlencode the header but leave out the body, this causes DETECT_SPEECH can not be processed correctly in some esl clients. This patch will fix this problem.

### DIFF
--- a/src/switch_event.c
+++ b/src/switch_event.c
@@ -1141,7 +1141,7 @@ static switch_status_t switch_event_base_add_header(switch_event_t *event, switc
 		len = 0;
 		for(j = 0; j < header->idx; j++) {
 			len += 2;
-			if (!header->array[j]) { 
+			if (!header->array[j]) {
 				continue;
 			}
 			len += strlen(header->array[j]);
@@ -1165,7 +1165,7 @@ static switch_status_t switch_event_base_add_header(switch_event_t *event, switc
 					memcpy(hv, "|:", 2);
 					hv += 2;
 				}
-				if (!header->array[j]) { 
+				if (!header->array[j]) {
 					continue;
 				}
 				memcpy(hv, header->array[j], strlen(header->array[j]));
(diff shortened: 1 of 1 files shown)


seven1240:
the buffer is copyed to body as is, can you show an example why you need that?

wuxz:
> the buffer is copyed to body as is, can you show an example why you need that?

In normal asr function, when freeswitch get the asr result , it will generate a DETECT_SPEECH event, put the result in the body. If the result contains Chinese char, each char will be sent to esl client using raw multi bytes. The java esl client lib I use is from https://github.com/esl-client/esl-client, it process esl event like this:

    private String readLine( ChannelBuffer buffer, int maxLineLength ) throws TooLongFrameException 
    {
        StringBuilder sb = new StringBuilder(64);
        while ( buffer.readable() ) 
        {
            // this read should always succeed
            byte nextByte = buffer.readByte();
            if (nextByte == LF) 
            {
                return sb.toString();
            }
            else 
            {
                // Abort decoding if the decoded line is too large.
                if ( sb.length() >= maxLineLength ) 
                {
                    throw new TooLongFrameException(
                            "ESL message line is longer than " + maxLineLength + " bytes.");
                }
                sb.append( (char) nextByte );
            }
        }
        
        return sb.toString();
    }

It append each byte as a char to a StringBuilder, when it put multi bytes of a single Chinese char, the actual content in the StringBuilder will be chaos and can not be restored to original Chinese char. I find this lib can process PLAY_AND_DETECT_SPEECH function correctly, because the this function will use CHANNEL_EXECUTE_COMPLETE event and save asr result in it's header, not the body. 
Freeswitch will urlencode the header of every event but not the body, so I create this patch to do the same work just as that in process logic to header, that is, urlencode the body. With this patch, the esl_client lib will see correct Chinese char in asr result.

seven1240:
you should fix that esl client not FS.

andywolk:
Closing as this is not a bug

PR #1976, opened 2023-02-27
[core] switch_ivr_originate set originate endpoint used

greenbea:
This commit introduces a new channel variable that sets the channel endpoint used by the originate. This is particulary useful when using `execute_on_originate` that will execute multiple times when using fake endpoints as `user/` or `group/`. With this variable, a user can determine which endpoint is being used by the originate, and whether they want to process it.


greenbea:
This was merged https://github.com/signalwire/freeswitch/pull/1976

Issue #1965, opened 2023-02-16
One-way audio due to OPUS payload being misinterpreted as DTMF


olofssonanton:
### Describe the bug
When bridging a call (sofia to sofia) to a terminal _not_ using `183 Session Progress`, the B-side's **telephone-event** is erroneously setup to be received using the payload type intended for **OPUS** (as stated in both local and remote SDP), resulting in the incoming audio stream from B being misinterpreted as DTMF by the FreeSWITCH.

### To Reproduce
This should be reproducible in a vanilla configuration, using similar SDP/payload types as shown in the attached logs, in something like sipp.
#### To succeed:
1. Bridge call from A to B.
2. B responds with `183 Session Progress` before `200 OK` (using the same SDP and payload types).
3. DTMF send/receive payload types can be confirmed to be correct in the log `Set 2833 dtmf send payload to 102 recv payload to 102`.
4. No audio issues.

#### To fail:
1. Bridge call from A to B.
2. B responds with `200 OK`.
3. DTMF send/receive payload types can be confirmed to be incorrect in the log `Set 2833 dtmf send payload to 102 recv payload to 101`.
4. Audio from B does not reach A. FreeSWITCH continually logs `TS changed from last packet, resetting....` and `RTP RECV DTMF`.

### Expected behavior
In this case, B's telephone-event should be both sent and received using payload type 102, regardless of whether or not the B-side responds with `183 Session Progress` before `200 OK`.

### Package version or git hash
 - Tested on versions ranging from 1.10.7 to 1.10.9

### Trace logs
Find the log `2023-02-16 13:39:29.079885 91.33% [DEBUG] switch_core_media.c:5854 sofia/SomeGateway/XYZ987 Set 2833 dtmf send payload to 102 recv payload to 101` in the logs below, and then work your way back to the SDP sent to and from the B-part (called XYZ987) and you will see how `Set 2833 dtmf [...] recv payload to 101` does not make sense.
```
2023-02-16 13:39:21.539881 92.10% [NOTICE] switch_channel.c:1123 New Channel sofia/SomeGateway/ABC123@X.X.X.X [p0OJ-35-zdiDyPoTV4Usdw..]
2023-02-16 13:39:21.539881 92.10% [DEBUG] switch_core_state_machine.c:581 (sofia/SomeGateway/ABC123@X.X.X.X) Running State Change CS_NEW (Cur 1 Tot 12)
[...]

PR #1979, opened 2023-02-27
Fix signalwire/freeswitch#1965 where received audio payload is misinterpreted as DTMF

olofssonanton:
This fixes an issue with one-sided audio from A to B, where the audio from B is misinterpreted by FreeSWITCH as DTMF (signalwire/freeswitch#1965).

As mentioned in the issue, the problem stems from a misbehaviour where FreeSWITCH sets up telephone-event to be received with a payload type which is already in use for the OPUS audio stream.

This commit solves the issue by setting `smh->mparams->recv_te` in the same manner as `smh->mparams->te`, respecting the payload space.


909877998 (approved):

shaunjstokes:
There is also an issue with DTMF RFC2833 being misinterpreted as audio payload, essentially the reverse of what this patch fixes. It occurs on transfer or 3-way conference of an Opus call.

PR #1992, opened 2023-03-08
added 180 and 183 to mod_sofia.c for mod_dptools: respond  

itsmelewis:
We encountered an issues that the ability to trigger ad-hoc 180 & 183 Ringing back to a SIP Devices was not possible. Once the initial 180 `ring_ready` had been sent from FS to the SIP Device any subsequent attempts to trigger another ad-hoc 180 using `ring_ready` although received by FS did not trigger the relevant SIP messaging. The next 180 SIP Messages is triggered by a timer.

In our build of FS & ESL the use case for this is when a call-forward no answer kicks in, another 180 Ringing is needed to be sent back to the SIP Device with updated headers `sip_ph_` to inform them of the new ringing location.

This has been tested and working as needed in our build. THis is purely around the trigger of an ad-hoc SIP Message for 180/183 and is to be used in conjunction with `ring_ready` & `pre_answer`

** I did noticed that in **mod_dptools: respond** there was a notice advising this functionality had been removed (See link below) however the details where limited, for our use case and I am sure other peoples this is a very useful function. If there is a decision as to why this was removed please let me know the details so I can put a PR in to update the docs a with some information and guidance for other.

https://developer.signalwire.com/freeswitch/FreeSWITCH-Explained/Modules/mod-dptools/6586984/#warnings

### DIFF
--- a/src/mod/endpoints/mod_sofia/mod_sofia.c
+++ b/src/mod/endpoints/mod_sofia/mod_sofia.c
@@ -725,7 +725,7 @@ static switch_status_t sofia_answer_channel(switch_core_session_t *session)
 
 		if (switch_channel_test_flag(tech_pvt->channel, CF_PROXY_MEDIA)) {
 			sofia_media_activate_rtp(tech_pvt);
-			switch_core_media_patch_sdp(tech_pvt->session);			
+			switch_core_media_patch_sdp(tech_pvt->session);
 			switch_core_media_proxy_remote_addr(tech_pvt->session, NULL);
 		}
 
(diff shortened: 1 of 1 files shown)


briankwest:
Please resubmit without the white space changes, to minimize changes.

briankwest:
We'll also need to get a test case, example, If you respond 183, then respond 180, what happens, you'll probably need to build a SIPP scenario to accomplish this test case to ensure we don't do the wrong thing if we go backwards, like remove or add media incorrectly.

itsmelewis:
I'll get that sorted and resubmit. Noted on the whitespace.

itsmelewis:
Reverted and Re added the change without the whitespace changes. Will do some extensive testing. We have scenarios where we can test this. Will update as soon as I can

itsmelewis:
I have updated the branch, not sure why the Unit tests failed. On reviewing them it looked to be something unrelated to this work.

As for the testing, we have done some testing but are unable to fully test as we have an issue with Early Media on ESL that we are working on.

PR #2012, opened 2023-03-31
Update mod_sofia.c handling long bye reason

bit4bit:
hello i recently starts a integration with msteams everything it's working,
 but when the reason is big the softswitch not sent  `bye` after enabling loglevel for nua i found `nua_stack.c:301 nua_stack_event() nua(0x7f430c014660): event r_bye 900 Internal error at nua_client.c:553` and after some research i found the `reason` was big
example:

~~~
sip::reason_str: "Q.850;cause=16;text="b5bc901e-8914-4b24-b457-488bb5c83738;Participant was removed from the conversation by another participant." 
~~~

so the `span_quoted` does not works correctly because not found the end "

after increasing `reason` to `256` now `bye` works ok.

another solution for the problems described in #1813

tested in 
~~~
FreeSWITCH (Version 1.10.10-dev git f5f7f76 2023-03-29 16:19:47Z 64bit)
~~~


bit4bit:
a partial solution is

~~~xml
<action application="export" data="disable_q850_reason=true"/>
~~~

PR #2052, opened 2023-04-17
[mod_dptools] coverity CID 1468646 (Unsigned compared against 0)

dragos-oancea:


src/mod/applications/mod_dptools/mod_dptools.c
```diff
@@ -4584,7 +4584,7 @@ SWITCH_STANDARD_APP(wait_for_silence_function)
 			timeout_ms = switch_atoui(argv[3]);
 		}
 
-		if (thresh > 0 && silence_hits > 0 && listen_hits >= 0) {
+		if (thresh > 0 && silence_hits > 0 && listen_hits > 0) {
```
andywolk, line 4587: Right solution would be to remove the check as `listen_hits = 0` must be allowed:
```
if (thresh > 0 && silence_hits > 0) {
```

andywolk (changes requested):

andywolk (approved):

PR #2070, opened 2023-05-01
Revert "[Core, mod_curl, mod_httapi, mod_http_cache] Compatible with libcurl>=7.87.0"

andywolk:
This reverts commit 039823848906b11fd6183ba81f037b7211a29497.


andywolk:
We are facing a few issues here. The form of our curl interface in FreeSWITCH was not ideal and it is still questionable. So we would like to revert with these notes for the future:
* We should do our best to avoid making breaking changes to the FreeSWITCH API.
* Replacing one curl related structure to another in FreeSWITCH API is something we would like to avoid as this creates issues (for example with out of tree modules).
* We would need a solution where `curl_mime **mimep` is abstracted and does not require libcurl headers to build.
* As a small note: In the current code there is an issue in the patched `switch_curl_process_form_post_params`. Was returning `NULL` in the past if `curl_formadd` was never called. Will not return `NULL` in the current patched version.

We are open to discuss any future solutions for this topic!

andywolk:
@azzbcc ^^

mjerris (approved):

Issue #1961, opened 2023-02-10
Problem with rfc2833-pt and DTMF on bridged calls ever since.


mahald:
**Describe the bug**

The DTMF transmission for bridged calls has been problematic for some time now, particularly when the setting "bypass_media_after_bridge=true" is used. This issue arises when the destination PBX utilizes a different PT value for RFC2833 than what is configured in the sofia profile.

To address this challenge, we have implemented a workaround that involves parsing the PT from the switch_r_sdp on the A leg and using a profile with the same value set as rfc2833-pt. For example, if the PT value is 101 on the A-Leg, we use the profile "sofia/outbound101", and if it is 103, we use "sofia/outbound103". This workaround has effectively resolved our DTMF transmission issues.

However, this solution requires extensive configuration and coding on our end, which some customers may not find favorable. Additionally, each profile must utilize a different source UDP Port, further complicating the matter.

It would be beneficial if there was a channel variable for "rfc2833-pt" that takes precedence over the profile setting when it is set.

briankwest:
Channel variables can't have dashes, it must be an _, and can you do a PR?

mahald:
added pullRequest #2083

mahald:
added pull request #2573 @briankwest can you please merge this? This Code is Live for > 1 year.

mahald:
@briankwest can you please merge this?

PR #2083, opened 2023-05-12
#1961 resolve [Core] add channel var rtp_2833_payload to overwrite rfc2833-pt in sip profile

mahald:
Problem with rfc2833-pt and DTMF on bridged calls ever since. New channel variable rtp_2833_payload to set default dmtf payload on outbound calls


src/switch_core_media.c
```diff
@@ -10408,6 +10408,12 @@ SWITCH_DECLARE(void) switch_core_media_gen_local_sdp(switch_core_session_t *sess
 			for (i = 0; i < smh->num_rates; i++) {
 				if (smh->rates[i] == 8000 || smh->num_rates == 1) {
 					smh->dtmf_ianacodes[i] = smh->mparams->te;
+
+					if (!zstr(switch_channel_get_variable(session->channel, "rtp_2833_payload"))) {
```
andywolk, line 10412: Can we get this once?

mahald:
ok done

mahald:
Are you going to merge this pull request? As far as i understand the failing notify has nothin to do with the code.

mahald:
I don't see what test acutally fail but it seems not be related to this work.

PR #2090, opened 2023-05-21
[mod_shout] init buffer to fix encode noise

elkpi:


andywolk:
Where is the noise from? Is it because `len` given by `lame_encode_buffer` is not checked for errors?
``` 
 * return code     number of bytes output in mp3buf. Can be 0
 *                 -1:  mp3buf was too small
 *                 -2:  malloc() problem
 *                 -3:  lame_init_params() not called
 *                 -4:  psycho acoustic problems
 ```

**Comment by elkpi** [comment]
> Where is the noise from? Is it because `len` given by `lame_encode_buffer` is not checked for errors?
> 
> ```
>  * return code     number of bytes output in mp3buf. Can be 0
>  *                 -1:  mp3buf was too small
>  *                 -2:  malloc() problem
>  *                 -3:  lame_init_params() not called
>  *                 -4:  psycho acoustic problems
> ```

https://android.googlesource.com/platform/external/lame/+/refs/heads/master/libmp3lame/lame.c

in lame_encode_buffer_sample_t 

`mp3out = copy_buffer(gfc, mp3buf, mp3buf_size, 0);`

If the output buffer is not initialized, dirty data may be imported

**Comment by andywolk** [comment]
Since `copy_buffer` does `copy data out of the internal MP3 bit buffer into a user supplied unsigned char buffer` I see no sense in initializing the output buffer prior to that. If that means garbage data is wrongly taken from the output buffer on the application level then we need to answer the question `why?`. I can see that it is possible just because `lame_encode_buffer` may return an error as a negative value that would be interpreted as non-zero and success. So my question remains: 
```
Is it because len given by lame_encode_buffer is not checked for errors?
```

elkpi:
These noises only appear on dual channels. 2 chs, return > 0, One of the channels may have used buffer data.
Or need to use` lame_ encode_ buffer_ interleaved `
Our online practice is to initialize the buffer

andywolk:
Can you explain a little bit why initializing only first byte in the buffer (not entire buffer) does help?

elkpi:
> Can you explain a little bit why initializing only first byte in the buffer (not entire buffer) does help?

My mistake, initializing entire buffer

andywolk:
We are talking about mp3 encoded buffer data, not RAW. How can one channel have used buffer data? I need some more technical explanations here to understand this PR.

andywolk:
@elkpi Do you have a way of replicating the issue?

PR #2093, opened 2023-05-25
Update phrase_en.xml

yichaocp:
Fix a invalid word  Beacuase -> Because.
This issue causes garbled characters when the phrase is translated as Chinese.

BTW:
The correct meaning of this phrase is "Because, the voices where pretty good"  or  "Because, the voices were pretty good" ?


ghzserg (approved):
Syntax error fixed

PR #2100, opened 2023-06-08
[Core] Fix switch_console.c for Galera Mariadb cluster

ghzserg:
Fix in switch_console.c file.
Adds column names when inserting data.
Calls to use the complete table in Galera Multi-Master Cluster


as4303 (approved):
This change avoids getting the error.
Column count doesn't match value count at row 1
In request
[insert into complete values (0,'local_stream', 'start', '', '', '', '', '', '', '', '', 'PFS-1')]
If Galera cluster is used in Multi-Master configuration

ghzserg:
@andywolk, this patch is not enough for Galera Cluster to work.  The patch does not violate compatibility.  Could you please review the code.

andywolk (approved):

mjerris (approved):

Issue #2063, opened 2023-04-24
Freeswitch + keepalived + MySQL(ODBC) - Deadlock found when trying to get lock; try restarting transaction


nikkadim:
1) Two Freeswitch (Version 1.10.9) servers Active and Backup with keepalived (as described [here](https://freeswitch.org/confluence/display/FREESWITCH/High+Availability) and [here](https://freeswitch.org/confluence/display/FREESWITCH/HA+keepalived)) to control floating/virtual IPs, that part works fine. 
2) Switched core_db and all profiles from SQLite MySQL (via ODBC)  to sync two FS and start getting the error: "Deadlock found when trying to get lock; try restarting transaction". 

```
2023-04-24 01:33:12.298041 98.90% [ERR] switch_odbc.c:529 ERR: [update sip_registrations set ping_expires = 1682339621 where hostname='fs' and profile_name='external' and ping_expires <= 1682339592 ]

[STATE: HY000 CODE 1213 ERROR: [MySQL][ODBC 5.3(a) Driver][mysqld-5.7.35-log]Deadlock found when trying to get lock; try restarting transaction

]

2023-04-24 01:33:12.298041 98.90% [ERR] switch_core_sqldb.c:728 [db="fs_core";type="odbc"user="broot";pass="**********"] ODBC SQL ERR [STATE: HY000 CODE 1213 ERROR: [MySQL][ODBC 5.3(a) Driver][mysqld-5.7.35-log]Deadlock found when trying to get lock; try restarting transaction

]

update sip_registrations set ping_expires = 1682339621 where hostname='fs' and profile_name='external' and ping_expires <= 1682339592
```

**Describe the bug**
Switched core_db and all profiles from SQLite MySQL (via ODBC) and start getting the error: "Deadlock found when trying to get lock; try restarting transaction". 
**To Reproduce**
1) two Freeswitch (Version 1.10.9) servers Active and Backup with keepalived to control floating/virtual IPs, that part works fine. 
2) Switched core_db and all profiles from SQLite MySQL (via ODBC) and start getting the error: "Deadlock found when trying to get lock; try restarting transaction". 

FreeSWITCH (Version 1.10.9 -release-21-a615e85afc 64bit)
Debian GNU/Linux 11 (bullseye)
MySQL 5.7.35
ODBC 5.3(a) Driver

nikkadim:
> try using mod_mariadb instead of odbc

@andywolk for the core_db?

nikkadim:
[...]

PR #2101, opened 2023-06-08
[mod_mariadb] Fix "DeadLock. The retries are over." message.

ghzserg:
Fix DeadLock. https://github.com/signalwire/freeswitch/issues/2063
If the message "Deadlock found when trying to get lock; try restarting transaction" is specified, then 100 attempts are made to retry the SQL request.
If all attempts fail, the message "DeadLock. The retries are over." is displayed in the logs.

### DIFF
--- a/src/mod/databases/mod_mariadb/mod_mariadb.c
+++ b/src/mod/databases/mod_mariadb/mod_mariadb.c
@@ -627,13 +627,27 @@ switch_status_t mariadb_send_query(mariadb_handle_t *handle, const char* sql)
 {
 	char *err_str;
 	int ret;
+	unsigned retries = 60; /* 60 tries, will take 30 to 60 seconds at worst */
 
 	switch_safe_free(handle->sql);
 	handle->sql = strdup(sql);
+    again:
 	handle->stored_results = 0;
 	ret = mysql_real_query(&handle->con, sql, (unsigned long)strlen(sql));	
 	if (ret) {
 		err_str = mariadb_handle_get_error(handle);
+		if (strstr(err_str, "Deadlock found when trying to get lock; try restarting transaction")) {
+			if (--retries > 0) {
+				switch_safe_free(err_str);
+				/* We are waiting for 500 ms and random time is not more than 500 ms.
+				  This is necessary so that the delay on the primary and secondary servers does not coincide and deadlock does not occur again. */
+				switch_yield(500 + (rand() & 511));
+				goto again;
+			}
+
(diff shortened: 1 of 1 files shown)


src/mod/databases/mod_mariadb/mod_mariadb.c
```diff
@@ -627,13 +627,22 @@ switch_status_t mariadb_send_query(mariadb_handle_t *handle, const char* sql)
...
 	handle->sql = strdup(sql);
+    again:
 	handle->stored_results = 0;
 	ret = mysql_real_query(&handle->con, sql, (unsigned long)strlen(sql));	
 	if (ret) {
 		err_str = mariadb_handle_get_error(handle);
+		if (strstr(err_str, "Deadlock found when trying to get lock; try restarting transaction")) {
+			if (--retries > 0) {
+			    switch_safe_free(err_str);
+			    goto again;
```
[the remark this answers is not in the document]
ghzserg: Changed the code.
Added a database wait for a time from 500 to 1000 ms.
Random locking is needed in case 2 instances of freeswitch work in parallel and simultaneously try to work with the same table

src/mod/databases/mod_mariadb/mod_mariadb.c
```diff
@@ -627,13 +627,25 @@ switch_status_t mariadb_send_query(mariadb_handle_t *handle, const char* sql)
...
 	handle->stored_results = 0;
 	ret = mysql_real_query(&handle->con, sql, (unsigned long)strlen(sql));	
 	if (ret) {
 		err_str = mariadb_handle_get_error(handle);
+		if (strstr(err_str, "Deadlock found when trying to get lock; try restarting transaction")) {
+			if (--retries > 0) {
+				switch_safe_free(err_str);
+				// We are waiting for 500 ms and random time is not more than 500 ms. 
+				// This is necessary so that the delay on the primary and secondary servers does not coincide and deadlock does not occur again.
+				switch_yield(500+rand()%500);
```
andywolk, line 644: Just a small suggestion: if it is possible to avoid using modulo operation then we should. how about using `and` with a power of 2 (where 512 is 2^9) instead?
`switch_yield(500 + (rand() & 511))`

Also we should use `/* */` comment style.
Also we should make a new line after an `if` block.

andywolk (changes requested):

src/mod/databases/mod_mariadb/mod_mariadb.c
ghzserg: Fixing

andywolk:
@seven1240 Are you happy with the fix?

andywolk (approved):

PR #2130, opened 2023-06-28
Correcly handle route uri parsing from contact, fixes #1897

### LINKED ISSUE #1897: OPTIONS ping not working when fs_path is set with "<>" on FS 1.10.8
**Describe the bug**
When setting fs_path variable through a proxy that is setting the Path header with "<>", OPTIONS pings are not being sent out towards the endpoint. URL in registrations table will look like this for example: 
"sofia/internal/sip:10@8.8.8.8:12345;transport=TCP;alias=8.8.8.8~12345~2;fs_path=%3Csip%3A192.168.0.1%3Blr%3Breceived%3Dsip%3A8.8.8.8%3A12345%3Btransport%3Dtcp%3E"

While the encoded "<>" are set, i get following error in the freeswitch console (sofia loglevel 7 enabled):
"nta.c:8149 outgoing_create() nta outgoing create: invalid URI
nua_stack.c:301 nua_stack_event() nua(0x7f412800a0c0): event r_options 900 Internal error at nua_client.c:713"

When i manually delete the encoded "<>" in the freeswitch database, so that the URL in registrations table looks like this:
"sofia/internal/sip:10@8.8.8.8:12345;transport=TCP;alias=8.8.8.8~12345~2;fs_path=sip%3A192.168.0.1%3Blr%3Breceived%3Dsip%3A8.8.8.8%3A12345%3Btransport%3Dtcp" OPTIONS pings towards that endpoint are being sent, but only until the next registration of the endpoint is received

**To Reproduce**
1. set up freeswitch as registrar
2. set up for example a kamailio proxy that sets "Path" header
3. enable "all-reg-options-ping" in freeswitch profile for continous OPTIONS pings

**Expected behavior**
OPTIONS pings should be sent out towards the endpoints when fs_path is set with "<>"

 - Version 1.10.8
 (0365bec could be related?)

Please let me know if you need more info in order to analyze the problem

Best regards
Thomas

xadhoom:
When the Path header contains `<>`, it will get stored along the contact, but we cannot send as-is into the route_uri params, because is used as an outbound proxy val (normally in NUTAG_PROXY which expects an uri), so we are extracting from the `<>` containing route param the correct uri to use.

### DIFF
--- a/src/mod/endpoints/mod_sofia/sofia_glue.c
+++ b/src/mod/endpoints/mod_sofia/sofia_glue.c
@@ -3057,6 +3057,21 @@ sofia_destination_t *sofia_glue_get_destination(char *data)
 				*p = '\0';
 			} while ((--p > route_uri) && *p == ' ');
 		}
+		if ((strlen(route_uri) > 1) && (p = strchr(route_uri, '<'))) {
+		    for (;p && *p; p++) {
+				if (*p == '>' || *p == ';') {
+					*p = '\0';
+					break;
+				}
+			}
+			p = route_uri;
+			if (!(route_uri = strdup(p + 1))) {
+				goto mem_fail;
+			}
+			if (p) {
+				switch_safe_free(p);
+			}
+		}
 	}
 
 	if (!(to = strdup(data))) {


taju20910:
Hi @xadhoom 
Firstly thank you for the handling here.
Want to understand why removing anything after semicolon is required here. To me replacing after > should have been enough ideally.
I faced issue due to above as my route_uri looked like this initially <sip:host.docker.internal:32070;transport=tcp;lr> and finally the route_uri looked like sip:host.docker.internal:32070 , eventually my NOTIFY instead of getting sent via TCP was sent via UDP. The server didn't accept the request eventually.
Changing if (*p == '>' || *p == ';') { -> if (*p == '>') { did the trick for me

PR #2133, opened 2023-06-28
[core] fix fct build warning on debian bookworm

seven1240:


seven1240:
this is already fixed the same way in master

Issue #2172, opened 2023-07-20
Windows/VS 2022 build errors in mod_enum and two missing vcxproj files


KrisD888:
**Describe the bug**
30 multiply defined symbols errors when building Freeswitch project mod_enum and errors loading docs.2017.vcxproj and FSComm.2017.vcxproj 

**To Reproduce**
1. Compiled Master from 07/19/2023 and on several occasions in the past 6 months
2. Using Windows Server 2019 and 
3. Visual Studio Community 2022 (64-bit) - Version 17.6.5 and .NET Framework 4.8.03761

The rest of the solution compiles OK and Freeswitch runs well without mod_enum.

**Expected behavior**
It would be nice to clear up these errors even though they don't seem to affect my operation.

**Package version or git hash**
 FreeSWITCH Version 1.10.10-dev-5874~e764f292b5~64bit (e764f292b5 64bit)

Build Errors

Severity	Code	Description	Project	File	Line	Suppression State
Error	LNK2005	SocketNotificationRetrieveEvents already defined in ldns.lib(rdata.obj)	mod_enum	E:\FS_GIT\src\mod\applications\mod_enum\ldns.lib(dname.obj)	1	
Error	LNK2005	SocketNotificationRetrieveEvents already defined in ldns.lib(rdata.obj)	mod_enum	E:\FS_GIT\src\mod\applications\mod_enum\ldns.lib(rr.obj)	1	
Error	LNK2005	SocketNotificationRetrieveEvents already defined in ldns.lib(rdata.obj)	mod_enum	E:\FS_GIT\src\mod\applications\mod_enum\ldns.lib(packet.obj)	1	
Error	LNK2005	SocketNotificationRetrieveEvents already defined in ldns.lib(rdata.obj)	mod_enum	E:\FS_GIT\src\mod\applications\mod_enum\ldns.lib(resolver.obj)	1	
Error	LNK2005	SocketNotificationRetrieveEvents already defined in ldns.lib(rdata.obj)	mod_enum	E:\FS_GIT\src\mod\applications\mod_enum\ldns.lib(host2str.obj)	1	
Error	LNK2005	SocketNotificationRetrieveEvents already defined in ldns.lib(rdata.obj)	mod_enum	E:\FS_GIT\src\mod\applications\mod_enum\ldns.lib(str2host.obj)	1	
[...]

PR #2188, opened 2023-08-05
[Build-System, ldns] config.h change to not define inline under Windows to avoid clash with Windows winsock2.h inline usage. Use updated ldns tarball on Windows.

Visytel:
Based on the open Issue (https://github.com/signalwire/freeswitch/issues/2172) this is the PR to resolve this issue.

This is all around the fact that the latest Windows SDK, containing winsock2.h, makes use of inline which clashes with the declaration of inline inside the ldns config.h.  This then causes LNK2004 warnings when ldns is built and hence the LNK2005 errors with mod_enum that uses ldns.

~~Also I needed to modify the windows ldns project file to force overwrite of libs\ldns\ldns\config.h with libs\win32\ldns\ldns-lib\config.h when ldns is compiled.~~

### DIFF
--- a/libs/win32/ldns/ldns-lib/ldns-lib.2017.vcxproj
+++ b/libs/win32/ldns/ldns-lib/ldns-lib.2017.vcxproj
@@ -95,7 +95,7 @@
       <DisableSpecificWarnings>4013;4101;4996;4267;4244;%(DisableSpecificWarnings)</DisableSpecificWarnings>
     </ClCompile>
     <PreBuildEvent>
-      <Command>type "$(ProjectDir)\config.h" &gt; "$(ProjectDir)..\..\..\ldns\ldns\config.h"
+      <Command>if not exist "$(ProjectDir)..\..\..\ldns\ldns\config.h" type "$(ProjectDir)\config.h" &gt; "$(ProjectDir)..\..\..\ldns\ldns\config.h"
 if not exist "$(ProjectDir)..\..\..\ldns\ldns\util.h" type "$(ProjectDir)\util.h" &gt; "$(ProjectDir)..\..\..\ldns\ldns\util.h"
 if not exist "$(ProjectDir)..\..\..\ldns\ldns\net.h" type "$(ProjectDir)\net.h" &gt; "$(ProjectDir)..\..\..\ldns\ldns\net.h"</Command>
     </PreBuildEvent>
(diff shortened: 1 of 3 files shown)


libs/win32/ldns/ldns-lib/ldns-lib.2017.vcxproj
```diff
@@ -95,7 +95,7 @@
       <DisableSpecificWarnings>4013;4101;4996;4267;4244;%(DisableSpecificWarnings)</DisableSpecificWarnings>
     </ClCompile>
     <PreBuildEvent>
-      <Command>if not exist "$(ProjectDir)..\..\..\ldns\ldns\config.h" type "$(ProjectDir)\config.h" &gt; "$(ProjectDir)..\..\..\ldns\ldns\config.h"
+      <Command>type "$(ProjectDir)\config.h" &gt; "$(ProjectDir)..\..\..\ldns\ldns\config.h"
```
andywolk, line 98: Do we really need this?

Visytel: The short answer is YES.  However whether it is the best solution is very debatable.

Let me explain why this change is needed.

Currently the tar ball (http://files.freeswitch.org/downloads/libs/ldns-1.6.9-1-win.tar.gz) downloaded by the  download_LDNS.props includes **freeswitch\libs\ldns\ldns\config.h** (which is the file that needs to be changed (hence this PR)).  However since the tar ball cannot be changed then the file I changed was **freeswitch\libs\win32\ldns\ldns-lib\config.h**.  So somehow freeswitch\libs\win32\ldns\ldns-lib\config.h needs to replace freeswitch\libs\ldns\ldns\config.h in order for this fix to take effect.

**So the options are:**

1) Change freeswitch\libs\win32\ldns\ldns-lib\ldns-lib.2017.vcxproj to perform this task - which is what I have done already inside this PR.  This was based on what was already inside the project file.

or

2) Change the original tar ball contents with the fix inside freeswitch\libs\ldns\ldns\config.h - which you stated previously seemed like not an option.

or

3) Add a copy feature to download_LDNS.props to copy freeswitch\libs\win32\ldns\ldns-lib\config.h to overwrite freeswitch\libs\ldns\ldns\config.h during the first build of ldns.  Perhaps this could also include the other header files inside freeswitch\libs\win32\ldns\ldns-lib to be a more complete solution.  

Since you are one of the maintainers of the FreeSWITCH build system (especially the windows build) then I would really appreciate your ideas as to how to best make this work.

You are more than welcome to select an option from the above or suggest another option and I will follow your lead on this one.  The main aim is to get this live prior to the next release of FreeSWITCH.

Andrew

andywolk: We can change the tarball.

andywolk: Pushed an update.

andywolk (approved):

Visytel (commented):
Looks good to me.  You are welcome to close this PR.  Thanks for sorting this Andy.

Visytel:
Sorry should not have closed this.  My bad.  Since it has not been merged yet.

Issue #2195, opened 2023-08-08
Segmentation fault on latest master on srtp teardown and openssl 3


xadhoom:
**Describe the bug**
Freeswitch segfault when tearing down an SRTP session and is built with openssl 3

**To Reproduce**
1. Place a call using DTLS/SRTP (webRTC client)
3. Hangup the call from FS side
4. FreeSWITCH crashes

**Expected behavior**
FS does not crash :)

**Package version or git hash**
 - Git revision 41507363f3fffcdad547b168e55fbe3383a24c3d

**backtrace from core file**
See attached log.
[backtrace.log](https://github.com/signalwire/freeswitch/files/12291574/backtrace.log)

PR #2197, opened 2023-08-09
[core] Use auto DH params with openssl3

xadhoom:
This fixes https://github.com/signalwire/freeswitch/issues/2195

From OpenSSL 3 docs for [SSL_CTX_set0_tmp_dh_pkey](https://www.openssl.org/docs/man3.0/man3/SSL_CTX_set0_tmp_dh_pkey.html):

`Applications may supply their own DH parameters instead of using the built-in values. This approach is discouraged and applications should in preference use the built-in parameter support described above. Applications wishing to supply their own DH parameters should call SSL_CTX_set0_tmp_dh_pkey() or SSL_set0_tmp_dh_pkey() to supply the parameters for the SSL_CTX or SSL respectively. The parameters should be supplied in the dhpkey argument as an EVP_PKEY containing DH parameters. Ownership of the dhpkey value is passed to the SSL_CTX or SSL object as a result of this call, and so the caller should not free it if the function call is successful.`

So the code in https://github.com/signalwire/freeswitch/blob/master/src/switch_rtp.c#L3738 is partially wrong because if the call of `SSL_CTX_set0_tmp_dh_pkey` is successful the key should *not* be freed, so `EVP_PKEY_free(dh_pk)` should not be called. But this is also discouraged, and using `SSL_set_dh_auto` is preferred.

This PR just does that, which result the segfault reported in #2195 to not happen anymore, because basically FS was freeing a key that was also freed by the underlying SSL_CTX_free when tearing down dtls.

### DIFF
--- a/src/switch_rtp.c
+++ b/src/switch_rtp.c
@@ -3642,10 +3642,8 @@ SWITCH_DECLARE(switch_status_t) switch_rtp_add_dtls(switch_rtp_t *rtp_session, d
 	unsigned long ssl_ctx_error = 0;
 	const SSL_METHOD *ssl_method;
 	SSL_CTX *ssl_ctx;
+#if OPENSSL_VERSION_NUMBER < 0x30000000
 	BIO *bio;
-#if OPENSSL_VERSION_NUMBER >= 0x30000000
-	EVP_PKEY *dh_pk;
-#else
 	DH *dh;
 #endif
 	switch_status_t status = SWITCH_STATUS_SUCCESS;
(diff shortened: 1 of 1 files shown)


src/switch_rtp.c
```diff
@@ -3726,21 +3724,19 @@ SWITCH_DECLARE(switch_status_t) switch_rtp_add_dtls(switch_rtp_t *rtp_session, d
...
 		SSL_CTX_set_tmp_dh(dtls->ssl_ctx, dh);
 		DH_free(dh);
 	}
-#else 
-	if((dh_pk = PEM_read_bio_PrivateKey(bio, NULL, NULL, NULL)) != NULL) {
-		SSL_CTX_set0_tmp_dh_pkey(dtls->ssl_ctx, dh_pk);
-		EVP_PKEY_free(dh_pk);
-	}
-
 	BIO_free(bio);
```
andywolk, line 3735: Double free?

andywolk (changes requested):

src/switch_rtp.c
xadhoom: Yes, explicit here and implicit by SSL_CTX_free, as per docs ```Ownership of the dhpkey value is passed to the SSL_CTX or SSL object as a result of this call, and so the caller should not free it if the function call is successful.```

xadhoom: Ok, got it, my error on reporting the patch. will fix immediately.

xadhoom: fixed

andywolk (approved):

PR #2210, opened 2023-08-18
[sofia_presence.c] fixes various incorrect select statements impacting database performance

themsley-voiceflex:
…database performance - issue 2209

### DIFF
--- a/src/mod/endpoints/mod_sofia/sofia_presence.c
+++ b/src/mod/endpoints/mod_sofia/sofia_presence.c
@@ -742,7 +742,7 @@ static void do_normal_probe(switch_event_t *event)
 
 								 "left join sip_dialogs on "
 								 "sip_dialogs.hostname = sip_registrations.hostname and sip_dialogs.profile_name = sip_registrations.profile_name and ("
-								 "sip_dialogs.presence_id = sip_registrations.sip_user %q '@' %q sip_registrations.sub_host "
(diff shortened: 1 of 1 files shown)


junction1153:
This has a bug if you are using multiple sip profiles and are sending subscribe messages across different SIP profiles,

example: SIP profile A hosts x101 and SIP profile B hosts x501

x201 sends a subscribe to monitor presence for x501

x501 gets on a call. 501's light is lit up on 201's phone

x501's subscribe expiration time has reached the threshold and sends a new subscribe. x501 will still be on the phone, but the presence light will no longer show that 501 is on a phone call.

themsley-voiceflex:
The only change to the select here is to stop it from retrieving what appears to be all rows in the database. There was a misplaced OR in the statement that meant that it would mistakenly select all rows that matched (among other things)
or sip_registrations.presence_hosts like '%%%q%%'
which is basically every row in the table. This is plainly wrong so the patch as given is correct as far as that goes. The loogic behind what it selects otherwise may be wrong but that doesn't affect the fix that is given as the unpatched behaviour is clearly a bug.

junction1153:
It’s possible that the patch just needs more tweaking to work across multiple sip profiles. I’m not a C programmer nor am I affiliated with Signalwire but I’m merely providing my test results and I want to make sure that the dev team sees this so that they can test and tweak further as needed before accepting this pull request

PR #2234, opened 2023-09-08
updates for the Debian Dockerfile

tmancill:
Hi, I'm not 100% sure what context(s) this Dockerfile is used in, but it appeared to be the right starting place for building images based on the package in the SignalWire repo.  The goals of the PR:

- default to Debian bookworm
- drop the dependency on `lsb-release` since `DEBIAN_VERSION` is supplied as an ARG (although `lsb-release` has gotten much smaller starting in bookworm)
- use Debian's [gosu](https://tracker.debian.org/pkg/gosu) package; it is available since at least buster
- add a `FS_META_PACKAGE` build-arg to make it easy to build other meta package-based variants
- add the `-c` argument to `SHELL`;  without this, a docker build `RUN` using this image as a base would fail with an exec error

I've tested building and running several variants like so:
``` sh
docker build --build-arg="TOKEN=${SIGNALWIRE_PAT}" \
    --build-arg="FS_META_PACKAGE=freeswitch-meta-vanilla" . -t fs:deb-vanilla
```

Creating a derived image:
```Dockerfile
# dockerfile that installs mod_switchwire
FROM fs:deb-vanilla
RUN apt-get update && apt-get install -y freeswitch-mod-signalwire
```

And then running it like so:
```sh
docker run -it \
    --network host \
    --volume="/etc/freeswitch:/etc/freeswitch:rw" \
    --volume="/var/lib/freeswitch:/var/lib/freeswitch:rw" \
    --volume="/var/log/freeswitch:/var/log/freeswitch:rw" \
    --volume="/usr/share/freeswitch:/usr/share/freeswitch:ro" \
    fs:deb-vanilla
```

I'm happy to split up the PR if you'd like it composed differently.   Thank you for maintaining FreeSWITCH!

### DIFF
--- a/docker/master/Dockerfile
+++ b/docker/master/Dockerfile
@@ -66,11 +66,9 @@ VOLUME ["/tmp"]
 COPY    build/freeswitch.limits.conf /etc/security/limits.d/
 
 # Healthcheck to make sure the service is running
-SHELL       ["/bin/bash"]
+SHELL       ["/bin/bash", "-c"]
 HEALTHCHECK --interval=15s --timeout=5s \
     CMD  fs_cli -x status | grep -q ^UP || exit 1
 
 ENTRYPOINT ["/docker-entrypoint.sh"]
-
-
 CMD ["freeswitch"]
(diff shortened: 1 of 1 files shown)


Len-PGH (approved):
thank you

PR #2260, opened 2023-10-04
[mod_verto] Support for JSON message body in the HTTP POST request

ajay-sabat:
In past mod_verto supported only urlencoded data in the message body of the received http request.

This fix is to support JSON encoded data in the message body when the Content-Type header is JSON as below.
`Content-Type: application/JSON`
Now from the Lua script you can directly fetch the message body as
`body = env:getBody()`

### DIFF
--- a/src/mod/endpoints/mod_verto/mod_verto.c
+++ b/src/mod/endpoints/mod_verto/mod_verto.c
@@ -1764,8 +1764,7 @@ new_req:
 		goto done;
 	}
 
-	if (!strncmp(request->method, "POST", 4) && request->content_length && request->content_type &&
-		!strncmp(request->content_type, "application/x-www-form-urlencoded", 33)) {
+    if (!strncmp(request->method, "POST", 4) && request->content_length && request->content_type) {
 
 		char *buffer = NULL;
 		switch_ssize_t len = 0, bytes = 0;
@@ -1797,7 +1796,22 @@ new_req:
 
 		*(buffer + bytes) = '\0';
 
-		kws_parse_qs(request, buffer);
+        // Supports both urlencoded and json message body in the http request
+        if (!strncmp(request->content_type, "application/x-www-form-urlencoded", 33)) {
+			kws_parse_qs(request, buffer);
+		} else if (!strncmp(request->content_type, "application/json", 16)) {
+			cJSON *json = NULL;
+			json = cJSON_Parse(buffer);
+			if (json) {
+				switch_event_set_body(stream.param_event, buffer);
+				cJSON_Delete(json);
+			} else {
+				switch_log_printf(SWITCH_CHANNEL_LOG, SWITCH_LOG_ERROR, "Invalid JSON data in message body. content_length: %ld, body: %s \n", request->content_length, buffer);
+				free(buffer);
+				goto request_err;
+			}
+		}
+
 		free(buffer);
 	}
 


seven1240:
this patch changes the body logic, it will break scripts that read body with lua `stream:read()`. I'd like to process body only on known content types. e.g. 

```
	is_json = request->content_type && !strncmp(request->content_type, "application/json", 16);
	is_url_encoded = request->content_type && !strncmp(request->content_type, "application/x-www-form-urlencoded", 33);
        if (is_json || is_url_encoded ...
```

And parsing/validating JSON is not necessary in C since you'll parse that again in Lua anyway if you really use it. 

Actually, I made this work many years ago without this patch but a simple Lua script:

```lua
		self.request_content_length = env:getHeader("Content-Length")
                                  received = 0
				while received < self.request_content_length do
					x = stream:read()
					received = received + x:len()
					-- print("received " .. received)
					if not x or x:len() == 0 then break end -- read eof
					json_text = json_text .. x
				end
```

ajay-sabat:
Hi Seven,

Thanks for reviewing and your feedback. Please see my comments inline below. Please check and let me know your opinion.

> this patch changes the body logic, it will break scripts that read body with lua stream:read()

I am sorry, I don't fully understand why it will break scripts. We are setting the http body into the event body in the stream, which is unused anyway. So it shouldn't cause any conflict.
`switch_event_set_body(stream.param_event, buffer);`

> I'd like to process body only on known content types

I believe currently it's already taken care in the code. 
```
if (!strncmp(request->content_type, "application/x-www-form-urlencoded", 33)) {
} else if (!strncmp(request->content_type, "application/json", 16)) {
}
```

> And parsing/validating JSON is not necessary in C since you'll parse that again in Lua anyway if you really use it.

it's a nice validation step which I believe is ncecessary to ensure the received message is sane. If the JSON body is invalid we shouldn't even process the message and pass it to Lua. Hence we send error there itself.
```
	switch_log_printf(SWITCH_CHANNEL_LOG, SWITCH_LOG_ERROR, "Invalid JSON data in message body. content_length: %ld, body: %s \n", request->content_length, buffer);
	free(buffer);
	goto request_err;
```

Regards,
Ajay

seven1240:
1. the event body doesn't support binary data, so it will break `stream:read()` in lua which handles binary data, e.g. when you upload a binary file. And, since you already read all the body data in C, `stream:read()` in lua will read nothing, so it will make upload binary or multi-part payload not possible.
2. validating JSON ahead seems reasonable, but parse and discard the result looks like a waste since 99% time you should got a valid JSON.

ajay-sabat:
Actually I have been using http server exposed by mod_verto as REST endpoint for quite some time. Since JSON is the widely (and likely the only) used content type for REST, without JSON support in mod_verto we always hacked e.g. at one time even converted JSON to url encoded on the client side. 

I didn't know that it's possible to achieve this using stream:read() from lua script. Even though it's possible, I still think env:getBody() is the straight forward and right way. That's how we use it now. It will benefit users in general, because http server by mod_verto is actually kind of cool, just that seems incomplete. But that's my personal opinion. I will leave it up to you.

Yes it will be great if could share the Lua code to read JSON body. 

Regards,
Ajay

seven1240:
I have nothing against the idea, I just mean it should not break code.

1. the patch seems breaking the logic, it will discard the entire body except it's url encoded and json data, and their's no chance to read the body in Lua for other content-type.
2. the event body only support text body, it's fine for JSON but we should leave a door for other bodies, e.g. upload a 2G movie.
3. parsing JSON and discard looks like a waste to me, if you don't care, maybe make that optional?

ajay-sabat:
Aah ok, I think I finally understood your concern. 

Let me modify the code in a way that will work for url-encoded and JSON body and also will retain the data for other content-types so that users can do whatever they want from the script.

Regarding your concern about JSON parsing, I don't like it either. It's only for validation which can also be done in the Lua script. My initial feeling was, if we pass to the script to process when the JSON data is invalid, that's a bigger crime. So it really comes down to the cost of JSON parsing for all messages vs handling a -ve scenario.

Regards,
Ajay

ajay-sabat:
Hi Seven,

I made some changes after your comments and committed. Tested scenarios with Content-Type as 'urlencoded', 'json' and anything but these two. Please review and let me know.

Regards,
Ajay

seven1240:
Thanks for the patch. look better, but looks like:

1. You still read and discard for all content-types, there's no chance for Lua to read the body later for other content-types
2. I guess it's better to get content-type before it enter the `if` block, so you can test content-type and only enter the `if` block when content-type is json or url-encoded.

I'm not 100% sure. can you try the lua `stream:read()` code and make sure it doesn't break the original logic?

Also, it's a good practice that don't  change the indents and white spaces when you fix something, coz it makes reading the diff harder. Do white spaces in a separate commit and PR. But since you already did that I'm fine with this, since Github has an option to read diff ignoring white spaces.

ajay-sabat:
> 1. You still read and discard for all content-types, there's no chance for Lua to read the body later for other content-types
> 2. I guess it's better to get content-type before it enter the `if` block, so you can test content-type and only enter the `if` block when content-type is json or url-encoded.

Right now we are doing exactly as you suggested, reading the body only when the content-type is `application/x-www-form-urlencoded` or `application/json`. For the rest, skipping the whole if block as it was before only for `application/x-www-form-urlencoded`. Could you please take a look again?

```
        content_type_urlencoded = !strncmp(request->content_type, "application/x-www-form-urlencoded", 33);
        content_type_json = !strncmp(request->content_type, "application/json", 16);

        if (content_type_urlencoded || content_type_json) {

        }
```

> I'm not 100% sure. can you try the lua `stream:read()` code and make sure it doesn't break the original logic?

I verified using stream:read() in the Lua script for all three content types: `application/x-www-form-urlencoded`, `application/json` and everything else.

> Also, it's a good practice that don't change the indents and white spaces when you fix something, coz it makes reading the diff harder.

Sorry I wasn't aware until I saw the diff now, my editor may have done something stupid. Thanks for bringing it up, I realize it's an inconvenience and will fix it.

seven1240:
LGTM. ping @andywolk

src/mod/endpoints/mod_verto/mod_verto.c
```diff
@@ -1764,41 +1764,52 @@ static void http_run(jsock_t *jsock)
...
+		switch_bool_t content_type_json = SWITCH_FALSE;
+
+		content_type_urlencoded = !strncmp(request->content_type, "application/x-www-form-urlencoded", 33);
+		content_type_json = !strncmp(request->content_type, "application/json", 16);
+
+		if (content_type_urlencoded || content_type_json) {
+			char *buffer = NULL;
+			switch_ssize_t len = 0, bytes = 0;
+
+			if (request->content_length && request->content_length > 10 * 1024 * 1024 - 1) {
```
andywolk, line 1778: This is redundant. `request->content_length` is already checked.

src/mod/endpoints/mod_verto/mod_verto.c
```diff
@@ -1764,41 +1764,52 @@ static void http_run(jsock_t *jsock)
...
+				char *data = "HTTP/1.1 413 Request Entity Too Large\r\n"
+					"Content-Length: 0\r\n\r\n";
+				kws_raw_write(jsock->ws, data, strlen(data));
```
andywolk, line 1781: We expect a new line after one or more declarations.

src/mod/endpoints/mod_verto/mod_verto.c
```diff
@@ -1764,41 +1764,52 @@ static void http_run(jsock_t *jsock)
...
+				len = request->content_length - bytes;
 
 #define WS_BLOCK 1
 
-			if ((len = kws_raw_read(jsock->ws, buffer + bytes, len, WS_BLOCK)) < 0) {
-				switch_log_printf(SWITCH_CHANNEL_LOG, SWITCH_LOG_ERROR, "Read error %" SWITCH_SSIZE_T_FMT"\n", len);
-				goto done;
+				if ((len = kws_raw_read(jsock->ws, buffer + bytes, len, WS_BLOCK)) < 0) {
+					switch_log_printf(SWITCH_CHANNEL_LOG, SWITCH_LOG_ERROR, "Read error %" SWITCH_SSIZE_T_FMT"\n", len);
+					goto done;
```
andywolk, line 1797: This is leaking the `buffer`

andywolk (changes requested):

src/mod/endpoints/mod_verto/mod_verto.c
ajay-sabat: Fixed in the latest commit.

src/mod/endpoints/mod_verto/mod_verto.c
ajay-sabat: Fixed in the latest commit.

src/mod/endpoints/mod_verto/mod_verto.c
ajay-sabat: Changed the buffer size to hold the message body content without overflow.

ajay-sabat:
Hi @andywolk , thanks for reviewing code. I changed code to address your concerns. Please check and let me know either way.

src/mod/endpoints/mod_verto/mod_verto.c
andywolk: Is memory freed in this case?

ajay-sabat: No it wasn't, good catch. I fixed it.
Sorry initially I was hesitant to do improvement and cleanup in the original code assuming it will not be appreciated. But I will keep that in my mind next time.

PR #2271, opened 2023-10-10
Stop generating SWITCH_RTP_CRYPTO_SEND upon late media i.e ack with sdp

bhavinmoradiya447:
Describe the bug

There is an issue in freeswitch on late media usecase with SRTP. 

When Freeswitch receives an ACK with SDP (SRTP) from receiver, fs creates a new key SWITCH_RTP_CRYPTO_SEND (master key + salt) which is not shared with receiver. This is causing decryption failures at receiver end as receiver is still having older key.

To Reproduce
1.	Create a Cube to receive a call from freeswitch
2.	Execute a dial command to establishing a call from freeswitch to cube with SRTP, This Invite will have SDP with parama=crypto:7 AES_CM_128_HMAC_SHA1_80 inline:fs_encoded_key) where receiver sends 200 with SDP and FS send ACK without SDP. 
3.	Once the call established, send a RE-INVITE from cube to freeswitch without SDP (Called late media)
4.	FS sends 200OK for RE-INVITE with SDP contains the same key as in initial invite(i.e SDP with parama=crypto:7 AES_CM_128_HMAC_SHA1_80 inline:same_fs_encoded_key)
5.	Then Cube sends a ACK with SDP(SDP with parama=crypto:7 AES_CM_128_HMAC_SHA1_80 inline:cube_encoded_key)
6.	Upon on getting ACK with SDP, Freeswitch internally creates new key for its own SRTP streams, that was not shared to Cube that results in decryption failures.

Expected behaviour

Here there are two options available (RFC3264 and RFC6337),
1. FS should not change the key in this specific case and use the key which shared as part of 200 SDP. So there will not be any encryption failures on other side. 
2. Send SIP Update to cube as RE-INVITE, so that other system can get the new key to decrypt the audio packets. ( This is costly Operation as single exchange takes several seconds(rfc3711) and it adds latency on already established call where FS already shared key as part of 200 Response. 

Fix: 
 - Fix is contains changes for option#1 where FS will not generate new SWITCH_RTP_CRYPTO_SEND key incase of late media and disable_crypto_rekeying_on_late_media channel variable is set to true


src/include/switch_types.h
```diff
@@ -1630,6 +1630,7 @@ typedef enum {
 	CF_VIDEO_READ_TAPPED,
 	CF_VIDEO_WRITE_TAPPED,
 	CF_DEVICES_CHANGED,
+    CF_LATE_MEDIA,
```
dragos-oancea, line 1633: maybe name this CF_LATE_OFFER ?

src/mod/endpoints/mod_sofia/sofia.c
```diff
@@ -7496,6 +7496,10 @@ static void sofia_handle_sip_i_state(switch_core_session_t *session, int status,
 		if (r_sdp) {
 			switch_channel_set_variable(channel, SWITCH_R_SDP_VARIABLE, r_sdp);
 
+			if(ss_state == nua_callstate_ready && status == 200){
```
dragos-oancea, line 7499: indentation matters. eg: space between "if" and "(" .
https://developer.signalwire.com/freeswitch/FreeSWITCH-Explained/Community/Contributing-Code/Coding-Guidelines_6586862/

src/include/switch_types.h
bhavinmoradiya447: @dragos-oancea  Thanks for review, have addressed review comment

src/mod/endpoints/mod_sofia/sofia.c
bhavinmoradiya447: @dragos-oancea Thanks for review, have addressed review comment

PR #2273, opened 2023-10-12
[mod_conference] Avoid a race in member->read_impl access

ford-prefect:
The conference output loop accesses the member's read_impl while resampling. The output loop also spawns off the input loop thread, which is where the member->read_impl is actually initialised. This results in an uncommon race where the output loop sometimes gets an uninitialised member->read_impl and bails with logs something like this:

```
2023-10-12 18:19:01.436844 [DEBUG] conference_loop.c:1340 Setup timer soft success interval: 20  samples: 960 from codec L16
2023-10-12 18:19:01.436844 [DEBUG] conference_loop.c:1497 Outbound conference channel answered, setting CFLAG_ANSWERED
2023-10-12 18:19:01.436844 [NOTICE] switch_core_media.c:15852 Activating write resampler
2023-10-12 18:19:01.436844 [DEBUG] switch_core_media.c:16097 Engaging Write Buffer at 1920 bytes to accommodate 0->1920
2023-10-12 18:19:01.436844 [ERR] switch_core_media.c:16112 Write Buffer 0 bytes Failed!
```

As a solution, we initialise the member->read_impl even before we start up the input loop, so that this race can never happen.

### DIFF
--- a/src/mod/applications/mod_conference/conference_loop.c
+++ b/src/mod/applications/mod_conference/conference_loop.c
@@ -1316,15 +1316,15 @@ void conference_loop_output(conference_member_t *member)
 	uint32_t flush_len;
 	uint32_t low_count, bytes;
 	call_list_t *call_list, *cp;
-	switch_codec_implementation_t read_impl = { 0 }, real_read_impl = { 0 };
+	switch_codec_implementation_t real_read_impl = { 0 };
 	int sanity;
 
-	switch_core_session_get_read_impl(member->session, &read_impl);
+	switch_core_session_get_read_impl(member->session, &member->read_impl);
 	switch_core_session_get_real_read_impl(member->session, &real_read_impl);
 
 
 	channel = switch_core_session_get_channel(member->session);
-	interval = read_impl.microseconds_per_packet / 1000;
+	interval = member->read_impl.microseconds_per_packet / 1000;
 	samples = switch_samples_per_packet(member->conference->rate, interval);
 	//csamples = samples;
 	tsamples = real_read_impl.samples_per_packet;


src/mod/applications/mod_conference/conference_loop.c
```diff
@@ -823,8 +823,6 @@ void *SWITCH_THREAD_FUNC conference_loop_input(switch_thread_t *thread, void *ob
 
 	channel = switch_core_session_get_channel(session);
 
-	switch_core_session_get_read_impl(session, &member->read_impl);
```
jakubkarolczyk, line 826: why you remove this line from here?

ford-prefect: `member->read_impl` is populated by the call `conference_loop_output()` before the thread starts, so calling it again here is redundant.

ford-prefect: Yes, as far as I can tell the `read_impl` does not change after the input/output loops are started. Am I missing something?

jakubkarolczyk: https://github.com/signalwire/freeswitch/blob/master/src/mod/applications/mod_conference/mod_conference.c#L2489

ford-prefect: I see, there's another path for spawning the input loop, thanks for catching that. I've rebased and updated the patch to not drop the initialisation in the input loop.

jakubkarolczyk (approved):

src/mod/applications/mod_conference/conference_loop.c
jakubkarolczyk: ty

andywolk:
Thank you. Good catch!

PR #2278, opened 2023-10-16
[mod_sofia] specify rtp ip by variable

HowellYan:


seven1240:
can you just set rtp-ip in profile? when do you need this?

Issue #2290, opened 2023-10-25
* [Core] Fix possible deadlock in switch_core_media_set_codec() This commit results in a deadlock


cdevelop:
SHA-1: b4e43214f7c29e658e9a44239a45343d28186779

* [Core] Fix possible deadlock in switch_core_media_set_codec()

This commit results in a deadlock

Thread 3 
switch_core_session_read_frame 
	switch_mutex_lock(session->codec_read_mutex);
sofia_receive_message
	switch_mutex_lock(tech_pvt->sofia_mutex);
switch_core_media_set_codec 
	switch_core_session_lock_codec_write(session);
	switch_core_session_lock_codec_read(session);

Thread 2 
switch_core_session_write_frame
	switch_mutex_lock(session->codec_write_mutex);
sofia_receive_message
	switch_mutex_lock(tech_pvt->sofia_mutex);

```
Thread 3 (Thread 0x7f8700ddc700 (LWP 44877)):
#0  0x00007f8d324e854d in __lll_lock_wait () from /lib64/libpthread.so.0
#1  0x00007f8d324e3eb6 in _L_lock_941 () from /lib64/libpthread.so.0
#2  0x00007f8d324e3daf in pthread_mutex_lock () from /lib64/libpthread.so.0
#3  0x00007f8d351eeb49 in fspr_thread_mutex_lock (mutex=<optimized out>) at locks/unix/thread_mutex.c:92
#4  0x00007f8d3507ca25 in switch_mutex_lock (lock=<optimized out>) at src/switch_apr.c:310
#5  0x00007f8d3509a98c in switch_core_session_lock_codec_write (session=<optimized out>) at src/switch_core_codec.c:64
#6  0x00007f8d350d0746 in switch_core_media_set_codec (session=session@entry=0x7f87302945b8, force=force@entry=0, codec_flags=0) at src/switch_core_media.c:3604
#7  0x00007f8d350d604e in switch_core_media_activate_rtp (session=0x7f87302945b8) at src/switch_core_media.c:8589
#8  0x00007f8d198103ce in sofia_media_activate_rtp (tech_pvt=tech_pvt@entry=0x7f873045e8e8) at sofia_media.c:58
#9  0x00007f8d1981044e in sofia_media_tech_media (tech_pvt=tech_pvt@entry=0x7f873045e8e8, r_sdp=<optimized out>, type=type@entry=SDP_TYPE_REQUEST) at sofia_media.c:189
[...]

PR #2300, opened 2023-11-05
[core] fix session deadlock  (#2290)

cdevelop:
https://github.com/signalwire/freeswitch/issues/2290

### DIFF
--- a/src/switch_core_io.c
+++ b/src/switch_core_io.c
@@ -146,13 +146,15 @@ SWITCH_DECLARE(switch_status_t) switch_core_session_read_frame(switch_core_sessi
 	if (session->read_codec && !session->track_id && session->track_duration) {
 		if (session->read_frame_count == 0) {
 			switch_event_t *event;
-			switch_core_session_message_t msg = { 0 };
+			switch_core_session_message_t *msg = switch_core_session_alloc(session, sizeof(*msg));
 
 			session->read_frame_count = (session->read_impl.samples_per_second / session->read_impl.samples_per_packet) * session->track_duration;
 
-			msg.message_id = SWITCH_MESSAGE_HEARTBEAT_EVENT;
-			msg.numeric_arg = session->track_duration;
-			switch_core_session_receive_message(session, &msg);
+			msg->message_id = SWITCH_MESSAGE_HEARTBEAT_EVENT;
+			msg->numeric_arg = session->track_duration;
+			MESSAGE_STAMP_FFL(msg);
+			switch_core_session_queue_message(session, msg);
+
 
 			switch_event_create(&event, SWITCH_EVENT_SESSION_HEARTBEAT);
 			switch_channel_event_set_data(session->channel, event);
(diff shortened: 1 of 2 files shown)


cdevelop:
@seven1240  Teacher Du and I explained this issue in detail.

seven1240:
@andywolk cdeveop has a system which left 3-10 dead locked channels per day, which could be fixed by this patch.

Review the patch, it looks like the read_frame parts are fine as the message will be processed on the next read. But I'm worrying about the write parts. What if this happens on a sendonly channel that never has read? or on a App which never read but write ( I don't know such an App in my head though). So the message will never be processed.

Best to find out the root cause of the dead lock instead of the work around I think.

bferreirq:
Hello, any news about this commit ?

KerryRJ:
I have an installation where 30-40 inbound and outbound calls are getting deadlocked on a daily basis, and the call center agent connected to the deadlocked call also gets stuck, until a restart or the extension leg of the call is hung up.

#2387 and #2390 did not resolve the the deadlocks nor the stuck call center agent.

This fix has stopped all deadlocks and subsequently no stuck call center agents.

televoicepl:
> I have an installation where 30-40 inbound and outbound calls are getting deadlocked on a daily basis, and the call center agent connected to the deadlocked call also gets stuck, until a restart or the extension leg of the call is hung up.
> 
> #2387 and #2390 did not resolve the the deadlocks nor the stuck call center agent.
> 
> This fix has stopped all deadlocks and subsequently no stuck call center agents.

same problem in my instance. 
1.10.11 and 1.10.10 is affected. 1.10.9 is ok
Random calls are dead and only core restart helps.

shaunjstokes:
This fix has fixed all deadlocks and stuck calls in our environment.

#2372 and #2390 did not fix the issue.

bfroemel:
I can also report that this patch fixes a similar issue with stuck channels and stuck call center agents in my installation.

boteman:
Why not make it pass the Unit Tests so that it can be merged? Many people seem to be experiencing this problem.

televoicepl:
@jakubkarolczyk can you check it?

gregoriusus:
Can I bump this thread. This issue is real problem and a lot of people have this issue...

boteman:
Yesterday on Office Hours I asked about this and the related tickets which exhibit a similar problem. BKW said that it is a much deeper problem than is solved by the proposed patch and that they are refarming the code to solve everything at once. So it is a bigger effort than we thought, no ETR.

hhadzem:
Is there any progress/update related to this issue?

zrd740:
I upgraded several servers from FreeSWITCH 1.10.8 to 1.10.12. This error showed up and it's creating dozens of stuck calls per day, as well as numerous client complaints on a daily basis. 1.10.8 does not have the issue. Is there any update on this, it seems like a problem that has been going on for a long time? FusionPBX support said thie solution to the problem is to apply this patch and compile from source as a resolution to the problem. I have wanted to use the signalwire packages on debian, but they do not work due to this issue. If we can't get a solution to this I'll need to start compiling from source.

zrd740:
We've resolved the issue by migrating away from prebuilt packages and compiling FS from source with this patch, and it works perfectly. However, I'm disappointed that after almost a year, this hasn't been included in the release despite its impact on many users.

boteman:
My understanding from what the core developers have said is that this ticket and at least 3 others are a combination of problems deep down in the core that require extensive re-farming of source code to correct. They found some long-standing bugs in the process and aim to correct all of them in this effort. This patch works for some, but not others, and that is why. I'm as anxious as you for a permanent fix.

jakubkarolczyk:
Fixed in https://github.com/signalwire/freeswitch/pull/2619

PR #2309, opened 2023-11-13
[core] Add support for multiple 'or' values in time conditions

greenbea:
This commit introduces using a pipe (`|`) delimiter to match multiple 'or' values in time conditions. This enhancement will make creating common and complex time condition scenarios much simpler and easier.

For instance, to match an extension on Monday, Wednesday, and Friday during the hours 9-12 and 13-17, Up until now, you would need multiple conditions for every day and for every hour range on that day. Now, we can have something simple like this.

```xml
<extension name="ext_time_cond" continue="true">
	<condition field="destination_number" expression="^201$"/>
	<condition wday="2|4|6" hour="9-11|13-16">
		<action application="bridge" data="user/201@default"/>
	</condition>
</extension>
```

### DIFF
--- a/src/switch_xml.c
+++ b/src/switch_xml.c
@@ -3204,93 +3204,274 @@ SWITCH_DECLARE(int) switch_xml_std_datetime_check(switch_xml_t xcond, int *offse
 	}
 
 	if (time_match && xdt) {
+		char *or[256], *dup = NULL, *matched = NULL;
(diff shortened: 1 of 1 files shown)


greenbea:
I just realized freeswitch already supports comma delimiter in the time condition expression to match multiple values, so closing the pull request,

PR #2311, opened 2023-11-15
Send event on incoming ACK

zayim:
For some use cases it would beneficial to have event when ACK message is received. This PR adds support for that.


ihusejnovic:
@andywolk can you maybe take a look at this, when you have time? Thanks!

Issue #2134, opened 2023-06-29
[mod_amr] Two terminals cannot hear each other's voice when they use AMR encoding to talk


v6bit:
**Describe the bug**
We are trying AMR codec with Freeswitch 1.10.10. The result is that neither party can hear the voice.

The amazing thing is that after I set media_mix_inbound_outbound_codecs to true, one side uses AMR, and the other side uses PCMU codec, and the two can communicate normally.

**To Reproduce**
1. Enable AMR
2. Start the call
3. The call is established, but no sound is heard

**Expected behavior**
The two terminals can communicate normally through the AMR and can hear each other's voice

**Package version or git hash**
 - Version 1.10.10

**Trace logs**

dragos-oancea:
There's a README in mod_amr, make sure AMR transcoding support is actually built.  
Be aware of "Octed Aligned" versus "Bandwidth Efficient" encoding mode.
You could also toggle these vars to adjust your setup:
```
freeswitch/sip_profiles/internal.xml:182:    <param name="inbound-codec-negotiation" value="greedy"/>
freeswitch/sip_profiles/internal.xml:249:    <param name="inbound-late-negotiation" value="false"/>
```

v6bit:
> There's a README in mod_amr, make sure AMR transcoding support is actually built. Be aware of "Octed Aligned" versus "Bandwidth Efficient" encoding mode. You could also toggle these vars to adjust your setup:
> 
> ```
> freeswitch/sip_profiles/internal.xml:182:    <param name="inbound-codec-negotiation" value="greedy"/>
> freeswitch/sip_profiles/internal.xml:249:    <param name="inbound-late-negotiation" value="false"/>
> ```

I recompiled mod_amr, the steps are as follows:

----------------
rm /usr/local/freeswitch/mod/mod_amr.*
apt remove libopencore-amrnb-dev libopencore-amrnb0 libopencore-amrwb0 libvo-amrwbenc0

tar xzvf opencore-amr-0.1.6.tar.gz
tar xzvf vo-amrwbenc-0.1.3.tar.gz

cd opencore-amr-0.1.6/
./configure
make
make install

cd ../vo-amrwbenc-0.1.3/
./configure
make
make install

echo "/usr/local/lib" >> /etc/ld.so.conf
/usr/sbin/ldconfig

cd /usr/src/freeswitch/src/mod/codecs/mod_amr
make
make install
-------------------------
[...]

PR #2369, opened 2024-01-23
[mod_amr, mod_amrwb] Fix input data corruption

dkunilov:
Two similar problems with AMR and AMRWB are fixed here:
1. decode() function calls switch_amr_unpack_be/switch_amr_unpack_oa which unpacks the data in-place, thus corrupting the frame being decoded. If a READ media bug is present in the session (e.g. when using record_session), the same incoming AMR frame is decoded twice: first time for passing to the media bug and second time for passing to B-party. This results to the recorded audio being clear, but B-party hearing a distorted sound. This is described in #2134 .
The solution is to work on a copy of original encoded data.
2. init() function calls switch_separate_string() on codec->fmtp_in, which corrupts the fmtp string as a side effect. So when switch_core_codec_copy() is called, the resulting copy has different settings. The solution is to work on a copy of fmtp_in like other codecs do.

### DIFF
--- a/src/mod/codecs/mod_amrwb/mod_amrwb.c
+++ b/src/mod/codecs/mod_amrwb/mod_amrwb.c
@@ -227,12 +227,15 @@ static switch_status_t switch_amrwb_init(switch_codec_t *codec, switch_codec_fla
 			switch_assert(fmtp_dup);
 
 			argc = switch_separate_string(fmtp_dup, ';', argv, (sizeof(argv) / sizeof(argv[0])));
+
 			for (x = 0; x < argc; x++) {
 				char *data = argv[x];
 				char *arg;
+
 				while (*data && *data == ' ') {
 					data++;
 				}
+
 				if ((arg = strchr(data, '='))) {
 					*arg++ = '\0';
 					if (!strcasecmp(data, "octet-align")) {
@@ -268,7 +271,9 @@ static switch_status_t switch_amrwb_init(switch_codec_t *codec, switch_codec_fla
 					} else if (!strcasecmp(data, "mode-set")) {
 						int y, m_argc;
 						char *m_argv[SWITCH_AMRWB_MODES-1]; /* AMRWB has 9 modes */
+
 						m_argc = switch_separate_string(arg, ',', m_argv, (sizeof(m_argv) / sizeof(m_argv[0])));
+
 						for (y = 0; y < m_argc; y++) {
 							context->enc_modes |= (1 << atoi(m_argv[y]));
 							context->enc_mode = atoi(m_argv[y]);
@@ -276,6 +281,7 @@ static switch_status_t switch_amrwb_init(switch_codec_t *codec, switch_codec_fla
 					}
 				}
 			}
+
 			free(fmtp_dup);
 		}
 
@@ -409,11 +415,7 @@ static switch_status_t switch_amrwb_decode(switch_codec_t *codec,
 	unsigned char buf[SWITCH_AMRWB_OUT_MAX_SIZE];
 	uint8_t tmp[SWITCH_AMRWB_OUT_MAX_SIZE];
 
-	if (!context) {
-		return SWITCH_STATUS_FALSE;
-	}
-
-	if (encoded_data_len > SWITCH_AMRWB_OUT_MAX_SIZE) {
+	if (!context || encoded_data_len > SWITCH_AMRWB_OUT_MAX_SIZE) {
 		return SWITCH_STATUS_FALSE;
 	}
 
--- a/src/mod/codecs/mod_amr/mod_amr.c
+++ b/src/mod/codecs/mod_amr/mod_amr.c
@@ -288,12 +288,15 @@ static switch_status_t switch_amr_init(switch_codec_t *codec, switch_codec_flag_
 			switch_assert(fmtp_dup);
 
 			argc = switch_separate_string(fmtp_dup, ';', argv, (sizeof(argv) / sizeof(argv[0])));
+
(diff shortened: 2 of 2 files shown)


dragos-oancea (approved):

src/mod/codecs/mod_amrwb/mod_amrwb.c
```diff
@@ -401,13 +406,15 @@ static switch_status_t switch_amrwb_decode(switch_codec_t *codec,
...
 	struct amrwb_context *context = codec->private_info;
-	unsigned char *buf = encoded_data;
+	unsigned char buf[SWITCH_AMRWB_OUT_MAX_SIZE];
 	uint8_t tmp[SWITCH_AMRWB_OUT_MAX_SIZE];
 
 	if (!context) {
 		return SWITCH_STATUS_FALSE;
 	}
 
+	memcpy(buf, encoded_data, encoded_data_len);
```
andywolk, line 416: Potential stack overflow if `encoded_data_len > SWITCH_AMRWB_OUT_MAX_SIZE` by any reason.

andywolk (changes requested):

src/mod/codecs/mod_amrwb/mod_amrwb.c
dkunilov: As input frames are never supposed to be larger than SWITCH_AMRWB_OUT_MAX_SIZE and it is really an error case, would you be okay with returning SWITCH_STATUS_FALSE and not processing such frames?

andywolk: Yeah. Let's drop it.

dkunilov: Added extra checks of encoded_data_len

dragos-oancea (approved):

andywolk (approved):

jakubkarolczyk (approved):

src/mod/codecs/mod_amrwb/mod_amrwb.c
olofssonanton: @andywolk @dkunilov I've detected a case where the `encoded_data_len > SWITCH_AMRWB_OUT_MAX_SIZE` comparison causes calls to be hung up during a re-INVITE where the other party responds with multiple codecs in their 200 OK.

Alice and Bob are in a call using PCMA, and FreeSWITCH sends a re-INVITE to Alice. Alice responds with all codecs she accepts, with AMR-WB at the top. FreeSWITCH momentarily assumes the call is switching to AMR-WB, but Alice continues to send PCMA. With the changes from this PR, FreeSWITCH hangs up the call after `switch_amrwb_decode` returns `SWITCH_STATUS_FALSE`. Here's a log excerpt:

> 2024-04-23 13:12:44.593843 91.27% [DEBUG] sofia.c:8524 Processing updated SDP
2024-04-23 13:12:44.593843 91.27% [DEBUG] switch_core_media.c:8654 Audio params are unchanged for sofia/foo/Alice@X.X.X.X.
2024-04-23 13:12:44.593843 91.27% [DEBUG] switch_core_media.c:3644 Changing Codec from PCMA@20ms@8000hz to AMR-WB@20ms@16000hz
2024-04-23 13:12:44.613817 91.27% [DEBUG] switch_rtp.c:4437 RE-Starting timer [soft] 320 bytes per 20ms
2024-04-23 13:12:44.613817 91.27% [DEBUG] switch_core_media.c:3743 Set Codec sofia/foo/Alice@X.X.X.X AMR-WB/16000 20 ms 320 samples 23850 bits 1 channels
2024-04-23 13:12:44.613817 91.27% [DEBUG] switch_core_codec.c:124 sofia/foo/Alice@X.X.X.X Original read codec replaced with AMR-WB:110
2024-04-23 13:12:44.654008 91.27% [DEBUG] switch_core_media.c:3199 alternate payload received (received 8, expecting 96)
2024-04-23 13:12:44.654008 91.27% [WARNING] switch_core_media.c:3210 Changing current codec to PCMA (payload type 8).
2024-04-23 13:12:44.654008 91.27% [ERR] switch_core_media.c:16066 Codec AMR-WB / Bandwidth Efficient decoder error!
2024-04-23 13:12:44.654008 91.27% [DEBUG] switch_ivr_bridge.c:819 sofia/foo/Bob ending bridge by request from write function

I have not yet deduced what the encoded_data actually contains, if it actually inserts the PCMA payload into the AMR-WB decoder or what's going on, but I have tested the same case without mod_amrwb loaded (so it momentarily chooses g722) and that works perfectly.

olofssonanton: A description of how I encountered this:

1. Incoming call from Alice to FreeSWITCH, with Alice offering AMR-WB, G722, AMR, PCMA, PCMU.
2. Alice is bridged to Bob, using late negotiation.
3. Bob answers with PCMA.
4. FreeSWITCH connects Bob with Alice, both using PCMA.
5. FreeSWITCH sends re-INVITEs to both Alice and Bob, containing only PCMA.
6. Alice detects no change in SDP, so she answers with her previous offer (the initial offer, containing _all_ codecs).
7. FreeSWITCH seemingly tries to switch to AMR-WB, and the call breaks.

Alice's behaviour in step 6 is a bit puzzling to me, but I see nothing in any RFC which says it is incorrect.

olofssonanton: @andywolk @dkunilov I found out that the above re-INVITE sequence actually resulted in a PCMA payload being inserted into `encoded_data`. A fix for this is available in #2452.

Issue #2361, opened 2024-01-17
Missing marker bit when switching from generated ringback to actual media


olofssonanton:
**Describe the bug**
When using `ringback` and `ignore_early_media`, there is a brief gap in packets sent to the A-part after receiving `200 OK` from B. In order to prevent this gap from causing an unnecessary slow-down in playback for A (due to jitter buffer), there should be a marker bit inserted when the media sent to A switches from generated ringback to actual media from B.

See below image for a screenshot with comments from that exact moment in a pcap.

<img width="1670" alt="image" src="https://github.com/signalwire/freeswitch/assets/6662439/9357a34d-2ac2-4917-8f64-e4fa88ecdb40">

**To Reproduce**
1. Using ringback and ignore_early_media, like `ringback=%(1000,4000,425.0)` and `ignore_early_media=true`
2. Make a call using Verto
3. Using Sofia, bridge it to an endpoint which will send early media
4. A brief gap in packets received can be observed for A. This can be detected by tracing, or looking at `packetsReceived/s` in chrome://webrtc-internals under "Stats graphs for inbound-rtp".

**Expected behavior**
Either no gap in packets being sent, or a market bit in the first packet which is sent after the gap.

**Package version or git hash**
 - Version 1.10.10

**Trace logs**
pcap file of the related media streams has been sent to Brian West.

olofssonanton:
All of this is 100% just a matter of a missing marker bit when A's ringback stream is switched for the actual media stream from B - it's super apparent when adding `instant_ringback` into the mix.

The call starts bridging, and generated ringback immediately starts being sent to A with a timestamp starting from 0.
After `td` amount of time, the B side then responds with early media, also with timestamps starting from 0.
These two streams then have a more or less constant timestamp delta of `td` between them.
When B answers, FreeSWITCH discards the ringback stream, which was currently at timestamp `tA`, and A starts receiving media from B, currently at timestamp `tB`.
These new packets, `n` (starting from 0), suddenly have a timestamp of `tA - td + (n * samples_per_interval)`, i.e. they are `td` amount of time old from A's perspective.
Since A has no idea that this is a new stream (as no SSRC has changed either), all packets from time `tA` until `tA + td`, when `tB + td ≥ tA`, will just look like duplicates.

In summary, a marker bit must be inserted when switching streams. Otherwise the receiver has no idea what just happened.

olofssonanton:
[...]

PR #2370, opened 2024-01-23
Add RTP bug flag "RTP_BUG_CHANGE_SSRC_ON_MARKER_ALSO_WHEN_SECURE" to allow SSRC change when using SRTP

olofssonanton:
Fixes signalwire/freeswitch#2361, where a jump backwards in RTP timestamps caused silence at the start of calls when switching from generated ringback to remote media.

### DIFF
--- a/src/switch_rtp.c
+++ b/src/switch_rtp.c
@@ -8215,7 +8215,7 @@ static int rtp_common_write(switch_rtp_t *rtp_session,
 			int delta = rtp_session->ts - rtp_session->last_write_ts;
 
 			if (!rtp_session->flags[SWITCH_RTP_FLAG_UDPTL] &&
-				((!rtp_session->flags[SWITCH_RTP_FLAG_RESET] && (abs(delta) > rtp_session->samples_per_interval * 10))
+				((rtp_session->flags[SWITCH_RTP_FLAG_RESET] && (abs(delta) > rtp_session->samples_per_interval * 10))
 				|| rtp_session->ts == rtp_session->samples_per_interval)) {
 				m++;
 			}
@@ -8251,7 +8251,8 @@ static int rtp_common_write(switch_rtp_t *rtp_session,
 		}
 
 		/* If the marker was set, and the timestamp seems to have started over - set a new SSRC, to indicate this is a new stream */
-		if (m && !switch_rtp_test_flag(rtp_session, SWITCH_RTP_FLAG_SECURE_SEND) && (rtp_session->rtp_bugs & RTP_BUG_CHANGE_SSRC_ON_MARKER) &&
+		if (m && (rtp_session->rtp_bugs & RTP_BUG_CHANGE_SSRC_ON_MARKER) &&
+			(!switch_rtp_test_flag(rtp_session, SWITCH_RTP_FLAG_SECURE_SEND) || rtp_session->rtp_bugs & RTP_BUG_CHANGE_SSRC_ON_MARKER_ALSO_WHEN_SECURE) &&
 			(rtp_session->flags[SWITCH_RTP_FLAG_RESET] || (rtp_session->ts <= rtp_session->last_write_ts && rtp_session->last_write_ts > 0))) {
 			switch_rtp_set_ssrc(rtp_session, (uint32_t) ((intptr_t) rtp_session + (switch_time_t) switch_epoch_time_now(NULL)));
 		}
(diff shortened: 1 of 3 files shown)


src/switch_rtp.c
```diff
@@ -8215,7 +8215,7 @@ static int rtp_common_write(switch_rtp_t *rtp_session,
 			int delta = rtp_session->ts - rtp_session->last_write_ts;
 
 			if (!rtp_session->flags[SWITCH_RTP_FLAG_UDPTL] &&
-				((!rtp_session->flags[SWITCH_RTP_FLAG_RESET] && (abs(delta) > rtp_session->samples_per_interval * 10))
```
olofssonanton, line 8218: This looks like an accidental leftover from testing - let me have a look

olofssonanton, line 8218: This condition looked backwards to me. The inverse will successfully trigger a marker bit insertion. I'd love to hear some thoughts on this.

src/switch_rtp.c
```diff
@@ -8251,7 +8251,8 @@ static int rtp_common_write(switch_rtp_t *rtp_session,
 		}
 
 		/* If the marker was set, and the timestamp seems to have started over - set a new SSRC, to indicate this is a new stream */
-		if (m && !switch_rtp_test_flag(rtp_session, SWITCH_RTP_FLAG_SECURE_SEND) && (rtp_session->rtp_bugs & RTP_BUG_CHANGE_SSRC_ON_MARKER) &&
```
olofssonanton, line 8254: Are there any restrictions on changing SSRC in SRTP that I am unaware of?

jakubkarolczyk: there are legitimate reasons to not do so.

src/switch_rtp.c
jakubkarolczyk: From your last comment to the connected issue I understood that SSRC change is enough to resolve the problem you observe. So why would you like to instert m bit?

olofssonanton: From my testing, both the marker bit and the SSRC change is necessary to fix the issue.

I see now when reading my previous comments that it seems like I discarded the marker bit entirely and solved it with just the SSRC, but no, both of these changes are needed for this case to be solved. Sorry about that!

jchavanton:
I am also facing similar one way audios with FS, webrtc getstats are reporting discarded packets.

Any reason why the review was not completed ?

@jakubkarolczyk  can you detail your statement 
"there are legitimate reasons to not do so."
Are you saying we can not change the SSRC with SRTP ?

src/switch_rtp.c
jchavanton: @jakubkarolczyk can you detail your statement
"there are legitimate reasons to not do so."
Are you saying we can not change the SSRC with SRTP ?

jchavanton: Looking at your comment 
"From your last comment to the connected issue I understood that SSRC change is enough to resolve the problem you observe. So why would you like to instert m bit?"

I guess you are still fine with that, but you just wanted to highlight that there may be issues.

I may dig around that fix and see is this is solving my issues, I just wanted to know why this ended up not being merged/reviewed.

jchavanton:
I was able to catch a trace of a once way audio where webrtc is discarding all the packets.
Seems like we can confirm the existence of the problem once more.

image

jchavanton:
I am looking at some traces on RTP streams where WebRTC discarded most packets, just sharing initial observation 

image

WebRTC getstats report :
`recv[11635]lost[0%]jitter[0.01]energy[0.08]discarded[11562]`
11635-11562= 173 (not very far from the first jump/markbit)

davehorton:
just wondering what became of this....I am seeing the same issue described in #2361 , if this was not the accepted fix is there another?

olofssonanton:
Hey @davehorton (thanks for drachtio, btw)!

We've been running this for more than a year now and it has mitigated the issues we've seen caused by generated ringback. It may not be the perfect fix, but it does the trick.

Lately I've been investigating a few tough bugs we've seen crop up in related areas in FreeSWITCH, and the media and timing handling does unfortunately seem to be a common point of failure in complex scenarios such as combining generated media/audio file playback, transcoding, recording, etc. 

But I'm keeping my fingers crossed as @briankwest did mention some upcoming work in these areas some time back.

PR #2378, opened 2024-02-03
Combined the master and base docker images to create an updated, minimized image

king-dopey:
Most of the concepts are taken from the docker images in the master and base_image folders. It's now all done in one Dockerfile, using buildkit and a staging image. It also includes build hooks so the Token can be added as an environment variable, on Docker Hub, for automated builds and security scans.
Other enhancements to the updated docker image:

1. Moved from from tar to cp and shell scripting for consistency and correct racing conditions documented in the shell script
2. Added erlang-base, curl, and ca-certificates
3. Copied ca-certificates, so Curl functions properly (including internally to freeswitch)
4. Fixed sound file downloading
5. Increased the complexity of the default generated password
6. Merged sudo process, but moved to su-exec from gosu, to clear vulnerabilities
7. Update Debian distro before freeswitch install, to keep dependencies fully up to date
8. Updated to the latest Busybox

This is currently being auto-built at https://hub.docker.com/r/dheaps/freeswitch.


king-dopey:
The PR drone seem to consistently fail when another commit is added to the PR, on the notify step, not on an actual test. Anyhow, those unit tests are irrelevant to this PR and no code being tested was updated.

PR #2387, opened 2024-02-15
[mod_sofia] Fixing deadlock in issue #2290

mishehu:
I cannot certify that on SWITCH_MESSAGE_INDICATE_TRANSCODING_NECESSARY that we should be following this code path, but it does not look like there's anything that gets changed when we enter sofia_receive_message() in this case.  Thus there's no sense in locking tech_pvt->sofia_mutex , thus avoiding the deadlock described in issue #2290.


bferreirq (approved):

andywolk:
Closing this as already fixed by #2619

PR #2392, opened 2024-02-21
[Core] Remove mod_kazoo from tree

andywolk:


crienzo (approved):

Outme1:
Hello, what is the reason for removing mod_kazoo? 

And is the self build mod_kazoo still usable after this?

Thank you

crienzo:
Hello. mod_kazoo is a contribution by the 2600hz team. We've moved it to https://github.com/freeswitch/mod_kazoo so that 2600hz (and the larger community) can more easily maintain the module. Last I heard, the module works fine with latest FreeSWITCH.

PR #2413, opened 2024-04-01
Fix ODBC errors replace SQLPrepare() -> SQLExecDirect()

NormB:
From the ODBC documentation located here: https://dev.mysql.com/doc/connector-odbc/en/connector-odbc-configuration-connection-parameters.html

MULTI_STATEMENTS - Allow multiple statements - 67108864

Enables support for batched statements. As of 8.0.24, preparing a query with multiple statements raises an error. The direct execution of parameter-less statements prepared using the SQLPrepare() function is not supported. Multiple statements can only be executed through the SQLExecDirec() ODBC function.


andywolk:
Why is `SQLExecute` kept after replacing `SQLPrepare` with `SQLExecDirect`?

PR #2429, opened 2024-04-12
fixed eavesdrop channel management via ESL

sergey-safarov:
The current `switch_ivr_eavesdrop_session` implementation does not call `switch_ivr_parse_all_events` and leads to the case when executing `eavesdrop` application blocks any command execution sent via ESL.

This PR fix this.
I have tested locally.

### DIFF
--- a/src/switch_ivr_async.c
+++ b/src/switch_ivr_async.c
@@ -2816,6 +2816,9 @@ SWITCH_DECLARE(switch_status_t) switch_ivr_eavesdrop_session(switch_core_session
 				}
 				switch_buffer_unlock(ep->buffer);
 			}
+
+			switch_ivr_parse_all_events(session);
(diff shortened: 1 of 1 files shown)


sergey-safarov:
The issue can be reproduced using this test script
```perl
#!/bin/perl

use ESL;

my $uuid = shift;
my $args = join(" ", @ARGV);

my $con = new ESL::ESLconnection("127.0.0.1", "8021", "ClueCon");

my $e = $con->execute('info', '', $uuid);
```

need to call the script using commands like
```sh
./test.pl a0aac03e-0956-4adc-82b5-08a19b6155be
```

Without this patch channels are not expected for calls with active `eavesdrop` applications.
With this patch channel variables will be printed into the FreeSwitch log.

PR #2452, opened 2024-05-02
[mod_sofia] Prevent unnecessary re-processing of unmodified SDP

olofssonanton:
Always clear flag TFLAG_NEW_SDP. 

This prevents re-processing of unmodified SDP during re-INVITEs. Before this change a remote party's UNMODIFIED answer to a re-INVITE would always be re-processed once, due to TFLAG_NEW_SDP still being set. Any subsequent re-INVITE thereafter would then behave as expected, as TFLAG_NEW_SDP would then have been cleared.


olofssonanton:
This fixes an issue where an unexpected payload is inserted into the wrong decoder after momentarily switching codecs (even though there was no need). This was made apparent after changes made in #2369, as mentioned [here](https://github.com/signalwire/freeswitch/pull/2369#discussion_r1576294542).

PR #2456, opened 2024-05-06
Fix logical error in rtp engine check in dtls flow

SouradeepBera:
For an inbound call leg requiring SRTP over DTLS, switch_channel_answer does not return. It gets stuck in an infinite do-while loop in switch_core_media_check_dtls()
`do {
	if (engine->rtp_session) checking = check_engine(engine);
} while (switch_channel_ready(session->channel) && checking);
`
The check_engine() function should return 0 on success to break the loop. However. in the second half of the function it flips the logic and returns 0 if it is not acceptable and 1 otherwise, resulting in the infinite loop

### DIFF
--- a/src/switch_core_media.c
+++ b/src/switch_core_media.c
@@ -12529,7 +12529,7 @@ static int check_engine(switch_rtp_engine_t *engine)
 
 	status = switch_rtp_zerocopy_read_frame(engine->rtp_session, &engine->read_frame, flags);
 
-	if (!SWITCH_READ_ACCEPTABLE(status)) {
+	if (SWITCH_READ_ACCEPTABLE(status)) {
 		return 0;
 	}
 


ChrisZhangJin:
Hi, we experienced the same issue now!!  Are you sure the change you made can solve this issue??   I saw your pr was not accepted.  @SouradeepBera 

> For an inbound call leg requiring SRTP over DTLS, switch_channel_answer does not return. It gets stuck in an infinite do-while loop in switch_core_media_check_dtls() `do { if (engine->rtp_session) checking = check_engine(engine); } while (switch_channel_ready(session->channel) && checking); ` The check_engine() function should return 0 on success to break the loop. However. in the second half of the function it flips the logic and returns 0 if it is not acceptable and 1 otherwise, resulting in the infinite loop

SouradeepBera:
Seems to work for us @ChrisZhangJin

SouradeepBera:
@andywolk  can you please review this fix?

PR #2488, opened 2024-06-14
hashtable memset issue fix

akscf:
Hi there, here seems is a mistake:
```c
switch_hashtable.c:129:
memset(newtable[h->tablelength], 0, newsize - h->tablelength);
```

if plan-b have ever happen, that lead to permanent crash, should be:
```c
memset(&newtable[h->tablelength], 0, ((newsize - h->tablelength) * sizeof(struct entry*)));
```

### DIFF
--- a/src/switch_hashtable.c
+++ b/src/switch_hashtable.c
@@ -126,7 +126,7 @@ hashtable_expand(switch_hashtable_t *h)
 				realloc(h->table, newsize * sizeof(struct entry *));
 			if (NULL == newtable) { (h->primeindex)--; return 0; }
 			h->table = newtable;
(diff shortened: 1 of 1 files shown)


andywolk (changes requested):
No need for extra parenthesis. Good catch btw.

andywolk (dismissed):

andywolk:
You should be careful making Pull Requests from your fork's master branch. Each time you change your master affects this PR.

akscf:
Got it, sorry, won't do it again.
I was a little dumb when creating branches, this cleaning have to be done ;)
Now, seems, realized the right way...finally

PR #2541, opened 2024-07-28
[mod_spandsp] Compatible with the new version of spandsp.

azzbcc:


andywolk:
Thank you. Will this work on Windows?

azzbcc:
I don't have windows environment.

ThomasLobker:
Thanks so much for sharing @azzbcc 

This works for me when building on Ubuntu 24.04 with the latest version of spandsp.

pfournier:
While this patch allows FreeSWITCH to compile with the newest spandsp code base, it won't be compatible with older spandsp anymore. But furthermore, it passes a null pointer as the handler to use by spandsp which doesn't check if the handler is set before calling it, so in case of V.18 mode change, FreeSWITCH would crash with a call to null-pointer.

A more complete pull request is available at #2668.

Issue #2226, opened 2023-09-01
Incorrect OPUS negotiation


admin-toneca:
**Describe the bug**
FreeSWITCH does not negotiate opus with different rate

**To Reproduce**
1. Make call with OPUS/48000 on A leg and OPUS/16000 on B leg.
2. FreeSWITCH will select PCMU codec 

**Package version or git hash**
 - It has happened since version 1.10.10 and has not reproduced on FreeSWITCH 1.10.9.

```
2023-09-01 17:47:56.424983 98.23% [DEBUG] switch_core_media.c:5526 Audio Codec Compare [opus:102:48000:20:0:1]/[opus:116:48000:20:0:2]
2023-09-01 17:47:56.424983 98.23% [DEBUG] switch_core_media.c:5569 Audio Codec Compare [opus:116:48000:20:0:2] is saved as a near-match
2023-09-01 17:47:56.424983 98.23% [DEBUG] switch_core_media.c:5526 Audio Codec Compare [opus:102:16000:20:0:1]/[PCMU:0:8000:20:64000:1]
2023-09-01 17:47:56.424983 98.23% [DEBUG] switch_core_media.c:5526 Audio Codec Compare [opus:102:16000:20:0:1]/[PCMA:8:8000:20:64000:1]
2023-09-01 17:47:56.424983 98.23% [DEBUG] switch_core_media.c:5526 Audio Codec Compare [PCMU:0:8000:20:64000:1]/[opus:116:48000:20:0:2]
2023-09-01 17:47:56.424983 98.23% [DEBUG] switch_core_media.c:5526 Audio Codec Compare [PCMU:0:8000:20:64000:1]/[PCMU:0:8000:20:64000:1]
2023-09-01 17:47:56.424983 98.23% [DEBUG] switch_core_media.c:5588 Audio Codec Compare [PCMU:0:8000:20:64000:1] ++++ is saved as a match
2023-09-01 17:47:56.424983 98.23% [DEBUG] switch_core_media.c:5526 Audio Codec Compare [PCMU:0:8000:20:64000:1]/[PCMA:8:8000:20:64000:1]
2023-09-01 17:47:56.424983 98.23% [DEBUG] switch_core_media.c:5526 Audio Codec Compare [PCMA:8:8000:20:64000:1]/[opus:116:48000:20:0:2]
[...]

PR #2582, opened 2024-08-26
Fix bug #2226

magiclin99:
bug link:
https://github.com/signalwire/freeswitch/issues/2226


shaunjstokes:
We've found that the patch https://github.com/signalwire/freeswitch/pull/2582 has the side effect of causing calls that fail to negotiate the same codec as Leg A for Leg B to fail with SIP 488 INCOMPATIBLE_DESTINATION. That's not an option for us, we need backwards compatibility with other codecs.

Rather than apply this for all codecs it should only apply for Opus. Now calls using Opus and other codecs establish correctly on both legs, and there are no issues with codec negotiation when the codecs on Leg A and Leg B don't match.

See PR #2623

PR #2597, opened 2024-09-13
no need to unlock and lock again

yutongqing:
the lock session->bug_rwlock has already been locked in write mode, no need to unlock and then lock again


seven1240:
I think it is intended. So it can release the CPU and other threads would have a chance to get that lock and do sth. useful. Actually, maybe better to add a switch_yield(1) before the next lock so it would be better to tell the kernel that the current thread would like to release the CPU.

PR #2599, opened 2024-09-17
Correct logging for a debug message

MarcGuiselin:
When debugging is enabled, this message is being logged as "critical" when it should be debug (I think).


seven1240 (approved):

MarcGuiselin:
Anything blocking this from being merged? I'm still seeing it requires review.

PR #2603, opened 2024-09-23
[GHA] Set proper apt repositories for releases.

s3rj1k:


andywolk:
This is not required.

PR #2623, opened 2024-10-14
Update switch_core_media.c - Patch for #2226

shaunjstokes:
Patch for #2226

### DIFF
--- a/src/switch_core_media.c
+++ b/src/switch_core_media.c
@@ -5517,6 +5517,10 @@ SWITCH_DECLARE(uint8_t) switch_core_media_negotiate_sdp(switch_core_session_t *s
 					uint32_t bit_rate = imp->bits_per_second;
 					uint32_t codec_rate = imp->samples_per_second;
 
+					if (!strcasecmp(map->rm_encoding, "opus")) {
+						codec_rate = imp->actual_samples_per_second;
+					}
+
 					if (imp->codec_type != SWITCH_CODEC_TYPE_AUDIO) {
 						continue;
 					}


siul72:
We are facing a similar issue on an armel build where A leg is using PCMA and B leg used Opus.
It as working on Version 1.10.3 (the latest one tested), not anymore on 1.10.12.
After using @shaunjstokes patch we got back :Audio Codec Compare [opus:116:48000:20:0:1] ++++ is saved as a match avoiding the INCOMPATIBLE DESTINATION from the server side.

celliso1:
It appears this PR has revealed a bug in the opus encoder affecting the timestamp increment. The RFC 7587 [states ](https://datatracker.ietf.org/doc/html/rfc7587#section-4.1) opus always increments at 48kHz regardless of the payload sample rate, but mod_opus.c is using 16 kHz or 8kHz samples per packet. See comments in https://github.com/signalwire/freeswitch/issues/2226#issuecomment-2749654387.

I can't say if this has any other impact, but it does fix the timestamp increment:

```c
diff --git a/src/mod/codecs/mod_opus/mod_opus.c b/src/mod/codecs/mod_opus/mod_opus.c
index 96933e6ea5..5f7972724a 100644
--- a/src/mod/codecs/mod_opus/mod_opus.c
+++ b/src/mod/codecs/mod_opus/mod_opus.c
@@ -1487,7 +1487,7 @@ SWITCH_MODULE_LOAD_FUNCTION(mod_opus_load)
        }

        /* 16 khz */
-       samples = 160;
+       samples = 480; /* Increment RTP timestamp as if 48 kHz */
        bytes = 320;
        mss = 10000;
        rate = 16000;
@@ -1584,7 +1584,7 @@ SWITCH_MODULE_LOAD_FUNCTION(mod_opus_load)
        }

        /* 8 khz */
-       samples = 80;
+       samples = 480; /* Increment RTP timestamp as if 48 kHz */
        bytes = 160;
        mss = 10000;
        rate = 8000;
```

Issue #2630, opened 2024-10-22
mod_sofia: no audio after 200 OK with different port in SDP from 183 Session in Progress


wmasilva:
Making a call to a remote side that sends RTP in sessions in progress but when it answers there is no audio. 
The problem is caused that it sends in the 200 OK SDP a different port number and crypto key, that SDP is not processed by FS. 

I could reproduce the issue using sipp escenario with plain RTP and using the parameter: 
NOTE: with plain RTP FS can ajust to the correct port or IP using the rtp_auto_adjust parameter, but with SRTP is not possible since the key is diferent from the SIP/183 SDP. 

I found old debates about this issue:
https://forum.signalwire.community/t/freeswitch-ignores-sdp-in-200-ok-after-183/748
https://freeswitch-dev.freeswitch.narkive.com/LoQ0LunD/fs-doesn-t-update-rtp-port-when-sdp-changed-in-200ok-response-for-invite-message

knowing that is not a rfc behaviour (it should send a re-invite first i think), in the code it looks that is supported, in sofia.c, line:8501

```
		if (!switch_channel_test_flag(channel, CF_PROXY_MODE) && !switch_channel_test_flag(channel, CF_PROXY_MEDIA) &&
			r_sdp && (!is_dup_sdp || sofia_test_flag(tech_pvt, TFLAG_NEW_SDP)) && switch_core_media_ready(tech_pvt->session, SWITCH_MEDIA_TYPE_AUDIO) && !sofia_test_flag(tech_pvt, TFLAG_NOSDP_REINVITE)) {
			/* sdp changed since 18X w sdp, we're supposed to ignore it but we, of course, were pressured into supporting it */
			uint8_t match = 0;

			sofia_clear_flag(tech_pvt, TFLAG_NEW_SDP);
			switch_channel_set_flag(tech_pvt->channel, CF_REINVITE);
```

From the results, it looks taht this only applies to aleg and not to the remote side (bleg), after answer i see: 

```
2024-10-22 12:22:42.256892 97.97% [DEBUG] sofia.c:8519 Processing updated SDP
2024-10-22 12:22:42.256892 97.97% [DEBUG] switch_core_media.c:8640 Audio params are unchanged for sofia/main/1000@default.com:5060.
```

It shows info only for channel 1000@default.com and not for the 00310@192.168.10.1:6070, the one that had changed SDP in the 200 OK. 
[...]

PR #2631, opened 2024-10-22
mod_sofia: fix no audio after early media when different SDP in 200OK

wmasilva:
Solve #2630

### DIFF
--- a/src/mod/endpoints/mod_sofia/sofia.c
+++ b/src/mod/endpoints/mod_sofia/sofia.c
@@ -8498,6 +8498,12 @@ static void sofia_handle_sip_i_state(switch_core_session_t *session, int status,
 			}
 		break;
 	case nua_callstate_ready:
+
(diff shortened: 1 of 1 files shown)


wmasilva:
Unfortunately, with my first patch, i got incorrect behavior on calls made with the option “inherit_codec=true” and using late negotiation.
FreeSWITCH responds in the 200 OK SDP to the A-leg with the codec that resulted from the inheritance, but then it generates RTP using the codec from the A-leg’s INVITE SDP. 

I've update the PR with a different approach, the detection occurs in the ready state. 
I couldn't see any issue during the tests with and without  inherit_codec. 

Here is a log after the change, it detects correctly when there is a change. 
[freeswitch-pr2631.log](https://github.com/user-attachments/files/23106751/freeswitch-pr2631.log)

And corresponding sipp to test it. 
[sipp-pr2631.xml](https://github.com/user-attachments/files/23106776/sipp-pr2631.xml)

PR #2642, opened 2024-11-05
[mod_verto] Fix memory leak by correctly freeing regex

Ansuel:
For mod_verto regex was never freed and was actually leaking memory. Correctly free the compiled regex to fix the memory leak.


andywolk:
Please never mind the failing CI. We are internally working on the workflow.

seven1240 (approved):

andywolk (approved):

PR #2668, opened 2024-12-17
[mod_spandsp] Fix compilation against >=2023/06/02 spandsp

pfournier:
This patch from my colleague allows FreeSWITCH to compile with latest or earlier spandsp code base:

spandsp, beginning with commit d9681c37 and coinciding with the SPANDSP_RELEASE_DATE of 20230620, introduced the following changes to its V.18 protocol API, which FreeSWITCH is not able to compile against:
- Certain V.18 constants were renamed.
- The v18_init function now requires passing a third function, handling the V.18 modem's status changes.

This patch allows FreeSWITCH to build against current versions of spandsp by:
- Using the new V.18 constant names.
- Implementing a simple status reporter callback function and passing it as the third function to v18_init.

Additionally, it retains backward compatibility with prior versions of spandp through #if conditions checking the value of SPANDSP_RELEASE_DATE.

### DIFF
--- a/src/mod/applications/mod_spandsp/mod_spandsp_dsp.c
+++ b/src/mod/applications/mod_spandsp/mod_spandsp_dsp.c
@@ -152,17 +152,28 @@ static void put_text_msg(void *user_data, const uint8_t *msg, int len)
 
 }
 
+#if SPANDSP_RELEASE_DATE >= 20230620
+static void handle_v18_status(void *user_data, int status)
+{
+	switch_tdd_t *pvt = (switch_tdd_t *) user_data;
+	switch_channel_t *channel = switch_core_session_get_channel(pvt->session);
+
+	switch_log_printf(SWITCH_CHANNEL_SESSION_LOG(pvt->session), SWITCH_LOG_INFO, "%s detected V.18 modem: %s\n", switch_channel_get_name(channel), v18_status_to_str(status));
+
+}
+#endif
+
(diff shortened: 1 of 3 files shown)


src/mod/applications/mod_spandsp/mod_spandsp_dsp.c
```diff
@@ -213,8 +224,11 @@ switch_status_t spandsp_tdd_send_session(switch_core_session_t *session, const c
 		return SWITCH_STATUS_FALSE;
 	}
 
+#if SPANDSP_RELEASE_DATE >= 20230620
+	tdd_state = v18_init(NULL, TRUE, get_v18_mode(session), V18_AUTOMODING_GLOBAL, put_text_msg, NULL, handle_v18_status, NULL);
```
andywolk, line 228: user data is null and will crash in handle_v18_status due to a null pointer dereference?

andywolk (changes requested):

src/mod/applications/mod_spandsp/mod_spandsp_dsp.c
andywolk: @pfournier ping

morbit85 (approved):

andywolk (approved):

pfournier:
Thanks Andy for catching that, I approve your changes.

Issue #2202, opened 2023-08-15
mod_av: avformat.c lines 458-459 reference private field within AVOutputFormat


Korynkai:
The following code is apparently referencing a private field within the `AVOutputFormat` structure of libavformat, which is described as "[not] part of the public API", "may not be used outside of libavformat" and "can be changed and removed at will":

https://github.com/signalwire/freeswitch/blob/b74245d48a1f65a05e853f24e973f9b9ff35f8f5/src/mod/applications/mod_av/avformat.c#L458
https://github.com/signalwire/freeswitch/blob/b74245d48a1f65a05e853f24e973f9b9ff35f8f5/src/mod/applications/mod_av/avformat.c#L459

As described here:
https://github.com/FFmpeg/FFmpeg/blob/566aa38d98f5f492995127e82ab9a516f59bf952/libavformat/avformat.h#L540

```
typedef struct AVOutputFormat {

...continued...

    /*****************************************************************
     * No fields below this line are part of the public API. They
     * may not be used outside of libavformat and can be changed and
     * removed at will.
     * New public fields should be added right above.
     *****************************************************************
     */
    /**
     * size of private data so that it can be allocated in the wrapper
     */
    int priv_data_size;

...continued...

} AVOutputFormat;
```

mod_av, as-is, causes the following error when building against FFmpeg 6.0, as this field has apparently been removed from FFmpeg 6.0:

```
avformat.c: In function 'mod_avformat_alloc_output_context2':
avformat.c:458:23: error: 'const struct AVOutputFormat' has no member named 'priv_data_size'
  458 |         if (s->oformat->priv_data_size > 0) {
      |                       ^~
avformat.c:459:53: error: 'const struct AVOutputFormat' has no member named 'priv_data_size'
  459 |                 s->priv_data = av_mallocz(s->oformat->priv_data_size);
      |                                                     ^~
make[4]: *** [Makefile:1026: libavmod_la-avformat.lo] Error 1
```

With the following patch, everything seems to build correctly:
```
diff -Naur freeswitch-1.10.10.pristine/src/mod/applications/mod_av/avformat.c freeswitch-1.10.10/src/mod/applications/mod_av/avformat.c
--- freeswitch-1.10.10.pristine/src/mod/applications/mod_av/avformat.c  2023-08-14 17:45:02.529331160 +0000
+++ freeswitch-1.10.10/src/mod/applications/mod_av/avformat.c   2023-08-14 23:52:55.012169884 +0000
[...]

PR #2681, opened 2024-12-30
Fixes for mod_av related to signalwire/freeswitch#2202

Korynkai:
This is refactored from the patch listed in #2202 to guard against the exact library versions these changes occurred at.
The changes are as follows:

- `AVOutputFormat::priv_data_size` is marked private and no longer exists in the public API as of FFMPEG 6.0, but found hidden in `AVOutputFormat` as part of `FFOutputFormat`.
- `AVCodecContext::ticks_per_frame` was deprecated as of FFMPEG 6.1.
- libavcodec notes 'do not use `avcodec_close()`' as of FFMPEG 3.1, use `avcodec_free_context()` instead.
- `AVFrame::key_frame` was deprecated as of FFMPEG 6.1, use `AVFrame::flags` instead. 
- `AVInputFormat::read_seek` and `AVInputFormat::read_seek2` no longer exists as of FFMPEG 7.0. This seems to only be used for logging purposes, so guard this functionality accordingly.

### DIFF
--- a/src/mod/applications/mod_av/avcodec.c
+++ b/src/mod/applications/mod_av/avcodec.c
@@ -1226,9 +1226,13 @@ static switch_status_t open_encoder(h264_codec_context_t *context, uint32_t widt
 	}
 
 	if (context->encoder_ctx) {
+#if (LIBAVCODEC_VERSION_INT < AV_VERSION_INT(57,48,101))
 		if (avcodec_is_open(context->encoder_ctx)) {
 			avcodec_close(context->encoder_ctx);
 		}
+#else
+		avcodec_free_context(&(context->encoder_ctx));
+#endif
 		av_free(context->encoder_ctx);
 		context->encoder_ctx = NULL;
 	}
@@ -1319,9 +1323,13 @@ FF_ENABLE_DEPRECATION_WARNINGS
 		}
 
 		if (context->encoder_ctx) {
+#if (LIBAVCODEC_VERSION_INT < AV_VERSION_INT(57,48,101))
 			if (avcodec_is_open(context->encoder_ctx)) {
 				avcodec_close(context->encoder_ctx);
 			}
+#else
+			avcodec_free_context(&(context->encoder_ctx));
+#endif
 			av_free(context->encoder_ctx);
 			context->encoder_ctx = NULL;
 		}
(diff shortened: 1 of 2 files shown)


arduent:
This seems to work for me with FFMpeg 7, but also need a few other changes i think.. (but also make sure it works with old versions of FFMpeg too, i suppose...

```
# ffmpeg -version
ffmpeg version N-119396-g48c0dba23b Copyright (c) 2000-2025 the FFmpeg developers
built with gcc 12 (Debian 12.2.0-14)
libavutil      60.  2.100 / 60.  2.100
libavcodec     62.  1.102 / 62.  1.102
libavformat    62.  0.102 / 62.  0.102
libavdevice    62.  0.100 / 62.  0.100
libavfilter    11.  0.100 / 11.  0.100
libswscale      9.  0.100 /  9.  0.100
libswresample   6.  0.100 /  6.  0.100
libpostproc    59.  1.100 / 59.  1.100
```

add_stream() around line 525 in src/mod/applications/mod_av/avformat.c

```
add_stream(arg,arg,arg,arg,,,,) {

const enum AVSampleFormat *sample_fmts;
int ret;

...etc...
c = av_get_codec_context(mst);

ret = avcodec_get_supported_config(c, *codec, AV_CODEC_CONFIG_SAMPLE_FORMAT,
                                                0, (const void **) &sample_fmts, NULL);
if (ret<0) {
	return status; //no sample formats!
}

/* i suppose we want sample_fmts[0] but maybe this 
    needs re-worked.. ie not return if ret<0... in this case 
    c->sample_fmt will never be set to AV_SAMPLE_FMT_FLTP
*/
switch ((*codec)->type) {
		case AVMEDIA_TYPE_AUDIO:
			c->sample_fmt  = sample_fmts ? sample_fmts[0] : AV_SAMPLE_FMT_FLTP;
			c->bit_rate    = 128000;
			
			...
```

also need to change some constants in avcodec.c:

src/mod/applications/mod_av/avcodec.c

```
FF_PROFILE_H264_BASELINE becomes AV_PROFILE_H264_BASELINE
FF_PROFILE_H264_MAIN becomes AV_PROFILE_H264_MAIN
FF_PROFILE_H264_HIGH becomes AV_PROFILE_H264_HIGH
```

PR #2697, opened 2025-01-06
fix(sofia.conf.xml): replace deprecated rtp-hold-timeout-sec

seuros:
I think the default configuration should not be using deprecated param


seuros:
@seven1240 Done, i replaced it in every examples, i also updated the tests.

ticpu:
Did you test that out?

media_timeout is not in seconds and I'm not even sure it can be set at this place.  Can you confirm the timeout really apply from that point and that calls can complete in high latency environment (>300ms) ?

ticpu:
@seuros I see you updated the pull request but the same issue persist.  `media_timeout` isn't a mod_sofia parameter anymore.  It is part of the media core and shouldn't be exposed as an example `sofia` param.

Furthermore, it is now in milliseconds so 300 means 0.3 second.

Either remove it from the sample configurations or add it somewhere in the dialplan instead.

seuros:
@ticpu Thank you for input, i did test that branch , and your concern is valid, the timeout changed from 300s to .3 s.

I never noticed the bug because my usage for FS is automated, and never trigger it. 

I rebased in the morning via the GUI to remember to handle this issue. 

I'm pushing the correction in few moment.

PR #2709, opened 2025-01-10
[mod_opusfile] New featuress:  bitrate settings for each samplerate, appending to existing ogg-opus, etc.

mishehu:
It is now possible to:

-Append to an existing recorded file.  Each appended stream becomes an additional audio stream in the ogg container.
-The ability to set different target bitrates for NB, WB, UWB, and FB, with defaults within the ranges suggested in RFC 6716 sec 2.1.1.
-The ability to limit the maximum sample sample rate for recording file, which can further limit anything set with record_sample_rate .  Thus a g711u call will not be upsampled to 48 kHz, and instead can use the lower necessary bitrate for NB.
-Metadata as defined by switch_audio_col_t now gets added to the ogg container's comments.

### DIFF
--- a/conf/vanilla/autoload_configs/opusfile.conf.xml
+++ b/conf/vanilla/autoload_configs/opusfile.conf.xml
@@ -1,0 +1,18 @@
+<configuration name="opusfile.conf">
+    <settings>
+        <!-- bitrate is in kbps, per channel, so an entry of 96 is 96 kbps -->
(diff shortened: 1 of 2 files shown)


shaunjstokes:
Tested on master and working.

This has solved our issues with using mod_opusfile when combined with record_append and record_stereo. Without this patch the module mod_opusfile is essentially broken and errors when attempting to start recording in certain scenarios.

PR #2714, opened 2025-01-10
mod_redis: remove deprecated module

seuros:
This module was deprecated but is still show up in a fresh installation.


seuros:
ping @andywolk 

Can you check if this is mergeable ?

PR #1397, opened 2021-10-22
[mod_gsmopen] Use TIME_T_FMT

micmac1:
```
libcs are implementing changes to fix the year 2038 issue on 32 bit
platforms (see [1]). musl libc already went ahead and implemented it,
starting with musl-1.2.0 (see [2]).

FreeSWITCH already supports time64 libcs since commit
aa71d87528643fd1b3897a64ecec8c11e92b5b55. But as it turns out, mod_gsmopen
doesn't use TIME_T_FMT (everything else does). And now with a time64 libc
compiling for a 32 bit target, warnings like this appear:

In file included from gsmopen_protocol.cpp:37:
gsmopen.h:121:97: warning: format '%ld' expects argument of type 'long int', but argument 16 has type 'time_t' {aka 'long long int'} [-Wformat=]
  121 | #define DEBUGA_GSMOPEN(...)  switch_log_printf(SWITCH_CHANNEL_LOG, SWITCH_LOG_DEBUG,            "rev %s [%p|%-7lx][DEBUG_GSMOPEN  %-5d][%-10s][%2d,%2d,%2d] " __VA_ARGS__ );
gsmopen.h:121:97: note: in definition of macro 'DEBUGA_GSMOPEN'
  121 | #define DEBUGA_GSMOPEN(...)  switch_log_printf(SWITCH_CHANNEL_LOG, SWITCH_LOG_DEBUG,            "rev %s [%p|%-7lx][DEBUG_GSMOPEN  %-5d][%-10s][%2d,%2d,%2d] " __VA_ARGS__ );
      |                                                                                                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
gsmopen_protocol.cpp:856:104: note: format string is defined here
  856 |                                         DEBUGA_GSMOPEN("GSMOPEN_STATE_RING call_incoming_time.tv_sec=%ld\n", GSMOPEN_P_LOG, tech_pvt->call_incoming_time.tv_sec);
      |                                                                                                      ~~^
      |                                                                                                        |
      |                                                                                                        long int
      |                                                                                                      %lld

This commit adds TIME_T_FMT to mod_gsmopen. All warnings are gone.

[1] https://sourceware.org/glibc/wiki/Y2038ProofnessDesign
[2] https://musl.libc.org/time64.html

Signed-off-by: Sebastian Kemper <sebastian_ml@gmx.net>
```

PR #2720, opened 2025-01-13
[mod_httapi] use SWITCH_TIME_T_FMT for better cross build

seven1240:


andywolk (approved):

andywolk:
After re-looking at it I have concerns that this patch is required.
`switch_epoch_time_now` is `time_t` so `TIME_T_FMT` is already correct. Unless I am missing something.

andywolk:
@micmac1 What was the issue with TIME_T_FMT?
https://github.com/openwrt/telephony/blob/master/net/freeswitch/patches/490-build-properly-fix-time_t-issues.patch#L27
Why did you have to do the patch this way?

micmac1:
Hi Andy, it's about the year 2038 libc problem if I remember correctly, see links below:

https://en.wikipedia.org/wiki/Year_2038_problem
https://github.com/signalwire/freeswitch/pull/1409
https://github.com/signalwire/freeswitch/pull/1397

Basically libc's are start to force 64bit time_t even on 32bit platforms. Not sure if glibc is already doing it, but musl is for example.

andywolk:
@micmac1 Would be nice if I could have a way to repro this behavior building FreeSWITCH locally. Do you have an idea how I could reproduce an environment for that? May be using Docker.

PR #2727, opened 2025-01-14
[mod_curl] Increase argument limit and enforce max args constraint

ar45:
Introduced a macro `MOD_CURL_MAX_ARGS` for the number of arguments for mod_curl functions with a default of 30. Adjusted the code to handle scenarios exceeding this limit by logging an error and returning failure. This improves flexibility and prevents potential overflow issues.


andywolk (approved):

andywolk:
@ar45 
Shouldn't `argc >= MOD_CURL_MAX_ARGS` be `argc > MOD_CURL_MAX_ARGS`?

ar45:
@andywolk  good catch

Indeed. Since we did `char *argv[MOD_CURL_MAX_ARGS + 1]`

andywolk:
Would you like to follow up a PR for that?

PR #2734, opened 2025-01-15
Add support for UUID versioning and generation of UUIDv7.

ar45:
This update introduces support for generating UUIDs with specific versions, including the newly implemented UUIDv7. The `create_uuid` API has been updated to accept optional version parameters. Additionally, a new utility function `switch_getentropy` was added to ensure secure random number generation.


seven1240:
I have a different patch based on https://github.com/LiosK/uuidv7-h, would you like to compare with this?

ar45:
> I have a different patch based on https://github.com/LiosK/uuidv7-h, would you like to compare with this?

Yes, can you share?

seven1240:
> > I have a different patch based on https://github.com/LiosK/uuidv7-h, would you like to compare with this?
> 
> Yes, can you share?

https://github.com/signalwire/freeswitch/pull/2746 see if you like it

PR #2738, opened 2025-01-16
[mod_clearmode] Remove from tree

andywolk:


mrambossek:
hi, may i ask why mod_clearmode was removed? (we are kind of still using it)

skainz92:
Yeah, i would also like to know that. 
I know that you can disable transcoding with proxy_media or bypass_media, but how do i offer a clearmode codec in sdp without mod_clearmode?

PR #2828, opened 2025-06-26
[mod_sofia] allow setting Retry parameter on 503

ticpu:
Was hard-coded at 300 seconds, new variable in sofia profile: `unavailable_retry_after_timeout`


balusreekanth:
Hi SignalWire team, I've tested this update and it's working fine on my side. I just wanted to check if there are any plans to merge this change, or if you'd prefer to implement a similar approach for OPTIONS responses in a different way.
This is specifically about making the Retry-After timer configurable in FreeSWITCH for OPTIONS responses ("503 Maximum Calls In Progress" in pause mode) . Please let me know your thoughts .. I'm happy to adjust if needed.
Thanks!
cc: @ticpu

PR #2832, opened 2025-07-02
[mod_opus] Fix to allow transcoding between different rates

ticpu:
This is in relation to #2226 and its discussions.

It includes a comprehensive list of patches that were suggested in isolation but that work better as a whole.

Thanks to @celliso1, @shaunjstokes, @magiclin99 and all others who have contributed to fixing this issue.

Obsoletes: #2582, #2623.
Fixes: #2226

### DIFF
--- a/src/switch_core_media.c
+++ b/src/switch_core_media.c
@@ -5512,6 +5512,9 @@ SWITCH_DECLARE(uint8_t) switch_core_media_negotiate_sdp(switch_core_session_t *s
 					}
 				}
 
+				if (fmtp_remote_codec_rate) {
+					remote_codec_rate = fmtp_remote_codec_rate;
+				}
 				for (i = 0; i < smh->mparams->num_codecs && i < total_codecs; i++) {
 					const switch_codec_implementation_t *imp = codec_array[i];
 					uint32_t bit_rate = imp->bits_per_second;
@@ -5533,10 +5536,7 @@ SWITCH_DECLARE(uint8_t) switch_core_media_negotiate_sdp(switch_core_session_t *s
 					} else {
 						match = (!strcasecmp(rm_encoding, imp->iananame) &&
 								 ((map->rm_pt < 96 && imp->ianacode < 96) || (map->rm_pt > 95 && imp->ianacode > 95)) &&
-								 (remote_codec_rate == codec_rate || fmtp_remote_codec_rate == imp->actual_samples_per_second)) ? 1 : 0;
-						if (fmtp_remote_codec_rate) {
-							remote_codec_rate = fmtp_remote_codec_rate;
-						}
+								 (remote_codec_rate == codec_rate || remote_codec_rate == imp->actual_samples_per_second)) ? 1 : 0;
 					}
 
 					if (match && bit_rate && map_bit_rate && map_bit_rate != bit_rate && strcasecmp(map->rm_encoding, "ilbc") &&


celliso1:
The commit shows redundant use of `actual_samples_per_second`:

					uint32_t codec_rate = imp->actual_samples_per_second;

					if (!strcasecmp(map->rm_encoding, "opus")) {
						codec_rate = imp->actual_samples_per_second;
					}`

Did you mean this instead ? So only opus is affected:

					uint32_t codec_rate = imp->samples_per_second;

					if (!strcasecmp(map->rm_encoding, "opus")) {
						codec_rate = imp->actual_samples_per_second;
					}`

See https://github.com/signalwire/freeswitch/pull/2582#issuecomment-2410918257 regarding #2623.

ticpu:
@celliso1 oh.. my bad. True, it does affect recordings in an unpredictable way like this.  I will correct this, thank you.

ticpu:
Can confirm this did cause issues with G722 not negotiating 8kHz vs 16kHz, we'll be testing that fix soon.

Right now, I also have other issues with our mod_oreka having weird saturated audio and are still diagnosing if it is related to this PR or FreeSWITCH update.

ticpu:
Moving to draft as hard-coding `samples = 480`  causes issues in many places other than to the call itself. Media bugs, recordings, captures are all affected.
I'm currently working on a different patch that will use `switch_rtp_set_interval` to fix it and looks promising.

ticpu:
Update on this, even my other fix to adjust RTP scaling isn't fully reliable.  Some situations still cause either recordings or audio issues.  The only way I found to get around all audio issues is to go through the normal transcoding path whenever there's a frequency mismatch.

I'll be testing the patch from @nltd101 in combination with @shaunjstokes one then mark this ready if all parts of the solution end up working. I improved the commit message from PR #2833 to include the details of the fix as well.

I think both commits will be required to have a fully working system using Opus both on the SBC and internally.

Edit: I'll go ahead and test these commits on a couple softphones, then move to the PBX and SBC. We really need Opus to never do near-matches that can result in broken audio. I'll come back in about 2 weeks with the results of our testing.

ticpu:
I wasn't able to get the RTP timestamp to align properly in all situations without transcoding.
- The samples = 480 patch broke other applications like recording and playback.
- My RTP timestamp rescaler patch didn't have a 100% success rate and could result in one-way audio which wasn't acceptable for our use-case and would've required having testing facilities we can't afford.
- The 2 current patches togheter ensure that different rate forces transcoding and prevents partial match that caused the RTP timestamp issues and seem to be successful.

 @celliso1 on your side, are you still using the samples = 480 patch successfully?  Do you have specific requirements where transcoding isn't acceptable?  I'm not sure how we could handle that especially that the patch requires modifications at modules/codec *registration*.

I mean, it would have to be a module option on-load; it can't easily be activated per provider, context or profile.

celliso1:
>  @celliso1 on your side, are you still using the samples = 480 patch successfully? Do you have specific requirements where transcoding isn't acceptable? I'm not sure how we could handle that especially that the patch requires modifications at modules/codec _registration_.

Yes we still use that patch. It's been a while, but if I recall correctly there was an issue with playback on the other end when the timestamps were incorrect. My use case is specifically transcoding opus VoIP clients to/from AMR-WB or G711 PSTN trunks. The opus clients use 16Khz to match AMR-WB. I'm not aware of any PTSN application that goes above 16kHz. There are two FreeSWITCH nodes, one facing PSTN trunks and another facing VoIP clients. All playback and call recording happens on the PSTN node, and all transcoding happens on the VoIP node, so we didn't encounter the issues you describe. 

Are you in need of testing facilities?

PR #2844, opened 2025-07-11
[UTILS] FSGET: Use `/etc/apt/auth.conf.d/*.conf` for APT auth

s3rj1k:
Use `/etc/apt/auth.conf.d/*.conf` to store APT netrc file


s3rj1k:
Note: Before we publish that script to well-known link we need to ensure that CI jobs known where to look for new auth file location, for cleanup.

PR #2863, opened 2025-07-16
mod_xml_curl: Add gateway-unix-socket-path option for direct UNIX socket connection

wmasilva:
to use if you are serving the configuration in the same machine and avoid the tcp connection to localhost.


andywolk (changes requested):
There are no UNIX sockets on Windows. Please `#ifdef`.

PR #2881, opened 2025-08-06
Fix segfault in get bleg_* time vars and add ability to get other leg variables

greenbea:


greenbea:
Segault explained in issue https://github.com/signalwire/freeswitch/issues/2883

Issue #2430, opened 2024-04-13
force codec not call the ffmpeg api in av_file_open with ffmpeg5.1 


visitewen:
**Describe the bug**
in Latest version， mod_av force rtmp to H264 codec，seems not take effect,code below:
	
![image](https://github.com/signalwire/freeswitch/assets/47912034/dd81fde3-2891-4aae-a534-779308d53355)

the codec variable only Assigned，but not call ffmpeg api to take effect.

**Package version or git hash**
  freeswitch1.10.11 ffmpeg 5.1.3

**Trace logs**
![image](https://github.com/signalwire/freeswitch/assets/47912034/c9ac6822-2994-46ea-b270-5052ef7f0cbb)

hthetiot:
I confirmed the issue started on `v1.10.10`

Captured lines in above issue description refer to https://github.com/signalwire/freeswitch/blob/9d69bd607b94a5ff4789ab5a5874d8e3b74dca98/src/mod/applications/mod_av/avformat.c#L2328

The issue appear when using `v1.10.11`, `v1.10.9` is not impacted and use `h264` instead of `flv` properly, see capture of proper Trace logs when `h264` is used. 

<img width="1306" height="87" alt="Image" src="https://github.com/user-attachments/assets/d2c93f56-73ce-4748-92f4-31599059f3ed" />

hthetiot:
Quick update of what I found to reproduce the issue.

It is related to `ffmpeg4` vs `ffmpeg5`, `1.10.9` working version was on debian `bullseye` and `1.10.10+` started on `bookworm`, if I build freeswitch `1.10.12` on debian `bullseye` I get h264 instead of FLV because of libav packages on 4.3.8 versions when on libav packages 5.1.6 on `bookworm` fails.

Freeswitch 1.10.12 on debian `bookworm` based on ffmpeg v5.1.6
```
2025-07-09 00:29:02.195969 96.47% [INFO] avformat.c:3337 use video codec implementation Video: flv1 (flv), yuv420p(pc, bt709/unknown/unknown), 1280x720, q=2-31, 1320 kb/s
2025-07-09 00:29:02.195969 96.47% [NOTICE] avformat.c:958 video thread start
freeswitch@sip.sylaps.net> /quit
root@sip:/usr/src/freeswitch# dpkg -l | grep libav
[...]

PR #2896, opened 2025-08-23
[mod_av] Add support for FFmpeg 8.0

bkuhls:
remove deprecated FF_API_FF_PROFILE_LEVEL:
https://github.com/FFmpeg/FFmpeg/commit/822432769868da325ba03774df1084aa78b9a5a0


hthetiot:
@bkuhls hello, can you check this fix i found for video codec when recording when using ffmpeg version is above 5

https://github.com/signalwire/freeswitch/issues/2430#issuecomment-3050877772

Could be nice to have both fixes if you want to invlude my patch. Otherwise will do PR but since you seams knowledgeable on this topic.

Issue #2688, opened 2025-01-02
FreeSWITCH angers Poly phones due to NOTIFY CSeq rollover at midnight January 1 (year boundary)


rlaager:
**Describe the bug**
Exactly at 00:00 on January 1, FreeSWITCH rolls over the CSeq values used in NOTIFY packets. This makes our Poly phones mad. When they get their first NOTIFY with a lower CSeq, they reply with "500 Internal Server Error" and start spamming the server with SUBSCRIBEs. The resulting server load is easily high enough to impact service. The phones are fine after a reboot.

In looking at the RFCs, the phones seem to be in the wrong for expecting an ever increasing CSeq. But I'm trying to see if anything better can be done within real world limitations, including obviously the fact that we can't fix the phones' firmware.

**To Reproduce**
1. Use presence with Poly phones.
2. Wait until 00:00 January 1.

**Expected behavior**
This is the tricky part. Obviously I don't want the Poly phones to crash. But it's unclear to me what can be done better on the FreeSWITCH side.

There were some discussions about this in the past:
https://lists.freeswitch.org/pipermail/freeswitch-dev/2018-August/007873.html
https://freeswitch-users.freeswitch.narkive.com/TYeYlPf7/notify-s-cseq-too-high

I believe the whole reason that FreeSWITCH is doing this presence CSeq thing, including the presence epoch, is to avoid this problem when FreeSWITCH is restarted (which, if nothing special was done, would otherwise cause the CSeq to roll over).

In that first linked message, @anthmFS said, "Many phones break until you reset the whole registration if they are unhappy with the cseq." That seems to suggest that the phones might accept a CSeq rollover when they re-REGISTER. If that's the case (which is a big if), would it be reasonably possible to calculate the presence CSeq based on the timestamp of the last registration? If phones really do accept the rollover when REGISTERing, this would be better. If they do not, then this would make the problem massively worse, as the phones would freak out on every REGISTER rather than just once a year.

**Package version or git hash**
1.10.11~release~25~f24064f7c9~buster-1~buster+1 but the issue is obviously still present in the code in git master: https://github.com/signalwire/freeswitch/blob/4470f8d801b3ac687a4b09b4a9cc7bab29689b0f/src/mod/endpoints/mod_sofia/sofia_presence.c#L2113

MartinJPaterson:
[...]

PR #2959, opened 2025-11-30
Fix for #2688 New Years Polycom NOTIFY Cseq Rollover

junction1153b:


markjcrane:
Thanks for bringing this up as a Fix. Without this fix or something like it everyone needs to reboot their phones every New Year.

This affects almost everyone who uses FreeSWITCH. The Cseq value is supposed to be incremented. At the beginning of the year this value resets. This requires phones to be restarted every New Year.

junction1153:
All our deployed systems hit 1/1/2026 over 24 hours ago, all is still stable. No phones had to be rebooted.

rlaager:
So this keeps the whole check_presence_epoch(), which if I am reading correctly, returns a value of 50 times the number of seconds into the year, but then just starts incrementing by 1 after the year rollover?

I think the original idea behind check_presence_epoch() is that it is deterministic, so that restarting freeswitch does not cause CSeq rollover. If you restart freeswitch, presumably all your phones will break now.

That’s the tricky part of this whole thing. It seems like the ideal solution in theory would be to keep track of the CSeq for each device in the database, but that would be likely terrible from a performance perspective.

junction1153:
> So this keeps the whole check_presence_epoch(), which if I am reading correctly, returns a value of 50 times the number of seconds into the year, but then just starts incrementing by 1 after the year rollover?
> 
> I think the original idea behind check_presence_epoch() is that it is deterministic, so that restarting freeswitch does not cause CSeq rollover. If you restart freeswitch, presumably all your phones will break now.
> 
> That’s the tricky part of this whole thing. It seems like the ideal solution in theory would be to keep track of the CSeq for each device in the database, but that would be likely terrible from a performance perspective.

Without an SBC, if Freeswitch is rebooted, in my experience everything breaks anyhow - Polycom/Yealink. Need a full phone reboot or a new REGISTER. With an SBC in place, the potential concern you addressed is a non issue

rlaager:
> Without an SBC, if Freeswitch is rebooted, in my experience everything breaks anyhow - Polycom/Yealink.

We reboot FreeSWITCH all the time without issues. We reboot once a month for system patching, for example.

junction1153:
> We reboot FreeSWITCH all the time without issues. We reboot once a month for system patching, for example.

What phones do you have there?

Have you tried this patch yet? If you are experiencing an issue after application of the patch (but can also verify there is no problem before implementation of it), I can see if there is the possibility to tweak this

rlaager:
Poly VVX, mostly [134]50 and 41[01]. Yealinks, mostly T4[36]U.

We didn't think this happened with Yealinks last year. Either we didn't notice, or Yealink broke this in a software update.

I have not tried the patch. I see how it would avoid the year rollover issue, but AFAICT it trades it for breaking on the next reboot, which for me is worse. Right now, we have automation to reboot all the (Poly, but next year Poly + Yealink) phones right after the year rollover.

junction1153:
I’ll test this out in the lab later today. Thinking out loud, we can preserve the CSEQ on reboot by pushing out the data to the database. I need to think about this one, if it is indeed posing an issue 

the new year issue explicitly affects Polycom phones

junction1153:
Tested in the lab with a few phones. Immediately after the reboot, the BLF did not work (very normal behavior even pre-patch). After the phones renewed their SUBSCRIBE, per the expiration threshold configured within the phone settings, the BLF lamp illuminated when it was commanded to.

rlaager:
Maybe this is the right fix then? The stock behavior keeps things working as much as possible through the year, then we switch to monotonically incrementing. With this patch, does freeswitch go back to the deterministic seconds * 50 thing at some point (like when the phone reREGISTERs or anything)?

junction1153:
I was able to produce an issue by setting the system to december 31 and rebooting freeswitch after the new year rolled over. What you described is correct. I am working on a solution and will post back here later. Thank you for bringing this to my attention.

junction1153b:
Give the latest commit a try. This now survives new years and reboots, as well as database wipes (in case you set your sqlite to ramdisk and reboot the server)

In order to survive freeswitch reboots, we are keeping record of the latest cseq by creating a table sip_presence_seq

If the table does not exist or is set to 0 due to a database wipe, we then check the storage directory / presence_seq_internal.txt. We are periodically logging the latest cseq to that file

rlaager:
It’d be nice if we could get a FreeSWITCH developer to weigh in with a design decision here.

I’m just looking at this PR on my cell phone, but it seems like this is still leaving all the year-based stuff in place. If the design is to keep everything on the database, shouldn’t that go away?

junction1153:
> I’m just looking at this PR on my cell phone, but it seems like this is still leaving all the year-based stuff in place. If the design is to keep everything on the database, shouldn’t that go away?

The design is to not roll-back the Cseq which was previously done and was causing these issues. Just keep incrementing the Cseq. Now that I look more into it, eventually, Cseq is going to have to roll back when it exceeds its limit. I am going to have to re-engineer this or close this PR completely and figure it out when I have time to dedicate to this.

Thank you for bringing up all these concerns, they are valid and will help lead to a permanent solution

rlaager:
> The design is to not roll-back the Cseq which was previously done and was causing these issues. Just keep incrementing the Cseq.

Right.

In the PR:

1. When the value is zero (unset): A. Check the database. B. Check a flat file. C. If neither, create the table.
2. Run the existing deterministic calculation.
3. If and only if the value from step 2 rolled over, increment the value from step 1.

I'm skeptical of the idea of trying to save this in two places. You're doubling the write workload. It seems like it should be one or the other. I'd imagine the database is better. I see the concern about the database being potentially ephemeral. But FreeSWITCH has non-ephemeral database tables, e.g. for voicemail.

The other database tables seem to be created in `sofia_glue_init_sql()` in `src/mod/endpoints/mod_sofia/sofia_glue.c`. So that (or if I'm wrong, whatever is creating the other tables) should be the place to create `sip_presence_seq`. (Unless that ends up putting them in the core db, which might be ephemeral?) Also, it should maybe have an index: `create index sps_multi on sip_presence_seq (profile_name, hostname);`

I was going to complain about the `DELETE + INSERT`, but there is no SQL standard way to do an UPSERT apparently, so I guess that's what has to be done.

If storing this in the database is acceptable, then why not:
1. When the value is zero: Check the database.
2. Increment the value.

> Now that I look more into it, eventually, Cseq is going to have to roll back when it exceeds its limit.

Sure, but 2^31 seconds is 68 years. (For example, 32-bit signed time_t goes from 1970 to 2038. This is an unsigned 31-bit integer, which works out the same.)

junction1153:
> I'm skeptical of the idea of trying to save this in two places. You're doubling the write workload. It seems like it should be one or the other. I'd imagine the database is better. I see the concern about the database being potentially ephemeral. But FreeSWITCH has non-ephemeral database tables, e.g. for voicemail.

Many users have their SQLite database in RAM disk. What happens when the server restarts? All the data is gone. Back to the 500 Internal Server error because Cseq went back to 0. I am at a dead end with that



> Sure, but 2^31 seconds is 68 years

Yes, but many Freeswitch users have thousands of endpoints registered. That 68-year figure can easily turn into a few months, or even a few years, for larger users. This can't be risked, and there is no way they can accept that PR with this type of instability for their enterprise users

rlaager:
> > I'm skeptical of the idea of trying to save this in two places. You're doubling the write workload. It seems like it should be one or the other. I'd imagine the database is better. I see the concern about the database being potentially ephemeral. But FreeSWITCH has non-ephemeral database tables, e.g. for voicemail.
> 
> Many users have their SQLite database in RAM disk. What happens when the server restarts? All the data is gone.

Again, there are existing tables in FreeSWITCH that need to be preserved. It's just a matter of ensuring that this is handled like voicemail_* (which need to be preserved), not like the other core-db things that may be moved to an ephemeral ramdisk.

> > Sure, but 2^31 seconds is 68 years
> 
> Yes, but many Freeswitch users have thousands of endpoints registered.

To check my understanding: The current code calculates a value profile-wide. So regardless of the number of endpoints, the CSeq value moves forward only with the passage of time. My proposal was to increment the CSeq value each time one is needed. So for multiple endpoints, this would increment quickly.

In my experience, FreeSWITCH presence already falls on its face if you get too many endpoints. But still, you make a good point.

Maybe this should be per endpoint, not profile-wide. That would make the CSeq per endpoint, which solves that concern. The problem there, of course, is that now we are amplifying the database load by the number of endpoints. Of course, there remains the option to only update the database every N updates, and on restart, bump the value from the database by N (or N*2 to be safe). I'm just not sure whether this would scale well enough in the real world.

mjerris:
this is a bug in the poly firmware.  all of these fixes break reliability in some other way.  the best way to fix this isn't to trade for
other bugs but to ask poly to fix their rollover handling in their firmware. (which is of legal and expected).

Issue #2960, opened 2025-12-05
Add Shell Script Handling to .gitattributes


nativegold:
When entrypoint.sh is edited on Windows, the file may end up with Windows-style line endings (CRLF).
This can cause the script to fail when the Docker container starts, because the shell inside the container expects Unix-style line endings (LF).

To prevent this, we should update .gitattributes to ensure all shell scripts use LF:

` 
*.sh text eol=lf
`

This will guarantee that entrypoint.sh (and any other .sh files) work correctly in Docker environments regardless of the host OS.

PR #2961, opened 2025-12-05
Add Shell Script Handling to .gitattributes #2960

nativegold:
When entrypoint.sh is edited on Windows, the file may end up with Windows-style line endings (CRLF).
This can cause the script to fail when the Docker container starts, because the shell inside the container expects Unix-style line endings (LF).

To prevent this, we should update .gitattributes to ensure all shell scripts use LF:

 *.sh text eol=lf

This will guarantee that entrypoint.sh (and any other .sh files) work correctly in Docker environments regardless of the host OS.


vishalk17:
intresting !

PR #3051, opened 2026-06-05
Add flag to skip application updates in channels table

dhruvecosmob:
**Summary**
Add a new FreeSWITCH flag ignore_application_channel_update to reduce unnecessary SQL update activity on the channels table.

**Problem**
FreeSWITCH updates the application field in the channels table whenever the executing application changes. In high call volume environments this generates a large number of SQL UPDATE statements, increasing database load and contention.
it makes sense to remove the extra queries if not needed; agree that in theory it's a bottleneck.

**Solution**
Introduce a new configuration flag:
`ignore_application_channel_update`
When enabled, FreeSWITCH skips updates to the application column in the channels table while continuing to maintain other channel state information.

**Benefits**
- Reduces SQL UPDATE traffic on the channels table
- Lowers database load in high CPS environments
- Improves scalability for deployments that do not rely on the application column

**Configuration**
`<param name="ignore-application-channel-update" value="true"/>`
Default behavior remains unchanged when the parameter is disabled.

**Testing**
- Verified normal call processing
- Verified channels table updates continue
- Verified application column updates are skipped when enabled
- Verified default behavior is unchanged when disabled
- Performed heavy load testing with approximately 2000 concurrent calls
- Verified stable call handling and reduced SQL update activity under load
- No regressions observed during high-load testing

### DIFF
--- a/conf/vanilla/autoload_configs/switch.conf.xml
+++ b/conf/vanilla/autoload_configs/switch.conf.xml
@@ -206,6 +206,9 @@
 
     <!-- <param name="max-audio-channels" value="2"/> -->
 
+    <!-- The system will ignore update application information in channel table, set this to false to avoid this behaviour -->
+    <param name="ignore_application_channel_update" value="true" />
+
   </settings>
 
 </configuration>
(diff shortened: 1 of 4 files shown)


davesoft11:
Key clarification is that the "extra" queries do not appear, based on the call state context, to be necessary, and therefore are extraneous (and very burdensome). Obviously essential queries cannot be eliminated. Thanks!

PR #3057, opened 2026-06-12
[mod_sofia] Fix broken sip: prefix check in deflect handler

### LINKED ISSUE #3049: SWITCH_MESSAGE_INDICATE_DEFLECT: strcasecmp check broken since FS-6164 (commit 8bf70dcf)
## Version
FreeSWITCH 1.11.2-dev (also affects all versions since 2014 commit 8bf70dcf)

## Description
In src/mod/endpoints/mod_sofia/mod_sofia.c, the deflect handler uses:

c
if (!strcasecmp(msg->string_arg, "sip:")) {

This is dead code. strcasecmp compares the entire string against "sip:" — it only matches if the input is literally just "sip:", which is never a valid deflect target.

## History
- **2008** (commit 021ab87f, MODENDP-66): Original code used !strstr(msg->string_arg, "sip:") — correctly detected absence of sip: and auto-constructed a URI for bare usernames.
- **2014** (commit 8bf70dcf, FS-6164): Changed to !strcasecmp(...) intending to add case-insensitivity, but broke the logic entirely since strcasecmp is a whole-string comparison, not a
substring/prefix check.

## Impact
Passing a bare number/username to deflect or uuid_deflect (e.g., uuid_deflect <uuid> 1234) results in a malformed Refer-To header without the sip: scheme. The remote device rejects or 
ignores the REFER.

Possibly related: #1697

## Fix
Replace:
c
if (!strcasecmp(msg->string_arg, "sip:")) {

With:
c
if (strncasecmp(msg->string_arg, "sip:", 4)) {

This checks if the string does NOT start with sip: (case-insensitive), restoring the original intent while adding the case-insensitivity that FS-6164 requested.

andywolk:
Fixes #3049

### DIFF
--- a/src/mod/endpoints/mod_sofia/mod_sofia.c
+++ b/src/mod/endpoints/mod_sofia/mod_sofia.c
@@ -1568,7 +1568,7 @@ static switch_status_t sofia_receive_message(switch_core_session_t *session, swi
 		const char *var;
 		const char *session_id_header = sofia_glue_session_id_header(session, tech_pvt->profile);
 
-		if (!strcasecmp(msg->string_arg, "sip:")) {
+		if (strncasecmp(msg->string_arg, "sip:", 4)) {
 			const char *format = strchr(tech_pvt->profile->sipip, ':') ? "sip:%s@[%s]" : "sip:%s@%s";
 
 			switch_snprintf(ref_to, sizeof(ref_to), format, msg->string_arg, tech_pvt->profile->sipip);


briankwest (approved):

src/mod/endpoints/mod_sofia/mod_sofia.c
```diff
@@ -1568,7 +1568,7 @@ static switch_status_t sofia_receive_message(switch_core_session_t *session, swi
 		const char *var;
 		const char *session_id_header = sofia_glue_session_id_header(session, tech_pvt->profile);
 
-		if (!strcasecmp(msg->string_arg, "sip:")) {
+		if (strncasecmp(msg->string_arg, "sip:", 4)) {
```
mirkobrankovic82, line 1571: @andywolk, @briankwest   why is this done, this is now breaking change for our `tel: number` Refer-To destinations? It also breaks `sips:` destinations

PR #3061, opened 2026-06-28
[Core] Avoid O(n^2) header dedup when cloning EF_UNIQ_HEADERS events

celliso1:
### Problem

`switch_event_dup()` clones a source event by re-adding every header with
`switch_event_add_header_string()`. When the source has `EF_UNIQ_HEADERS`
set — which is the case for `SWITCH_EVENT_CHANNEL_DATA`,
`SWITCH_EVENT_REQUEST_PARAMS` and `SWITCH_EVENT_MESSAGE`, i.e. the events
used to hold a channel's variable list — the destination inherits the
flag, so `switch_event_base_add_header()` calls `switch_event_del_header()`
(a full linear scan of the list built so far) on **every** add to enforce
uniqueness. Duplicating an N-header event is therefore **O(n²)**.

The scan is wasted work: the source list is already unique (uniqueness was
enforced when each header was first inserted), so copying it can never
create a duplicate, with or without the per-add scan.

### Why it matters

`switch_event_dup()` backs `switch_core_get_variables()` and
`switch_channel_get_variables()`, which clone an `EF_UNIQ_HEADERS`
variable event. Any workload that duplicates large `EF_UNIQ_HEADERS`
events — channel-variable snapshots, or ESL consumers cloning big
events — pays this O(n²) on every dup.

On a production voicemail box carrying ~191 channel variables per call,
`switch_event_del_header_val()` (the per-add uniqueness scan) is 47–69%
of FS CPU in `perf`, and call-graph profiles show a substantial share
reached through `execute_on -> *_get_variables -> switch_event_dup` — the
path this patch fixes.

**Scope:** this patch fixes the *dup* path only. `del_header_val` is also
reached via `switch_event_merge()` (also in `execute_on`) and per-variable
`switch_channel_set_variable()`; those are separate O(n²) paths this
change does not address (future work: avoid the dup+merge in `execute_on`,
or hash-index event headers). The exact prod split between these callers
is not yet quantified.

### Fix

Clear `EF_UNIQ_HEADERS` on the destination for the duration of the copy
loop in `switch_event_dup()`, then restore it. ~10 lines, no API change.

### Microbench

`tests/unit/switch_event.c` (`dup_uniq_bench`), dup of an N-header
`CHANNEL_DATA` event, 4000 iterations, time per dup:

| N | before | after | speedup |
|---|---|---|---|
| 50 | 4.80 µs | 3.19 µs | 1.5× |
| 100 | 11.34 µs | 5.88 µs | 1.9× |
| 191 | 28.29 µs | 10.99 µs | 2.6× |
| 400 | 107.05 µs | 23.27 µs | 4.6× |

### Real-call validation

**Validated on production voicemail hosts under real deposit load — across
CPU, hypervisor, and FreeSWITCH version.** Same-node, before vs after
swapping in the patched lib (40 s `perf` windows, self-time):

| environment | `del_header_val` stock | patched | reduction |
|---|---|---|---|
| VMware / Intel / 1.10.12 | 56.7% of FS CPU | 8.5% | −85% |
| VMware / AMD / 1.10.12 | 51.4% | 7.9% | −85% |
| Proxmox / EPYC / 1.11.1 | ~47% | 6.4% | −86% |

A consistent **~85% reduction** in the per-add uniqueness scan across all
three environments — the cache-bound Intel box (highest stock cost) drops
the most in absolute terms, **−48 points of FS CPU**. The call-graph
confirms the mechanism: with the patch the
`switch_channel_execute_on → *_get_variables → switch_event_dup` chains no
longer reach `del_header_val`; the ~6–8% residual is a separate path
(`switch_event_merge`, also reached from `execute_on`) that this change
deliberately does not touch.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [X] Code cleanup / refactor

## Testing

Profiled on production voicemail deposit service.

- [X] Added/updated unit tests
- [X] Tested manually

## Checklist

- [X] I have read the [CONTRIBUTING](CONTRIBUTING.md) guidelines
- [X] My code follows the project's style guidelines
- [X] I have added tests for my changes (if applicable)
- [X] All existing tests pass

### DIFF
--- a/src/switch_event.c
+++ b/src/switch_event.c
@@ -1346,11 +1346,13 @@ SWITCH_DECLARE(switch_status_t) switch_event_dup(switch_event_t **event, switch_
 	(*event)->bind_user_data = todup->bind_user_data;
 	(*event)->flags = todup->flags;
 
-	/* todup's headers are already unique (uniqueness is enforced on insert for an
-	 * EF_UNIQ_HEADERS event), so the per-add duplicate scan that
-	 * switch_event_base_add_header() performs when EF_UNIQ_HEADERS is set is
-	 * redundant while cloning and makes this loop O(n^2) in the header count.
-	 * Suppress the scan for the copy, then restore the flag. */
+	/* The destination inherited todup's flags above. When EF_UNIQ_HEADERS is set,
+	 * switch_event_base_add_header() rescans the partial header list on every add
+	 * to keep names unique, making this copy loop O(n^2). A clone should reproduce
+	 * todup's list verbatim rather than run a fresh uniqueness pass, and a
+	 * well-formed EF_UNIQ_HEADERS source is already unique (its names were deduped
+	 * on insert), so that scan only adds cost here. Suppress it for the copy, then
+	 * restore the flag so later adds enforce uniqueness again. */
 	if (switch_test_flag((*event), EF_UNIQ_HEADERS)) {
 		switch_clear_flag((*event), EF_UNIQ_HEADERS);
 		restore_uniq_headers = 1;
--- a/src/switch_event.c
+++ b/src/switch_event.c
@@ -1335,6 +1335,7 @@ SWITCH_DECLARE(void) switch_event_merge(switch_event_t *event, switch_event_t *t
 SWITCH_DECLARE(switch_status_t) switch_event_dup(switch_event_t **event, switch_event_t *todup)
 {
 	switch_event_header_t *hp;
+	int restore_uniq_headers = 0;
 
 	if (switch_event_create_subclass(event, SWITCH_EVENT_CLONE, todup->subclass_name) != SWITCH_STATUS_SUCCESS) {
 		return SWITCH_STATUS_GENERR;
@@ -1344,6 +1345,17 @@ SWITCH_DECLARE(switch_status_t) switch_event_dup(switch_event_t **event, switch_
 	(*event)->event_user_data = todup->event_user_data;
 	(*event)->bind_user_data = todup->bind_user_data;
 	(*event)->flags = todup->flags;
+
+	/* todup's headers are already unique (uniqueness is enforced on insert for an
+	 * EF_UNIQ_HEADERS event), so the per-add duplicate scan that
+	 * switch_event_base_add_header() performs when EF_UNIQ_HEADERS is set is
+	 * redundant while cloning and makes this loop O(n^2) in the header count.
+	 * Suppress the scan for the copy, then restore the flag. */
+	if (switch_test_flag((*event), EF_UNIQ_HEADERS)) {
+		switch_clear_flag((*event), EF_UNIQ_HEADERS);
+		restore_uniq_headers = 1;
+	}
+
 	for (hp = todup->headers; hp; hp = hp->next) {
 		if (todup->subclass_name && !strcmp(hp->name, "Event-Subclass")) {
 			continue;
@@ -1357,6 +1369,10 @@ SWITCH_DECLARE(switch_status_t) switch_event_dup(switch_event_t **event, switch_
 		} else {
 			switch_event_add_header_string(*event, SWITCH_STACK_BOTTOM, hp->name, hp->value);
 		}
+	}
+
+	if (restore_uniq_headers) {
+		switch_set_flag((*event), EF_UNIQ_HEADERS);
 	}
 
 	if (todup->body) {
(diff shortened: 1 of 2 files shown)


celliso1:
@andywolk — would you have a moment to look at this? It's a self-contained ~10-line hot-path fix (O(n^2) -> O(n) in `switch_event_dup()` for `EF_UNIQ_HEADERS` events) with unit + regression tests.

Since opening it I validated the change on a production voicemail fleet — same-node before/after under real deposit load — and the win holds across CPU, hypervisor, and FreeSWITCH version:

| environment | `del_header_val` stock | patched |
|---|---|---|
| VMware / Intel / 1.10.12 | 56.7% | 8.5% |
| VMware / AMD / 1.10.12 | 51.4% | 7.9% |
| Proxmox / EPYC / 1.11.1 | ~47% | 6.4% |

A consistent ~85% reduction in the per-add uniqueness scan (also summarized in the PR body). Happy to adjust anything — thanks!

tests/unit/switch_event.c
```diff
@@ -104,6 +104,130 @@ FST_TEST_BEGIN(benchmark)
 }
 FST_TEST_END()
 
+FST_TEST_BEGIN(dup_uniq_bench)
```
morbit85, line 165: Could you please gate the whole `dup_uniq_bench` behind `#ifdef BENCHMARK` (like the existing benchmark test in this file)? `dup_faithful_copy` already covers the correctness side.

morbit85 (changes requested):

src/switch_event.c
```diff
@@ -1344,6 +1345,17 @@ SWITCH_DECLARE(switch_status_t) switch_event_dup(switch_event_t **event, switch_
 	(*event)->event_user_data = todup->event_user_data;
 	(*event)->bind_user_data = todup->bind_user_data;
 	(*event)->flags = todup->flags;
+
+	/* todup's headers are already unique (uniqueness is enforced on insert for an
```
andywolk: `todup's headers are already unique (uniqueness is enforced on insert for an
	 * EF_UNIQ_HEADERS event),`
This comment is not accurate and too strict. Not always. Only if `EF_UNIQ_HEADERS` is set. Please rephrase.

tests/unit/switch_event.c
celliso1: Done, sorry I missed that.

andywolk (changes requested):

tests/unit/switch_event.c
```diff
@@ -104,6 +104,132 @@ FST_TEST_BEGIN(benchmark)
...
+  fst_xcheck(status == SWITCH_STATUS_SUCCESS, "create CHANNEL_DATA event");
+  fst_xcheck(switch_test_flag(src, EF_UNIQ_HEADERS) != 0, "CHANNEL_DATA event is EF_UNIQ_HEADERS");
+
+  for (i = 0; i < 191; i++) {
+    switch_snprintf(name, sizeof(name), "var_%d", i);
+    switch_snprintf(val, sizeof(val), "value_%d_some_padding_payload", i);
+    switch_event_add_header_string(src, SWITCH_STACK_BOTTOM, name, val);
+  }
+  S = 0;
+  for (hp = src->headers; hp; hp = hp->next) S++;
```
andywolk: One-line loops are not allowed. I understand that this is just a test, Let's follow our coding guidelines. You can find examples here: https://github.com/signalwire/libks/blob/master/CODING_GUIDELINES

andywolk (changes requested):

src/switch_event.c
celliso1: Thanks @andywolk — both addressed in 7cac871:
- Scoped the comment to EF_UNIQ_HEADERS events and the verbatim-copy intent.
- Expanded the one-line loops in the tests per the guidelines.

tests/unit/switch_event.c
celliso1: Thanks @andywolk — both addressed in 7cac871:
- Scoped the comment to EF_UNIQ_HEADERS events and the verbatim-copy intent.
- Expanded the one-line loops in the tests per the guidelines.

morbit85 (dismissed):

andywolk (dismissed):

andywolk:
Please fix conflicts

celliso1:
Rebased onto current master — the conflicts were between the new expand_headers_offset test from 30cdb387 and this PR's tests landing at the same spot in tests/unit/switch_event.c; both are kept. No changes to the fix itself.

andywolk (approved):

PR #3063, opened 2026-07-01
[mod_cdr_pg_csv] optionally honor force_process_cdr for B-leg CDRs

realsama:
## Description

`mod_cdr_pg_csv` chooses which legs to log solely from the global `legs` setting
(`a` | `b` | `ab`); there is no per-call override. With the default `legs=a`, the
B-leg of a bridged call (e.g. an outbound `bridge` / `<Dial>`) never produces a CDR
and cannot be billed from the pg_csv record, even when the application deliberately
wants it recorded.

`mod_json_cdr` and `mod_format_cdr` already handle this via the standard
`force_process_cdr` channel variable (`SWITCH_FORCE_PROCESS_CDR_VARIABLE`), reported
regardless of the leg filter. `mod_cdr_pg_csv` had no equivalent.

This adds an opt-in `honor-force-process-cdr` setting (default **off**). When
enabled, a B-leg whose `force_process_cdr` variable is set is logged even if the
`legs` filter would otherwise skip it. Enable in `cdr_pg_csv.conf.xml`:

    <param name="honor-force-process-cdr" value="true" />

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Code cleanup / refactor

## Related Issues

None.

## Testing

Built against `master`. The same change (on a 1.10.12 base) is running in production:
with the setting on and a bridged B-leg exporting `force_process_cdr=true`, the dialed
leg now produces its own `cdr` row; with it off, output is identical to current behaviour.

- [x] Tested manually

## Checklist

- [ ] All existing tests pass

### DIFF
--- a/src/mod/event_handlers/mod_cdr_pg_csv/mod_cdr_pg_csv.c
+++ b/src/mod/event_handlers/mod_cdr_pg_csv/mod_cdr_pg_csv.c
@@ -88,6 +88,7 @@ static struct {
 	spool_format_t spool_format;
 	int rotate;
 	int debug;
+	int honor_force_process_cdr;
 } globals;
 
 static switch_xml_config_enum_item_t config_opt_cdr_leg_enum[] = {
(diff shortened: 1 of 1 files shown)


realsama:
Hi @andywolk  could you please take a quick look at this one file PR when you get a chance? Thanks.

Issue #3091, opened 2026-08-03
[mod_callcenter] instance_id column hardcoded as 'single_box' — breaks multi-node deployments


gaurang1102:
**Describe the bug**
The agents and members tables in mod_callcenter have an instance_id column specifically designed to differentiate agents registered on different FreeSWITCH nodes in a multi-node deployment. However, every SQL write in cc_agent_set_value() and the agent load functions hardcodes the string 'single_box' as the instance_id, making the column permanently non-functional.

This means that in any deployment with more than one FreeSWITCH node sharing a common database (a standard HA or geographic distribution scenario), all nodes write identical instance_id values, making it impossible to:

Query which node currently owns a given agent
Filter member queries to agents registered on the same node (serving_system is also hardcoded)
Perform proper node-scoped agent state cleanup on restart

**To Reproduce**
(1) Deploy two FreeSWITCH nodes sharing a common PostgreSQL/MySQL callcenter database.
(2) Register agents on both nodes.
(3) Query: SELECT name, instance_id FROM agents;
(4) Observe that every row shows instance_id = 'single_box' regardless of which node the agent is registered on.
(5) Attempt to filter member queries by serving_system to route calls to the correct node — the filter is non-functional.

**Expected behavior**
Each node writes its own configurable identifier to instance_id. The column meaningfully partitions data by node, enabling multi-node agent isolation and proper serving-system filtering.

**Actual behavior**
All nodes write 'single_box'. The instance_id column is permanently non-functional. Multi-node deployments cannot use callcenter's built-in node partitioning.

**Affected Versions**
All known versions of mod_callcenter (1.6 through 1.10.12, 1.11.1). The column exists in the schema but has never been populated dynamically.

PR #3092, opened 2026-08-03
[mod_callcenter] instance_id column hardcoded as 'single_box' — breaks multi-node deployments

gaurang1102:
## Problem Description ##
The agents and members tables in mod_callcenter have an instance_id column specifically designed to differentiate agents registered on different FreeSWITCH nodes in a multi-node deployment. However, every SQL write in cc_agent_set_value() and the agent load functions hardcodes the string 'single_box' as the instance_id, making the column permanently non-functional.

This means that in any deployment with more than one FreeSWITCH node sharing a common database (a standard HA or geographic distribution scenario), all nodes write identical instance_id values, making it impossible to:

Query which node currently owns a given agent
Filter member queries to agents registered on the same node (serving_system is also hardcoded)
Perform proper node-scoped agent state cleanup on restart

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Code cleanup / refactor

## Related Issues

#3091

## Testing

- [x] Tested manually

## Checklist

- [x] All existing tests pass

### DIFF
--- a/src/mod/applications/mod_callcenter/mod_callcenter.c
+++ b/src/mod/applications/mod_callcenter/mod_callcenter.c
@@ -907,8 +907,8 @@ cc_status_t cc_agent_add(const char *agent, const char *type)
 		/* Add Agent */
 		switch_log_printf(SWITCH_CHANNEL_LOG, SWITCH_LOG_DEBUG, "Adding Agent %s with type %s with default status %s\n",
 				agent, type, cc_agent_status2str(CC_AGENT_STATUS_LOGGED_OUT));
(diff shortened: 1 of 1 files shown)


davesoft11:
Most of FS scales horizontally properly to multiple instances. But mod_callcenter is a notable exception, not supporting instance_id dynamically. The importance of affinity in ACD applications makes this particularly problematic. I'd love to see this upstreamed so our home-grown solutions can be ripped out.

PR #3106, opened 2026-08-06
[Core, mod_sofia] Make the rtp-autofix-timing codec check limit configurable

RicardoMDomingues:
## Description
The CBR branch of the autofix-timing codec check stopped after a hardcoded 50 frames.

Add a sofia profile parameter, rtp-autofix-max-codec-check-frames:

  > 0   inspect that many frames (default 50, unchanged)
  < 0   inspect for the whole call

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Code cleanup / refactor

## Testing

- [ ] Tested manually

## Checklist

- [x] All existing tests pass

## Additional Notes

I need this change because I use FreeSWITCH (FS) with Trader turret systems. In some setupus, media is routed to a media server, and the packet rate can change dynamically between all participants (from the Turret phones, to their gateway, through the customer's SBC, and finally to our system).

When the Turret system changes the packet rate, it does not send a SIP re-INVITE or any other SIP signaling to notify the gateway or SBC. Because of this, there is no way to detect the rate change via SIP. Instead, it can only be detected and handled using the RTP timestamp and payload information.
FreeSWITCH already had a feature to handle this, but it previously only worked at the very beginning of the call. My change extends this functionality so FreeSWITCH by making configurable how many packets we want to check and, if its negative, check for entire call

### DIFF
--- a/conf/vanilla/sip_profiles/internal.xml
+++ b/conf/vanilla/sip_profiles/internal.xml
@@ -373,6 +373,8 @@
     <!--<param name="outbound-use-uuid-as-callid" value="true"/>-->
     <!-- set to false disable this feature -->
     <!--<param name="rtp-autofix-timing" value="false"/>-->
+    <!-- how many initial frames rtp-autofix-timing inspects before stop, -1 to inspect for the whole call (default 50) -->
+    <!--<param name="rtp-autofix-max-codec-check-frames" value="-1"/>-->
 
     <!-- set this param to false if your gateway for some reason hates X- headers that it is supposed to ignore-->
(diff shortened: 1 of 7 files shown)


RicardoMDomingues:
@andywolk, @morbit85 could you please provide feedback on this? I have been keeping a fork of FS for long time because of it. The change does not add any new mechanism, only adds a configuration to replace the old constant `MAX_CODEC_CHECK_FRAMES` while keeps 50 as default

